PR #1, opened 2015-11-12
ref(docs): rename _docs to docs

bacongobbler:
As discussed in https://github.com/deis/helm/pull/138#issuecomment-153182367,
we won't be writing .go files in the docs folder so it doesn't make sense to
add an underscore here.


mboersma:
Oh, ok, I was taking my cues from [deis/etcd](https://github.com/deis/etcd) and [deis/postgres](https://github.com/deis/postgres), but your explanation makes sense to me.

technosophos:
Yeah, the prototype for new repos still has `_docs` instead of `docs`. I'll change it.

PR #5, opened 2015-11-12
[WIP] ref(controller): tons of breaking changes

bacongobbler:
List of changes are as follows:

 - removed legacy web UI
 - added ADMIN_ENABLED flag for /admin endpoint
 - replaced confd with kubernetes service discovery (PoC)
 - removed LDAP auth (for now)
 - removed etcd data migrations (v2 = code cleanup)
 - docker client now reads DOCKER_URL via service discovery
 - refactored rootfs to reflect final filesystem layout

TODO:

 - remove database schema migrations, start from clean slate
 - fix up scheduler tests to use scheduler.mock
 - add back LDAP auth if user opts in
 - fix tests
 - m0ar cleanup

Main motivation for removing confd entirely is so that the controller can take over PID 1 after we're done all the database migrations. That way we don't require all that `trap` logic. However that means we won't have a way to configure LDAP auth on the fly, which may or may not be important.

rebased on top of #2


bacongobbler:
closing as it's grown **way** too big for any person to reasonably review. Even I'm having a hard time reviewing my own changes!

I'll break this up into smaller pieces.

PR #10, opened 2015-11-13
ref(controller): use kubernetes service discovery, remove publish loop

### LINKED ISSUE #8: replace confd templating with kubernetes service discovery
In v1 we published the components using a publishing loop, which would push host/port information into etcd. We would then template out the host/port using confd, and that was our service discovery. In v2, kubernetes has service discovery baked in so this is no longer necessary. However, we still need confd for things like LDAP auth support.

bacongobbler:
closes #8

### DIFF
--- a/rootfs/bin/boot
+++ b/rootfs/bin/boot
@@ -53,7 +53,6 @@ etcd_set_default secretKey "${DEIS_SECRET_KEY:-$(openssl rand -base64 64 | tr -d
 etcd_set_default builderKey "${DEIS_BUILDER_KEY:-$(openssl rand -base64 64 | tr -d '\n')}"
 etcd_set_default registrationMode "enabled"
 etcd_set_default webEnabled 0
-etcd_set_default unitHostname default
 
 # safely create required keyspaces
 etcd_safe_mkdir /deis/domains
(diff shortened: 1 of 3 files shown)


rootfs/bin/boot
```diff
@@ -48,12 +48,8 @@ function etcd_safe_mkdir {
 	set -e
 }
 
-etcd_set_default protocol "${DEIS_PROTOCOL:-http}"
 etcd_set_default secretKey "${DEIS_SECRET_KEY:-$(openssl rand -base64 64 | tr -d '\n')}"
 etcd_set_default builderKey "${DEIS_BUILDER_KEY:-$(openssl rand -base64 64 | tr -d '\n')}"
-etcd_set_default registrationMode "enabled"
```
mboersma: The controller still supports `registrationMode` options through etcd, so doesn't this setup line need to remain here?

rootfs/deis/settings.py
```diff
@@ -284,8 +284,8 @@
 TEST_RUNNER = 'api.tests.SilentDjangoTestSuiteRunner'
 
 # etcd settings
-ETCD_HOST = os.environ.get('DEIS_ETCD_1_SERVICE_HOST', '127.0.0.1')
-ETCD_PORT = os.environ.get('DEIS_ETCD_1_SERVICE_PORT_CLIENT', 4001)
+ETCD_HOST = os.environ.get('ETCD_SERVICE_HOST', '127.0.0.1')
```
mboersma: I've been testing with deis/etcd, and `DEIS_ETCD_1_SERVICE_HOST` and friends are the environment variables it makes available, so I don't think this will work without some changes over there.

rootfs/bin/boot
bacongobbler: oops!

rootfs/bin/boot
```diff
@@ -61,29 +57,10 @@ etcd_safe_mkdir /deis/platform
 etcd_safe_mkdir /deis/scheduler
 etcd_safe_mkdir /deis/services
 
-# HACK: set up keys for database
-etcd_safe_mkdir /deis/database
-ETCD_PATH=/deis/database etcd_set_default engine postgresql_psycopg2
-ETCD_PATH=/deis/database etcd_set_default name deis
-ETCD_PATH=/deis/database etcd_set_default user deis
-ETCD_PATH=/deis/database etcd_set_default password changeme123
-ETCD_PATH=/deis/database etcd_set_default host "${DEIS_DATABASE_SERVICE_HOST:-127.0.0.1}"
```
mboersma: The test database that `make kube-create-database` installs exports these environment variables. We don't want to use it long-term, but I think these are the appropriate env vars to expect.

rootfs/deis/settings.py
bacongobbler: I was considering changing them but I agree let's just do that later

rootfs/bin/boot
bacongobbler: I changed deis/settings.py to DEIS_DATABASE_SERVICE_HOST

manifests/deis-workflow-rc.yml
```diff
@@ -15,7 +15,7 @@ spec:
     spec:
       containers:
       - name: deis-workflow
-        image: deis/workflow:latest
+        image: leeroy.bacongobbler.com/deis/workflow:git-c2992c8
```
mboersma: We should leave this at deis/workflow:latest for now--that's the one that is being built from latest master. If you do `make docker-push` it will replace the image field for you in the manifest that is installed with `make kube-create`.

bacongobbler: oops, thanks for catching this.

rootfs/bin/boot
```diff
@@ -131,28 +111,4 @@ confd -node "$ETCD" --confdir /app --log-level error --interval 5 &
...
-		etcdctl --no-sync -C "$ETCD" set "$ETCD_PATH/host" "$HOST" --ttl "$ETCD_TTL" >/dev/null
-		etcdctl --no-sync -C "$ETCD" set "$ETCD_PATH/port" "$EXTERNAL_PORT" --ttl "$ETCD_TTL" >/dev/null
-		sleep $((ETCD_TTL/2)) # sleep for half the TTL
-	done
-
-	# if the loop quits, something went wrong
-	exit 1
-
-fi
-
```
carmstrong: So glad to get rid of that code.

carmstrong:
Really wish we could get these components rewritten in Go, but the removed logic was the clunkiest anyhow.

bacongobbler:
little steps :)

PR #12, opened 2015-11-14
fix(confd_settings): remove old etcd dirs

mboersma:
These should have been removed from the confd config in #10; this gets controller running again.


carmstrong:
Did #10 work without these changes, @bacongobbler? Unsure how/why it passed

bacongobbler:
Travis just runs the unit tests, not travis. Up until this point we haven't been able to get deis/workflow running on kubernetes until @mboersma did the grunt work in deis/charts#30.

PR #25, opened 2015-11-20
fix(.travis.yml): remove errant list dash

mboersma:
Travis CI's yaml validator pointed out this error in the addons: section. Also there was a typo in the vendor cache directive, and re-using Travis' canonical location for a python virtualenv seems to have backfired, so I moved it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,7 +10,7 @@ cache:
   directories:
     - $HOME/.cache/pip
     - $GOPATH/src/github.com/deis/workflow/vendor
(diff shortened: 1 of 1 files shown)


sgoings:
Just another reminder that we probably want to roll this logic into the build system itself at some point so that people can benefit from this scripting and not just Travis. :smile:

mboersma:
> roll this logic into the build system itself at some point

I totally agree. This is just to get dogfooding of Deis v2 rolling along. :poodle:

PR #26, opened 2015-11-23
feat(client): identify app by apptype

smothiki:
Identifies which type of app based on procfile or dockerfile present in the directory 
if dockerfile port is 2224 if procfile port is 2222.
Needs respective changes in the router

### DIFF
--- a/client/cmd/apps.go
+++ b/client/cmd/apps.go
@@ -18,15 +18,13 @@ import (
 )
 
 // AppCreate creates an app.
+
 func AppCreate(id string, buildpack string, remote string, noRemote bool) error {
+	apptype := "buildpack"
 	c, err := client.New()
-	if err != nil {
-		return err
-	}
 
 	fmt.Print("Creating Application... ")
 	quit := progress()
-	app, err := apps.New(c, id)
 
 	quit <- true
 	<-quit
@@ -35,8 +33,17 @@ func AppCreate(id string, buildpack string, remote string, noRemote bool) error 
 		return err
 	}
 
-	fmt.Printf("done, created %s\n", app.ID)
+	if _, err := os.Stat("Dockerfile"); err != nil {
+		if os.IsNotExist(err) {
+			fmt.Println("Dockerfile not present creating a heroku app")
+		}
 
+	} else {
+		apptype = "dockerfile"
+		fmt.Println("Dockerfile detected, creating a docker app")
+	}
+	app, err := apps.New(c, apptype, id)
+	fmt.Printf("done, created %s\n", app.ID)
 	if buildpack != "" {
 		configValues := api.Config{
 			Values: map[string]interface{}{
@@ -47,9 +54,8 @@ func AppCreate(id string, buildpack string, remote string, noRemote bool) error 
 			return err
 		}
 	}
-
 	if !noRemote {
-		if err = git.CreateRemote(c.ControllerURL.Host, remote, app.ID); err != nil {
+		if err := git.CreateRemote(c.ControllerURL.Host, remote, app.ID); err != nil {
 			if err.Error() == "exit status 128" {
 				fmt.Println("To replace the existing git remote entry, run:")
 				fmt.Printf("  git remote rename deis deis.old && deis git:remote -a %s\n", app.ID)
--- a/client/controller/api/apps.go
+++ b/client/controller/api/apps.go
@@ -5,6 +5,7 @@ type App struct {
 	Created string `json:"created"`
 	ID      string `json:"id"`
 	Owner   string `json:"owner"`
+	Apptype string `json:"apptype"`
 	Updated string `json:"updated"`
 	URL     string `json:"url"`
 	UUID    string `json:"uuid"`
@@ -12,7 +13,8 @@ type App struct {
 
 // AppCreateRequest is the definition of POST /v1/apps/.
 type AppCreateRequest struct {
-	ID string `json:"id,omitempty"`
+	ID      string `json:"id,omitempty"`
+	Apptype string `json:"apptype,omitempty"`
 }
 
 // AppUpdateRequest is the definition of POST /v1/apps/<app id>/.
--- a/client/controller/models/apps/apps.go
+++ b/client/controller/models/apps/apps.go
@@ -27,12 +27,12 @@ func List(c *client.Client, results int) ([]api.App, int, error) {
 }
 
 // New creates a new app.
-func New(c *client.Client, id string) (api.App, error) {
+func New(c *client.Client, apptype, id string) (api.App, error) {
 	body := []byte{}
 
 	var err error
 	if id != "" {
-		req := api.AppCreateRequest{ID: id}
+		req := api.AppCreateRequest{ID: id, Apptype: apptype}
 		body, err = json.Marshal(req)
 
 		if err != nil {


client/cmd/apps.go
```diff
@@ -7,7 +7,7 @@ import (
 	"strings"
 	"time"
 
-	"github.com/deis/pkg/prettyprint"
+	"github.com/deis/deis/pkg/prettyprint"
```
arschles: can you use `github.com/deis/pkg/prettyprint` instead. it pulls into far fewer dependencies and still does what you need.

client/cmd/apps.go
```diff
@@ -35,8 +34,18 @@ func AppCreate(id string, buildpack string, remote string, noRemote bool) error
 		return err
 	}
 
-	fmt.Printf("done, created %s\n", app.ID)
+	if _, err := os.Stat("Dockerfile"); err != nil {
+		if os.IsNotExist(err) {
```
arschles: wouldn't it be simpler to test for existence:

```go
if _, err := os.Stat("Dockerfile"); err == nil { 
  apptype="dockerfile"
  //...
}
```

smothiki:
@arschles @technosophos  myself have  just discussed maintaining a single git server which handles both type of apps . which means this PR might get obsolete. for now keeping it open as a backup plan

client/cmd/apps.go
smothiki: yeah that's right way to do

arschles:
@smothiki ok, can you please write up details on how the new server will work, as well as timeline on when you'd like to have it done by. I am just starting to write a proposal in https://github.com/deis/builder for

1. merging the dockerfile and buildpack build functionalities into a single builder, and 
2. rewriting most of the shell scripts in Go

technosophos:
Here's what we whiteboarded really quickly:

1. There is one git server. It's job is to git-receive, send a source archive to the object store, and then figure out what kind of "builder" backend to start.
2. There are multiple "builder"-like services:
    - slugbuilder: reads the source tgz and generates a built slug.tgz
    - dfbuilder (or whatever we name it): reads the source tgz, does a docker build, and pushes the image to a docker registry

Ideally, we could later plug in other build types, like an rkt builder or a Garden builder or something.

The Git gateway should allow...
- turning on and off individual backends (e.g. I can enable slugbuilder, but disable dfbuilder)
- configuring the S3 backend

smothiki:
changed according to the new builder design. Now client just identifies app type

smothiki:
Requires changes to deis controller

client/cmd/apps.go
```diff
@@ -18,15 +18,13 @@ import (
 )
 
 // AppCreate creates an app.
+
```
mboersma: This blank line detaches the comment from its func and should be removed.

smothiki: sure

client/cmd/apps.go
```diff
@@ -18,15 +18,13 @@ import (
...
 func AppCreate(id string, buildpack string, remote string, noRemote bool) error {
+	apptype := "buildpack"
 	c, err := client.New()
-	if err != nil {
-		return err
-	}
```
mboersma: This error handling should be left intact, otherwise we're ignoring something potentially useful.

client/cmd/apps.go
```diff
@@ -18,15 +18,13 @@ import (
...
 
 	fmt.Print("Creating Application... ")
 	quit := progress()
-	app, err := apps.New(c, id)
 
 	quit <- true
 	<-quit
```
arschles: not familiar with this codebase, but want to make sure. is another goroutine listening on `quit`?

mboersma:
I'm going to mark this as "in progress" until it is passing client unit tests and includes the related controller changes (and tests :green_heart:).

bacongobbler:
Personally I think `appType` should be a part of the Build rather than the app as we do here: https://github.com/deis/deis/blob/master/controller/api/models.py#L742-L745

Basically I can see this real-world use cases happening which can cause these edge cases to occur:

 - I `deis create` with `--no-remote`, intending to use `deis pull` strictly (this is a documented workflow)
 - I `deis create` a newly created git repository without any files in it, intending to add application data later
 - I want to switch from buildpack types to Dockerfile types by introducing a Dockerfile to the app. In v1 this was handled transparently at build time, then the user had to run `deis scale web=0 cmd=1` to migrate from one type to the other.

Essentially I'm concerned about the workflow that this kind of attribute would introduce and I believe that this is an artifact of the build, not the app itself.

arschles:
@bacongobbler is `appType` something we can 'rolling migrate' later?

for example, let's say we release a v2.1 of the controller which infers the `appType` of your build based on the `appType` listed under the app.

bacongobbler:
> @bacongobbler is appType something we can 'rolling migrate' later?

If we can do it in a backwards compatible way without breaking v2.0 API endpoint/behaviour then yes. We've had users create third party clients for the API in their own workspaces, so as long as a "rolling migration" won't break the API then that's fine on that end. In other words, we can add new behaviour to the API once we hit 2.0, but we cannot remove or change existing behaviour.

Thinking about this further, consider the following effect this will make on the user...

 - I create an app, which defaults to "buildpack" appType
 - I create a few releases, then decide to switch to "Dockerfile" types. The buildpacks were great getting started, but now I want to slim down the Docker images by supplying my own.
 - I create a few more releases, but then want to roll back to a release that previously was of the "buildpack" type again. I missed the simplicity of it.

To do that, I need to one thing in Deis v1: `deis scale web=0 cmd=1`. To switch back, `deis rollback v3 && deis scale cmd=0 web=1` (this could've been automated with #2861 but we opted out to not be overly complex). With this change, we introduce even more to this workflow:

```
deis apps:update --type=dockerfile
deis scale web=0 cmd=1
deis rollback v3
deis apps:update --type=buildpack
deis scale web=1 cmd=0
```

Just seems to be a lot of boilerplate just because the builder can't handle this logic itself (which it should, IMO). This just doesn't "feel" like the right direction when the builder already gets all this information just by reading the incoming source code from a `git push`.

Sorry to be a worry wart on all this... I'm pretty fanatical about how data should be modeled, especially when it comes to 12 Factor. If there's a solid use case/documentation for each of the workflows I described above then I guess I'd be fine with this. My main gripes are rooted from the idea that this changes the app workflow in a "bad" way, further separating us from Heroku's simplicity and "behind the scenes" magic.

smothiki:
this can be made independent of client as well. Whenever there is a git push builder has the capability to identify the kind of app and build the artifact accordingly . If can change the generate buildhook API. Then the client User experience stays intact.

technosophos:
I'm a little confused about which thing is the real cause of concern. There seem to be several questions here:

1. Should application type be determined client or server side?
2. Must the user specify a type, or can it be inferred?
3. What is the workflow behind converting an app from one type to another?

These questions seem to be more or less discrete, right? (For example, I can detect app type on the client, but infer it... or I can detect on the server side and infer it).

I strongly agree with @bacongobbler's desire to keep this as straightforward for the user as possible. But I'm not sure that requires that we do all of this on the controller. I also think it should be possible for the user to explicitly state that a thing is a buildpack or a dockerfile build _when necessary_ (but by default, maybe do something sensible).

client/cmd/apps.go
Joshua-Anderson: :+1: Without this creating an app without logging in would cause an panic.

arschles:
I can't contribute much right now because I'm still becoming familiar with the deis CLI and the controller. It seems that, in a v2.x version, we can start ignoring the app-wide `appType` and port the `os.Stat("Dockerfile")` logic to the server so inference happens server-side, and the workflow for changing app type between builds is minimized (scale down, `git push`, scale up I think?)

I'm probably missing something here, but hopefully this continues discussion.

bacongobbler:
> I'm a little confused about which thing is the real cause of concern.

The real cause of concern here is the client workflow changes to accomodate this. The client workflow should dictate how the components work together, not the other way around.

> These questions seem to be more or less discrete, right?

Yes. They are more or less all related to each other.

> I strongly agree with @bacongobbler's desire to keep this as straightforward for the user as possible. But I'm not sure that requires that we do all of this on the controller.

We've been able to strip this information during the build without having to touch the controller at all. The information on the controller side right now is all strictly metadata, save for [determining the default process type](https://github.com/deis/deis/blob/master/controller/api/models.py#L502). :)

> I also think it should be possible for the user to explicitly state that a thing is a buildpack or a dockerfile build when necessary (but by default, maybe do something sensible).

That's certainly an interesting thought. That could help solve issues for users like deis/deis#4749!

> we can start ignoring the app-wide appType

Unfortunately that's a breaking change and cannot happen post-v2. We'd have to rely on both server-side inference as well as this app-wide appType attribute.

> this can be made independent of client as well. Whenever there is a git push builder has the capability to identify the kind of app and build the artifact accordingly. If can change the publish release API. Then the client User experience stays intact.

That kind of goes along what I was saying about moving the `appType` attribute to the Build rather than the App. That way it's part of the release, but it's not application-specific. We already do this in places with the `dockerfile`, `procfile` and `sha` attributes.

bacongobbler:
@smothiki is this ok to close now that we're moving in another direction?

smothiki:
closing refer https://github.com/deis/workflow/pull/40

PR #34, opened 2015-11-25
fix(settings): default to empty host environment variable

bacongobbler:
If you do not supply a host variable, Then psycopg2 will connect using a Unix socket in the same
manner as psql. Surprisingly, when you specify a HOST, psycopg2 will connect with TCP/IP, which
requires a password. This broke my environment with the following when running `make test-unit`:

        OperationalError: fe_sendauth: no password supplied

See: http://stackoverflow.com/a/23871618

### DIFF
--- a/rootfs/deis/settings.py
+++ b/rootfs/deis/settings.py
@@ -330,7 +330,7 @@ DATABASES = {
         'NAME': os.environ.get('DEIS_DATABASE_NAME', 'deis'),
         'USER': os.environ.get('DEIS_DATABASE_USER', ''),
         'PASSWORD': os.environ.get('DEIS_DATABASE_PASSWORD', ''),
-        'HOST': os.environ.get('DEIS_DATABASE_SERVICE_HOST', '127.0.0.1'),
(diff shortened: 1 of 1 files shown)


mboersma:
This change is fine with me, but note that it doesn't help when running PostgreSQL in a container via `make postgres`: I still have to do `export DEIS_DATABASE_SERVICE_HOST=192.168.99.100` to get unit tests to work.

bacongobbler:
Ha! Without this change I've been having to use `export DEIS_DATABASE_SERVICE_HOST=""` since I'm using postgres running directly on my machine instead of a dockerized postgres. It appears that no matter which way it's decided, one will have to work around it.

Was it easier for you to run tests locally before the envvar change? I'm curious to hear how it worked for you. Perhaps there is a better way to approach this (if any)

mboersma:
> Was it easier for you to run tests locally before the envvar change?

No, this doesn't change that: I still have to set `DEIS_DATABASE_SERVICE_HOST` as before. But this does simplify testing when running PostgreSQL natively, so LGTM.

PR #36, opened 2015-11-25
fix(scheduler): use app name when creating services

### LINKED ISSUE #17: question(controller): is possible to add restrictions for application names in V2?
I opened this here https://github.com/deis/deis/issues/4014
(now the created services follow the naming `app-\d{5}`)

### LINKED ISSUE #32: k8s.py creates a service with a randomized name for the app
After deploying an app:

```
[vagrant@kubernetes-minion-1 ~]$ deis pull deis/example-go -a lively-dragster
Creating build... done
```

I see that the RC was named properly, but not the service:

```
><> k get rc --namespace=lively-dragster
CONTROLLER               CONTAINER(S)          IMAGE(S)                                 SELECTOR                                   REPLICAS
lively-dragster-v2-cmd   lively-dragster-cmd   10.247.252.167:5000/lively-dragster:v2   name=lively-dragster,type=cmd,version=v2   1
><> k get svc --namespace=lively-dragster
NAME       LABELS                 SELECTOR                        IP(S)          PORT(S)
app-7216   name=lively-dragster   name=lively-dragster,type=cmd   10.247.84.53   80/TCP
```

We should have the service name as `lively-dragster`, or else other apps in the same namespace (once we introduce something like `deis teams` into the mix) won't be able to discover each other in a sane way. I'll write up a proposal so we can discuss what needs to change in order to have this work.

bacongobbler:
When we eventually allow applications to be grouped together in the same namespace, we'll want to
have the service names discoverable to other instances in the same namespace. Having a name like
"app-1234" was created so we could get over the limitation of 24 characters for service names. Now
that we're in v2, we can just restrict app names to 24 characters.

The app name validation regex was not changed because the docker repository name regex is a subset of the kubernetes service name regex, so we still need to keep ourselves restricted to that name regex so we can still push those images to the internal registry.

closes #17 
closes #32

### DIFF
--- a/rootfs/api/models.py
+++ b/rootfs/api/models.py
@@ -153,7 +153,7 @@ class App(UuidAuditedModel):
     """
 
     owner = models.ForeignKey(settings.AUTH_USER_MODEL)
-    id = models.SlugField(max_length=64, unique=True, default=select_app_name,
+    id = models.SlugField(max_length=24, unique=True, default=select_app_name,
                           validators=[validate_id_is_docker_compatible,
                                       validate_reserved_names])
     structure = JSONField(default={}, blank=True, validators=[validate_app_structure])
--- a/rootfs/api/south_migrations/0025_auto__chg_field_app_id.py
+++ b/rootfs/api/south_migrations/0025_auto__chg_field_app_id.py
@@ -1,0 +1,159 @@
+# -*- coding: utf-8 -*-
+from south.utils import datetime_utils as datetime
+from south.db import db
+from south.v2 import SchemaMigration
+from django.db import models
+
+
+class Migration(SchemaMigration):
+
+    def forwards(self, orm):
+
+        # Changing field 'App.id'
+        db.alter_column(u'api_app', 'id', self.gf('django.db.models.fields.SlugField')(unique=True, max_length=24))
+
+    def backwards(self, orm):
+
+        # Changing field 'App.id'
+        db.alter_column(u'api_app', 'id', self.gf('django.db.models.fields.SlugField')(max_length=64, unique=True))
+
+    models = {
+        u'api.app': {
+            'Meta': {'object_name': 'App'},
+            'created': ('django.db.models.fields.DateTimeField', [], {'auto_now_add': 'True', 'blank': 'True'}),
+            'id': ('django.db.models.fields.SlugField', [], {'default': "'unisex-mainsail'", 'unique': 'True', 'max_length': '24'}),
+            'owner': ('django.db.models.fields.related.ForeignKey', [], {'to': u"orm['auth.User']"}),
+            'structure': ('json_field.fields.JSONField', [], {'default': '{}', 'blank': 'True'}),
+            'updated': ('django.db.models.fields.DateTimeField', [], {'auto_now': 'True', 'blank': 'True'}),
+            'uuid': ('api.fields.UuidField', [], {'unique': 'True', 'max_length': '32', 'primary_key': 'True'})
+        },
+        u'api.build': {
+            'Meta': {'ordering': "[u'-created']", 'unique_together': "((u'app', u'uuid'),)", 'object_name': 'Build'},
+            'app': ('django.db.models.fields.related.ForeignKey', [], {'to': u"orm['api.App']"}),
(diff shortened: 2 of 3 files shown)


rootfs/scheduler/k8s.py
```diff
@@ -373,9 +372,6 @@ def _get_service(self, name, namespace):
         return resp.status_code, resp.text, resp.reason
 
     def _create_service(self, name, app_name, app_type):
-        random.seed(app_name)
-        app_id = random.randint(1, 100000)
-        appname = "app-" + str(app_id)
```
mboersma: Goodbye to this evil code, it will not be missed. :-)

rootfs/api/models.py
```diff
@@ -168,7 +168,7 @@ class App(UuidAuditedModel):
     """
 
     owner = models.ForeignKey(settings.AUTH_USER_MODEL)
-    id = models.SlugField(max_length=64, unique=True, default=select_app_name,
+    id = models.SlugField(max_length=24, unique=True, default=select_app_name,
```
mboersma: This probably needs to add a Django South migration for the schema change. (I tried to test it but was able to create app names much longer than 24 chars since the database schema was defined earlier with 64 chars.)

bacongobbler: forgot about that. Good catch!

aledbf:
Code LGTM.

(changing the schema)
@bacongobbler thanks for this change.

rootfs/scheduler/k8s.py
krancour: Ha.  I was looking at this the other day and not liking it either.  Glad to see it go.

bacongobbler:
done, now with schema migrations!

smothiki:
along with restricting App name to 24 character also App name should be compatible with C variable standards.
Does the regex checks whether App names follow C variable format ?

bacongobbler:
No, but it is compatible. We still use the same regex as the docker registry. Since the docker registry regex (only lowercase, alphanumeric, dashes allowed) is a subset of kubernetes's regex, we must primarily comply with that standard, or else we will not be able to push images to the internal registry. Since it is a subset of the standard, we are automatically compatible with it.

rootfs/api/south_migrations/0025_auto__chg_field_app_id.py
```diff
@@ -0,0 +1,159 @@
...
+        u'contenttypes.contenttype': {
+            'Meta': {'ordering': "('name',)", 'unique_together': "(('app_label', 'model'),)", 'object_name': 'ContentType', 'db_table': "'django_content_type'"},
+            'app_label': ('django.db.models.fields.CharField', [], {'max_length': '100'}),
+            u'id': ('django.db.models.fields.AutoField', [], {'primary_key': 'True'}),
+            'model': ('django.db.models.fields.CharField', [], {'max_length': '100'}),
+            'name': ('django.db.models.fields.CharField', [], {'max_length': '100'})
+        }
+    }
+
+    complete_apps = ['api']
```
arschles: can you put a newline here?

arschles:
@bacongobbler this has a `needs manual testing` label on it. does that mean I should do so?

rootfs/api/south_migrations/0025_auto__chg_field_app_id.py
bacongobbler: sure. for some reason south/django doesn't like adding new lines to migrations. I'll add that.

bacongobbler:
Anything scheduler-related can't be covered by unit tests until we add some mock tests to a kubernetes endpoint, so for now we have to test these manually, though since myself and @mboersma have I guess this can be omitted.

rootfs/api/south_migrations/0025_auto__chg_field_app_id.py
bacongobbler: done

arschles:
ok. looks like I'll be testing this functionality a bit this week anyway. `LGTMe`-ed

bacongobbler:
lol, github labels are playing with us :P

arschles:
haha. race condition :allthethings:

PR #40, opened 2015-12-02
WIP[build]: identify app types and select respective RC template

smothiki:
according to the publish release will identify if app is using dockerfile or not and select specific image and RC template.

requires https://github.com/deis/builder/pull/12
https://github.com/deis/builder/pull/9
issue with deis:config set which will not deal with changing images rather updates RC templates to update environment variables


rootfs/api/models.py
```diff
@@ -203,7 +203,7 @@ def _get_command(self, container_type):
             # and run arbitrary commands on the host
             # FIXME: remove slugrunner's hardcoded entrypoint
             release = self.release_set.latest()
-            if release.build.dockerfile or not release.build.sha:
+            if release.build.dockerfile:
```
bacongobbler: I think this conditional is still required for `deis pull` workflows, since it won't have a Dockerfile or a sha present.

rootfs/api/models.py
```diff
@@ -326,7 +326,10 @@ def scale(self, user, structure):  # noqa
     def _scale_containers(self, scale_types, to_remove):
         release = self.release_set.latest()
         for scale_type in scale_types:
-            image = release.image
+            if release.build.dockerfile:
+                image = release.image
+            else:
+                image = release.image.split(":")[0]+":git-"+release.build.sha
```
bacongobbler: this is the kind of information that builder should push up to the controller on a push hook (I believe that's how it's done today IIRC). Otherwise this will break `deis pull`.

rootfs/api/models.py
```diff
@@ -476,7 +479,10 @@ def deploy(self, user, release):
 
     def _deploy_app(self, scale_types, release, existing):
         for scale_type in scale_types:
-            image = release.image
+            if release.build.dockerfile:
+                image = self.release.image
+            else:
+                image = self.release.image.split(":")[0]+":git-"+self.release.build.sha
```
bacongobbler: same here, this breaks `deis pull`

rootfs/api/models.py
```diff
@@ -881,7 +891,8 @@ def publish(self, source_version='latest'):
             source_image = "{}:{}".format(source_image, source_tag)
         # If the build has a SHA, assume it's from deis-builder and in the deis-registry already
         deis_registry = bool(self.build.sha)
-        publish_release(source_image, self.config.values, self.image, deis_registry)
+        if self.build.dockerfile:
+            publish_release(source_image, self.config.values, self.image, deis_registry)
```
bacongobbler: why wouldn't we want to publish a release if no dockerfile is present? Logically speaking, any build we create should be published.

smothiki: publish_release only changes the tag of the image by pulling the existing image tagging it to verison instead of git sha and pushes the image back to registry.
For buildpack type apps. the image is always slugrunner  so this step is not necessary.

rootfs/scheduler/k8s.py
```diff
@@ -323,18 +376,28 @@ def _create_rc(self, name, image, command, **kwargs):
...
             "id": app_name,
             "appversion": kwargs.get("version", {}),
             "version": self.apiversion,
-            "image": self.registry + "/" + image,
+            "image": imgurl,
             "num": kwargs.get("num", {}),
             "containername": container_name,
             "type": app_type,
         }
+<<<<<<< e29708e9bc307a8bfcdc4c2bc6bc25a5dd9b85e7
```
mboersma: Looks like a merge or rebase error crept in here.

smothiki:
closing this PR in favor of https://github.com/deis/workflow/pull/95

PR #41, opened 2015-12-02
fix(make): use DEIS_REGISTRY as an env var for consisteny

helgi:
`deis/router` is doing it and seems more consistent with our naming policies


bacongobbler:
I think we also need to make the change elsewhere: https://github.com/deis/workflow/blob/master/_scripts/ci/deploy.sh#L9-L11

helgi:
Ah yeah, I also updated other places that needed it

bacongobbler:
related: #39

PR #42, opened 2015-12-02
ref(k8s): tidy up Python logic in k8s scheduler

helgi:
Consolidate code and unify a bit more under single quote (as much of the code already is).
Rename raise_err to error to be clearer about its functionality

Clean up logic in _delete_pod and have waiting work in it.

More to come, split out to simplify review

### DIFF
--- a/rootfs/scheduler/k8s.py
+++ b/rootfs/scheduler/k8s.py
@@ -102,10 +102,17 @@ MATCH = re.compile(
     r'(?P<app>[a-z0-9-]+)_?(?P<version>v[0-9]+)?\.?(?P<c_type>[a-z-_]+)')
 
 
-def raise_err(resp, errmsg, *args):
+def error(resp, errmsg, *args):
     errmsg = errmsg.format(*args)
     errmsg = "failed to {}: {} {}\n{}".format(errmsg, resp.status_code, resp.reason, resp.json())
     raise RuntimeError(errmsg)
+
+
+def unhealthy(status_code):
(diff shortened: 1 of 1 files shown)


bacongobbler:
most of it is just switching around names and adding whitespace, but the chained `replace` calls and `unhealthy` function calls are a nice cleanup. +1

helgi:
The reason for the whitespaces was to have control structures separated out instead having them smashed together in one long function.  Def a personal preference on that one but nothing essential :)

`_delete_pod` needs the closest look since the logic in it looked broken so I used my best judgement to fix it.

bacongobbler:
nope. The logic for both is the exact same AFAICT. It just looks awkward with the way it was set up. Your implementation is more clear as to its intention.

rootfs/scheduler/k8s.py
```diff
@@ -145,8 +153,8 @@ def _get_old_rc(self, name, app_type):
                 break
         if exists:
             return prev_rc
-        else:
```
mboersma: You don't like the explicit `if` -- `else`? I've been writing a lot of Go code so this feels right, but not sure which is more pythonic.

Ok, I've already spent too much time typing about this detail. :+1:

helgi: The way I've always approached it (can't remember if it is in K&R or similar style guides) is that you are already returning a default value at the end, no need to `else` those, `exists` is the special condition, etc etc :)

PR #46, opened 2015-12-04
feat(travis): hook up docs to travis

### LINKED ISSUE #28: hook up docs to travis
Right now the docs are not covered by travisCI, but it certainly would be useful to at least run a [markdown linter](https://github.com/mivok/markdownlint) on the docs.

bacongobbler:
closes #28

### DIFF
--- a/docs/Makefile
+++ b/docs/Makefile
@@ -16,7 +16,7 @@ clean:
 	rm -rf $(BUILDDIR)/*
 
 deps:
-	pip install --disable-pip-version-check -r requirements.txt
+	pip install -r requirements.txt
 
 test: build
 	grep -q '<h1 id="welcome">Welcome</h1>' _build/html/index.html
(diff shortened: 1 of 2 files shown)


docs/Makefile
```diff
@@ -16,7 +16,7 @@ clean:
 	rm -rf $(BUILDDIR)/*
 
 deps:
-	pip install --disable-pip-version-check -r requirements.txt
```
mboersma: Why was this removed? Without that flag you get the annoying upgrade message all the time.

bacongobbler: just sticking to being consistent with lines 31 and 32 in .travis.yml. I can revert this if you'd prefer :)

mboersma: Oh I see. Now I'm torn between improving consistency and reducing pointless log messages...

Meh, this is trivial either way, LGTM.

PR #48, opened 2015-12-07
ref(scheduler): split up k8s.py and make it more modular

helgi:
Splits each resource type into its own class, creates a KubeHTTPClient that handles HTTP interactions
and moves the k8s code up into `__init__`. Separation of responsibility in the code is stronger as the
scheduler code the Deis API interfaces with doesn't touch HTTP and individual resource classes deal
very little with it as well.

Many HTTP interactions have been consolidated and by default return a RunTimeError Exception if a
non-2xx code is returned unless the `check=False` flag is passed


helgi:
There are still a few TODO items in there and I'm wanting to rip out `logging` from all resource type classes and have it live all the way up in the scheduler, at least for the type of logging we've been doing so far.

Still things to do here but it's generally ready for feedback. I don't have a working `deis/etcd` to actually properly test this tho

rootfs/scheduler/base.py
```diff
@@ -0,0 +1,8 @@
+from django.conf import settings
+from .client import KubeHTTPClient
+
+
+class Base(object):
```
helgi: Ingesting `settings` directly here - we were anyway doing that in a different part of the code instead of using what's passed in from the model but at the end of the day we may want to abstract the passing out.

helgi:
This is ready for reviewing now. Note it does lack unit tests but that's a different ticket

PR #51, opened 2015-12-07
ref(models): remove unused functions

helgi:
These functions are not used anywhere

* https://github.com/deis/workflow/search?utf8=%E2%9C%93&q=validate_comma_separated
* https://github.com/deis/workflow/search?utf8=%E2%9C%93&q=validate_domain&type=Code


mboersma:
LGTM, but I think this is blocked in CI until #52 is merged. I did some `git grep`s and ran `make test` locally.

PR #57, opened 2015-12-08
fix(migrations): remove etcd migration code from boot

helgi:
Needed for boot script

### DIFF
--- a/rootfs/bin/boot
+++ b/rootfs/bin/boot
@@ -64,15 +64,6 @@ etcd_safe_mkdir /deis/services
 etcd_safe_mkdir /deis/platform
 ETCD_PATH=/deis/platform etcd_set_default domain localhost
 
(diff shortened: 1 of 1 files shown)


bacongobbler:
let's just remove the etcd migrations part from the boot script instead :)

helgi:
Could do that, no chance we'd ever want to use it for something? :P k8s migrations instead?

bacongobbler:
nah I don't think so if we're wanting to eventually move away from etcd in Deis: https://github.com/deis/etcd/issues/10

helgi:
Ditched it

bacongobbler:
good bye etcd migrations; you won't be missed :balloon: :balloon:

bacongobbler:
merging as this is a showstopper and we have 2 LGTMs at this point. Thanks @helgi!

PR #58, opened 2015-12-09
feat(*): change format of one off commands response and bump api to v2

### LINKED ISSUE #59: Proposal: Change format of `/v1/apps/<app id>/run`
Currently run returns a dynamic array (Ewww!) `[0, "hi\n"]`. The original docs docs say that run should return `{"rc": 0, "output": "hi"}`, which seems much more sensible to me.

I would like to propose to change the response of run to the format suggested to the docs:

```json
{
  "rc": <return code>, 
  "output": "<command output>"
}

```

This is a backwards incompatible change, so this would probably would require API version bump to version 2.

Joshua-Anderson:
This is a port of deis/deis#3999 and issue deis/deis#3988 to workflow.

Since this is a backwards incompatible change, I bumped the api to version v2 and updated docs.
Is bumping to 2.0 ok? Is there a version number you would prefer? Should I not bump the version at all yet?

Sorry, I'm not very clued in to v2 yet, but I'd love to see this change go in because muti-type arrays don't play well with the golang client and are pretty evil :smiling_imp:, this was also the way the command was originally documented (though not implemented).

If this is merged, then the controller hooks in builder would also need to update urls to V2.

Fixes #59

### DIFF
--- a/client/controller/api/apps.go
+++ b/client/controller/api/apps.go
@@ -10,22 +10,22 @@ type App struct {
 	UUID    string `json:"uuid"`
 }
 
-// AppCreateRequest is the definition of POST /v1/apps/.
+// AppCreateRequest is the definition of POST /v2/apps/.
 type AppCreateRequest struct {
 	ID string `json:"id,omitempty"`
 }
 
-// AppUpdateRequest is the definition of POST /v1/apps/<app id>/.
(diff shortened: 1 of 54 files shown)


mboersma:
Great work--thanks @Joshua-Anderson! It does seem like the time to bump API v2.0. :rocket:

PR #62, opened 2015-12-10
ref(api): split api models into a class per file

helgi:
Allows a separate of a class per file and makes it easier to read through the code, reason with what logic belongs to which functionality and generally work with it

See:
* https://code.djangoproject.com/wiki/CookBookSplitModelsToFiles
* http://paltman.com/breaking-apart-models-in-django/


krancour:
+1 on this and the rationale for it.

mboersma:
This is failing several unit tests: https://travis-ci.org/deis/workflow/builds/95955481

helgi:
Yeah, it's the `_etcd_client` mocking and how imports are now interacting with one another. Trying to figure out a better way to do the mocking and see if there is a more sensible way to do `_etcd_client`

rootfs/api/models/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+
+    return get_etcd_client.client
+
+
+# define update/delete callbacks for synchronizing
+# models with the configuration management backend
+
+def _etcd_purge_user(**kwargs):
+    try:
+        url = '/deis/builder/users/{}'.format(kwargs['instance'].username)
```
arschles: not sure what the old code did, but does this need a check to see if `'instance'` exists in the dict?

rootfs/api/models/__init__.py
```diff
@@ -0,0 +1,78 @@
...
+# Load in all models
+from app import App  # noqa
+from build import Build  # noqa
+from certificate import Certificate  # noqa
+from config import Config  # noqa
+from container import Container  # noqa
+from domain import Domain  # noqa
+from key import Key  # noqa
+from push import Push  # noqa
+from release import Release  # noqa
```
arschles: why not put these at the top? something to do with flake?

rootfs/api/models/__init__.py
mboersma: That would be good defensive coding practice, but in this case I think [Django signals](https://docs.djangoproject.com/en/1.6/ref/signals/#post-delete) effectively guarantee that `instance` will be there.

rootfs/api/models/__init__.py
helgi: Had to put them at the bottom or python would get angry me importing `get_etcd_client` into individual models. I could have left out those imports completely but for consistency (and smaller chance of other code breaking) I left imports in there with `# noqa` for flake

rootfs/api/models/__init__.py
arschles: :+1: cool

rootfs/api/models/__init__.py
arschles: got it

rootfs/api/models/app.py
```diff
@@ -0,0 +1,528 @@
...
+        config = Config.objects.create(owner=self.owner, app=self)
+        Release.objects.create(version=1, owner=self.owner, app=self, config=config, build=None)
+
+    def delete(self, *args, **kwargs):
+        """Delete this application including all containers"""
+        try:
+            # attempt to remove containers from the scheduler
+            self._destroy_containers([c for c in self.container_set.exclude(type='run')])
+        except RuntimeError:
+            pass
```
arschles: at least log here?

again, don't know what the old code did so not a blocker if this is copied

rootfs/api/models/app.py
```diff
@@ -0,0 +1,528 @@
...
+
+            raise EnvironmentError('Error accessing deis-logger')
+
+        return r.content
+
+    def run(self, user, command):
+        """Run a one-off command in an ephemeral app container."""
+        if self.release_set.latest().build is None:
+            raise EnvironmentError('No build associated with this release to run this command')
+        # TODO: add support for interactive shell
```
arschles: ticket for this? can't tell if this is a requirement or a nice-to-have

helgi: This is from the old code - no action required for this PR

rootfs/api/models/app.py
helgi: Old code

rootfs/api/models/app.py
arschles: cool

arschles:
@helgi :+1: on splitting up that huge file

PR #65, opened 2015-12-11
ref(models): remove docker.utils from app model as it is not needed

helgi:
Looks like the image variable is needed used and thus we do not need docker lib in the App model


mboersma:
I'm surprised; this does appear to be useless code. Nice find! LGTM.

PR #70, opened 2015-12-11
Revert "ref(scheduler): split up k8s.py and make it more modular"

mboersma:
This reverts commit 218bfe25cbb6c1d80324a1be7919cf82e9c2e54c. While it passed unit tests, it introduced errors into actual app workflow that were difficult to pin down. Again, we need to resurrect the functional tests, which would have caught the problems.


arschles:
@mboersma gave a LGTM. worth testing locally before merge though? I can do so

mboersma:
@arschles more testing is always helpful, but I did test this interactively on a local cluster already.

bacongobbler:
sorry for the delays, had my head down in the sand. :shipit:!

arschles:
@mboersma ok, this may unblock me anyway. I'll test after it's built and pushed

krancour:
fwiw, sorry to see this undone.  I assume we'll come back to this eventually?

mboersma:
I hope so! But first there needs to be a component-level or functional test that would have the breakage here.

PR #71, opened 2015-12-11
feat(travis): upload docs and client to S3

### LINKED ISSUE #61: ship a v2 client for testing
now that we've merged #58, the default install docs no longer work because `curl -sSL http://deis.io/deis-cli/install.sh | sh` will install v1.12.2. What we need is for `_scripts/deploy.sh` to ship a v2 client :)

bacongobbler:
This change uses Travis CI's artifact feature to store build information
in S3. Specifically we'd like a client and a doc build so that we can test
and view changes for a specific build with these artifacts, as the rest of
the build artifacts are Docker images, uploaded to Quay.io and DockerHub.

We need to some new environment variables to the workflow job in order for this to work:

 - ARTIFACTS_KEY
 - ARTIFACTS_SECRET
 - ARTIFACTS_BUCKET
 - ARTIFACTS_S3_REGION

closes #61

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,8 @@ $ kubectl get pods --namespace=deis
 Once this is done, you can SSH into the minion running the controller and run the following:
 
 ```
-$ curl -sSL http://deis.io/deis-cli/install.sh | sh
+$ curl -sSL https://s3-us-west-1.amazonaws.com/bacongobbler/bacongobbler/workflow/19/19.1/client/deis
+$ chmod 755 deis
 $ sudo mv deis /bin
 $ kubectl get service --namespace=deis deis-workflow
 $ deis register 10.247.59.157 # or the appropriate CLUSTER_IP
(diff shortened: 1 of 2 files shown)


bacongobbler:
note that we still need to also update the relevant documentation. I'll file a followup PR for that :)

bacongobbler:
also one more thing yet to be checked out: how to symlink the latest artifact built. That way devs won't have to go check what the latest travis build on master said the download URL is for the client.

this seems relevant: https://github.com/travis-ci/artifacts#example-multiple-target-paths

bacongobbler:
I've also updated the documentation so it should be mostly correct once we get this PR up and running.

README.md
```diff
@@ -37,7 +37,8 @@ $ kubectl get pods --namespace=deis
 Once this is done, you can SSH into the minion running the controller and run the following:
 
 ```
-$ curl -sSL http://deis.io/deis-cli/install.sh | sh
+$ curl -sSL https://s3-us-west-1.amazonaws.com/bacongobbler/bacongobbler/workflow/19/19.1/client/deis
```
carmstrong: Could we host this at get-deis instead of your path?

bacongobbler: Yeah this is the intention when we get this merged. This was just temporary to get other devs unblocked from testing v2. Once this is merged and the relevant envvars are set in Travis, builds will start showing up in get-deis and I'll change this over.

bacongobbler:
keys have been added to the job. Here goes nothing!

PR #72, opened 2015-12-12
feat(client): use unix pipes with config:push

Joshua-Anderson:
This was briefly discussed as a V2 thing in deis/deis#4675 and deis/deis#4697.

I personally like it cause it's more unixy, but I understand if this is rejected because this is really a personal preference thing :smiley:.

If there is interest I will make the same change for `config:pull`.

### DIFF
--- a/client/parser/config.go
+++ b/client/parser/config.go
@@ -152,8 +152,9 @@ func configPush(argv []string) error {
 	usage := `
 Sets environment variables for an application.
 
-The environment is read from <path>. This file can be read by foreman
-to load the local environment for your app.
+This file can be read by foreman
+to load the local environment for your app. The file should be piped via
+stdin, 'deis config:push < .env', or using the --path option.
 
 Usage: deis config:push [options]
 
--- a/client/cmd/config.go
+++ b/client/cmd/config.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"bytes"
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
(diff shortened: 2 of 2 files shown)


client/parser/config.go
```diff
@@ -153,15 +153,14 @@ func configPush(argv []string) error {
...
-to load the local environment for your app.
+to load the local environment for your app. The file should be piped via
+stdin, for example 'deis config:push < .env'
 
 Usage: deis config:push [options]
 
 Options:
   -a --app=<app>
     the uniquely identifiable name for the application.
-  -p <path>, --path=<path>
```
helgi: Can't we support both?

Joshua-Anderson: There isn't any technical reason we can't. I figured it would be confusing to offer two ways to do the same thing, but I don't have any strong feeling either way.

technosophos:
@Joshua-Anderson We're getting close to freezing an Alpha of Deis v2. In your opinion, is this a risky one for us to try to merge at the last minute? Or should we wait until after Alpha.

Joshua-Anderson:
@technosophos Works on my machine (Ubuntu 15.10) :tm: :smiley:

I think merging this more of a decision if this change is desirable (some people may prefer flags over pipes). It's a small scoped and pretty easily testable change, so it wouldn't be too much work to land ( I have no idea how soon you want to release or how busy you are :smiley:)

If this doesn't make it into the alpha, can backwards compatible changes still land in V2 before the GA release?

This change also needs a complimentary change to `config:pull` to support pipes as well, but I was hanging off on that until this change was approved or rejected. So releasing one without the other would leave an odd UX where `pull` and `push` behaved differently.

client/parser/config.go
arschles: IMO we should support both, but doesn't have to be blocking for this PR

arschles:
@Joshua-Anderson I think it's wise to implement `config:pull` in this patch as well

Joshua-Anderson:
> I think it's wise to implement config:pull in this patch as well

Ok. Sounds good! I'm super busy right now, but if I have some time I'll fix it up.

Joshua-Anderson:
@arschles I updated this PR to support both pipes and command line arguments.

arschles:
@Joshua-Anderson cool. My `LGTM1` stands, but it looks like you need a rebase now

Joshua-Anderson:
:frowning: As a side effect of enabling required status (which I am totally :+1: on), every time there is a commit on master, PRs are required to update before merging (to prevent accidental test breakage), which is a little annoying.

bacongobbler:
hmm, that might be why everything needs to be rebased all the time. deis/deis does not have that enabled so that might be the bug. :)

PR #73, opened 2015-12-12
fix(k8s.py): access API server by IP address

mboersma:
Accessing the k8s api by hostname fails sometimes from within the deis/workflow pod. This changes the `requests.Session` to use the IP address and port, which is more reliable, but at least for now requires disabling verification of the local SSL cert.

Cert verification is `False` here since the IP address in the request does match any of the hostnames in the server's cert, and neither that verification nor the HTTP `Host` header can be overridden in this case in the python requests library. That is, I couldn't see any way short of monkeypatching, after reading the requests source code and overriding several classes involved.

### DIFF
--- a/rootfs/scheduler/k8s.py
+++ b/rootfs/scheduler/k8s.py
@@ -132,7 +132,10 @@ class KubeHTTPClient(AbstractSchedulerClient):
             'Authorization': 'Bearer ' + token,
             'Content-Type': 'application/json',
         }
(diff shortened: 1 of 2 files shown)


bacongobbler:
And I guess the k8s API can't be accessed by plain old http, eh? I haven't seen a way, so makes sense to me to disable SSL verification if we can't override the Host header.

PR #74, opened 2015-12-13
Controller healthcheck

mboersma:
Rebases deis/deis#4457--thanks again @benwilber! Adds a matching Kubernetes [liveness probe](http://kubernetes.io/v1.1/docs/user-guide/production-pods.html#liveness-and-readiness-probes-aka-health-checks) to the ReplicationController manifest.


helgi:
Looks good - missing `charts` PR is all

PR #93, opened 2015-12-15
docs(readme): update client installation docs

bacongobbler:
addresses @carmstrong's comment in #71


carmstrong:
The link appears to be broken? https://s3.amazonaws.com/get-deis/deis/workflow/220/220.1/client/deis

bacongobbler:
AH looks like travis was outputting the wrong URL. Indeed https://get-deis.s3.amazonaws.com/deis/workflow/220/220.1/client/deis works. I'll update it.

Issue #94, opened 2015-12-15
migrate client builds to bintray


bacongobbler:
Travis artifacts is great for creating builds and uploading to S3, but one thing it misses on is the ability to grab the "latest" build (as seen in the README). In order for a user to grab the latest client build, he or she must know the latest Travis build that went green and download the client using that URL.

@sgoings fronted the work with releasing the [helm](https://github.com/helm/helm) client on [bintray](https://bintray.com/) which so far has been a great success, so perhaps we should move over to bintray for that one-liner (`curl https://... | sh`) special sauce.

bacongobbler:
this is even moreso important because travis artifacts will only generate a linux client. That's great for testing on the k8s machines themselves, but not on a local cluster.

sgoings:
I wholeheartedly approve of this direction. :smile: Extra points if the publishing logic can be placed in a makeup kit similar to: https://github.com/sgoings/makeup-kit-go-cli

jchauncey:
alright for the time being should we build the client locally? I keep getting this error when downloading from travis - 

```
╰─○ ./deis
zsh: exec format error: ./deis
```

bacongobbler:
@jchauncey yes, that's because travis only compiles linux binaries. You'll have to build it locally for now.

helgi:
Already happened in https://github.com/deis/workflow/commit/35abd9d4b60fe4b78513b090491c7a3b1f4af193

helgi:
Open up a diff ticket if you want to generate more than just linux clients :)

PR #98, opened 2015-12-16
feat(travis): deploy client builds to bintray

bacongobbler:
Travis artifacts is great for creating builds and uploading to S3,
but one thing it misses on is the ability to grab the "latest" build
(as seen in the README). In order for a user to grab the latest
client build, he or she must know the latest Travis build that went
green and download the client using that URL.

Seth Goings fronted the work with releasing the helm client on
bintray, which so far has been a great success. This PR moves us
over to bintray for that one-liner `(curl https://... | sh)` special
sauce.

refs #94. Still need to update docs with the first uploaded release
after this PR is merged in order to close it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -32,20 +32,29 @@ install:
   - pip install -r rootfs/dev_requirements.txt
   - make -C docs/ deps
   - GLIDE_HOME=/home/travis/.glide make -C client bootstrap
+  - make prep-bintray-json
 script:
   - make test
   - make -C client/ build test
   - make -C docs/ build test
 deploy:
-  provider: script
-  # ensure client/doc builds aren't removed
-  # see https://docs.travis-ci.com/user/deployment/#Uploading-Files
-  skip_cleanup: true
-  script: _scripts/deploy.sh
-  on:
-    branch: master
+  - provider: script
+    # ensure client/doc builds aren't removed
+    # see https://docs.travis-ci.com/user/deployment/#Uploading-Files
+    skip_cleanup: true
+    script: _scripts/deploy.sh
+    on:
+      branch: master
+  - provider: bintray
+    skip_cleanup: true
+    file: _scripts/ci/bintray-ci.json
+    user: deis-admin
+    key:
+      secure: "JVCyPvXZIabZBhWgZ2wX4UDoTpIAjVnp+Ox6cXywYbACzKZSTz6G1kFcItzpUiEygUtye1213Zb+2050jq3BK7wohy5lcZCDJgrhyw6RxTRWIQQ5o+pS+O/AOYCSbpPn2E5goNutAhlOVsf2TlXkt4wz0jl5qOaecm0QXoiXBPUH5H1a3ifnCylybVG2jc2Kj/9S5uMGDQqocrPTXedZo9E/Es61MbKttlJGfcIrjbS71J8QZvDTcTSjzGT5CVdQulzkLNmFI5y31XwBE9XC6ro/Can10bIvy6yzYSWraUBTXVLWY2mPrPSlohOqNiYg2goQFQ2KwAGe6mVbq3UqOrYqNLXDdpSnCsRkx2KBw+ifET+0neq1NI3v5oSjKZ+p2zKCWQoOxahU40Eg+hA12oN17yHglaj2PGLuxYicDc+BQEGcGdBHAPJNXALd+rSDCdq5Gnd9HsCQE2Tyc+YK2bKvfpgcQNLS7gtiIxoRLZ1/qRBq3SB3IyQik7jjPe9Y0Meqnmdk8PeXM113/MSGdqZtVfyaOcT8SPgN22dhV42fs/BQtplTT3Hcs3yhmDwtl1w1udynerHcWx0PqZFn3h95SozJFPi8UdsbZog5V/CY/OAFs3K4bm3ay4Re1r2vTFCRuukp6UwtT5QR4kjsKWrLkewFbjId2FazJA8kMc4="
+    on:
+      tags: true
+      branch: master
 addons:
   artifacts:
     paths:
-      - client/deis
       - docs/_build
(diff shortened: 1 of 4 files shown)


sgoings:
Do you care about any of these addon related errors?

```
~/go/src/github.com/deis/workflow  ⑂ 39f566c    travis lint
Warnings for .travis.yml:
[x] has multiple addons entries, keeping last entry
[x] value for addons section is empty, dropping
[x] in addons section: dropping artifacts section: missing key bucket, missing key key, missing key secret
```

.travis.yml
```diff
@@ -32,20 +32,31 @@ install:
   - pip install -r rootfs/dev_requirements.txt
   - make -C docs/ deps
   - GLIDE_HOME=/home/travis/.glide make -C client bootstrap
+  - make prep-bintray-json
```
sgoings: Why not put this in the `before_deploy` stage?

.travis.yml
```diff
@@ -32,20 +32,31 @@ install:
...
+  - provider: script
+    # ensure client/doc builds aren't removed
+    # see https://docs.travis-ci.com/user/deployment/#Uploading-Files
+    skip_cleanup: true
+    script: _scripts/deploy.sh
+    on:
+      branch: master
+  - provider: bintray
+    # ensure client/doc builds aren't removed
+    # see https://docs.travis-ci.com/user/deployment/#Uploading-Files
```
sgoings: You probably don't want these comments, eh?

.travis.yml
```diff
@@ -32,20 +32,31 @@ install:
...
+    skip_cleanup: true
+    file: _scripts/ci/bintray-ci.json
+    user: deis-admin
+    key:
+      secure: "JVCyPvXZIabZBhWgZ2wX4UDoTpIAjVnp+Ox6cXywYbACzKZSTz6G1kFcItzpUiEygUtye1213Zb+2050jq3BK7wohy5lcZCDJgrhyw6RxTRWIQQ5o+pS+O/AOYCSbpPn2E5goNutAhlOVsf2TlXkt4wz0jl5qOaecm0QXoiXBPUH5H1a3ifnCylybVG2jc2Kj/9S5uMGDQqocrPTXedZo9E/Es61MbKttlJGfcIrjbS71J8QZvDTcTSjzGT5CVdQulzkLNmFI5y31XwBE9XC6ro/Can10bIvy6yzYSWraUBTXVLWY2mPrPSlohOqNiYg2goQFQ2KwAGe6mVbq3UqOrYqNLXDdpSnCsRkx2KBw+ifET+0neq1NI3v5oSjKZ+p2zKCWQoOxahU40Eg+hA12oN17yHglaj2PGLuxYicDc+BQEGcGdBHAPJNXALd+rSDCdq5Gnd9HsCQE2Tyc+YK2bKvfpgcQNLS7gtiIxoRLZ1/qRBq3SB3IyQik7jjPe9Y0Meqnmdk8PeXM113/MSGdqZtVfyaOcT8SPgN22dhV42fs/BQtplTT3Hcs3yhmDwtl1w1udynerHcWx0PqZFn3h95SozJFPi8UdsbZog5V/CY/OAFs3K4bm3ay4Re1r2vTFCRuukp6UwtT5QR4kjsKWrLkewFbjId2FazJA8kMc4="
+    on:
+      all_branches: true
+      condition: ${TRAVIS_PULL_REQUEST} == false
```
sgoings: This is by default true - see: https://docs.travis-ci.com/user/deployment/#Pull-Requests

Makefile
```diff
@@ -12,6 +12,16 @@ check-docker:
 	  exit 2; \
 	fi
 
+prep-bintray-json:
+# TRAVIS_TAG is set to the tag name if the build is a tag
+ifdef TRAVIS_TAG
+	@jq '.version.name |= "$(VERSION)"' _scripts/ci/bintray-template.json | \
+		jq '.package.repo |= "deis"' > _scripts/ci/bintray-ci.json
```
sgoings: We probably want to add `_scripts/ci/bintray-ci.json` to .gitignore...

.travis.yml
bacongobbler: we could, but then it'd get called twice; once for each deploy target. I figured we only needed to call it once.

.travis.yml
bacongobbler: I kept them there as an intention that both deploy scripts do the same thing, but I can probably remove it as a duplicate.

.travis.yml
bacongobbler: point taken! We should probably remove it from helm too then. I got it from there and was wondering if there was a specific reason you guys did it that way.

Makefile
bacongobbler: good point

.travis.yml
sgoings: I didn't realize it got called twice... where do you see that behavior documented (I couldn't find either case)?

bacongobbler:
> [x] in addons section: dropping artifacts section: missing key bucket, missing key key, missing key secret

those are fine. In the [travis artifacts document](https://docs.travis-ci.com/user/uploading-artifacts/) it says to use ARTIFACTS_BUCKET, ARTIFACTS_KEY and ARTIFACTS_SECRET. It could be just a travis linting error. The other errors are a little odd, though.

.travis.yml
sgoings: Yeah I think @mboersma was the one to find out about that through @Joshua-Anderson :smile:

.travis.yml
bacongobbler: when I was testing out #71, I tried creating a central dir and upload it all to S3, so I used the following:

```
before_deploy: mkdir build_artifacts
```

What I found was the build failed during `before_deploy` because the directory already existed. That was because there was two deploy targets. As soon as I commented out one, the build succeeded.

TL;DR I found it out through trial and error

bacongobbler: Here's how I found out: https://travis-ci.org/bacongobbler/workflow/builds/96340912#L1588

bacongobbler:
addressed your comments :)

PR #103, opened 2015-12-16
feat(glide): add glide.lock

sgoings:


helgi:
does this not require a bump from glide 0.7.2 to 0.8+ ? Look at the travis file

sgoings:
@helgi to get the speed improvements, yes. I'll fix that up.

PR #108, opened 2015-12-17
doc(quickstart.md,configuring-object-storage.md): start documenting object storage configs

### LINKED ISSUE #3: [storage] Design for Off-cluster Storage
For the alpha, we need to point a deis installation at off-cluster storage. This ticket is a placeholder for that design.

* [x] design off-cluster storage configuration (https://github.com/deis/builder/pull/21)
* [ ] document storage configuration in deis/workflow (see #63) _note: https://github.com/deis/workflow/pull/108 has the fix for this, and it was moved to the `v2.0-beta1` milestone._

### LINKED ISSUE #63: document how to configure storage
docs should include

- how workflow & builder detect minio
- the env vars to set in order to use external object storage (#3)

arschles:
Fixes #63 
Fixes #3

(note that in #3, this closes the second and final bullet point)

### DIFF
--- a/docs/src/installing-deis/configuring-object-storage.md
+++ b/docs/src/installing-deis/configuring-object-storage.md
@@ -18,7 +18,8 @@ Additionally, Deis ships with a [Minio](http://minio.io) [component](https://git
 
 The Deis components determine what object storage system to use via environment variables that you set up:
 
-TODO: list the lookup order
+- `DEIS_MINIO_SERVICE_HOST` and `DEIS_MINIO_SERVICE_PORT` - The in-cluster Minio service
+- `DEIS_OUTSIDE_STORAGE_HOST` and `DEIS_OUTSIDE_STORAGE_PORT` - The external S3-compatible object storage system
 
 # Limitations
 
--- a/docs/src/installing-deis/configuring-object-storage.md
+++ b/docs/src/installing-deis/configuring-object-storage.md
@@ -1,0 +1,26 @@
+# Configure Object Storage
+
+A variety of Deis components rely on an object storage system to do their work. These components are:
+
+- [workflow](https://github.com/deis/workflow)
+- [builder](https://github.com/deis/builder)
+- [slugbuilder](https://github.com/deis/slugbuilder)
+- [slugrunner](https://github.com/deis/slugrunner)
+- [registry](https://github.com/deis/registry)
+
+These components are built flexibly, so they can work out of the box with almost any system that is compatible with the [S3 API](http://docs.aws.amazon.com/AmazonS3/latest/API/APIRest.html).
+
+# Minio
+
+Additionally, Deis ships with a [Minio](http://minio.io) [component](https://github.com/deis/minio). This component runs as a Kubernetes service, and the components listed above are configured to automatically look for that service and use it as object storage if it's available.
+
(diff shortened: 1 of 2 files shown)


arschles:
moved to `v2.0-beta1` and will comment on the corresponding issues

cc @technosophos

jchauncey:
:+1:

docs/src/installing-deis/configuring-object-storage.md
```diff
@@ -0,0 +1,26 @@
...
+- [slugrunner](https://github.com/deis/slugrunner)
+- [registry](https://github.com/deis/registry)
+
+These components are built flexibly, so they can work out of the box with almost any system that is compatible with the [S3 API](http://docs.aws.amazon.com/AmazonS3/latest/API/APIRest.html).
+
+# Minio
+
+Additionally, Deis ships with a [Minio](http://minio.io) [component](https://github.com/deis/minio). This component runs as a Kubernetes service, and the components listed above are configured to automatically look for that service and use it as object storage if it's available.
+
+# Telling Deis What to Use
```
carmstrong: Should we omit this section entirely until we have content, or are we okay with this TODO?

docs/src/installing-deis/configuring-object-storage.md
```diff
@@ -0,0 +1,26 @@
...
+
+The Deis components determine what object storage system to use via environment variables that you set up:
+
+TODO: list the lookup order
+
+# Limitations
+
+- In some cases, the minio service must be started up before other components, so that they see the Minio service's environment variables. If you use [helm](https://github.com/helm/helm) to install deis, this ordering won't be ensured. (builder, slugbuilder and slugrunner are fine)
```
technosophos: What do you mean "won't be ensured" here?

docs/src/installing-deis/configuring-object-storage.md
arschles: forgot about this TODO, I'll fill it in now that I have the lookup order. thanks @carmstrong

docs/src/installing-deis/configuring-object-storage.md
arschles: I mean that a component that depends on minio may be installed before the minio service is up. Currently, registry is the only component that would suffer in this scenario.

arschles:
@carmstrong listed the lookup order

arschles:
@technosophos are you ok with the "this ordering won't be ensured" language?

docs/src/installing-deis/configuring-object-storage.md
technosophos: I don't see how that would happen with Helm, given that services are installed before RCs, daemonsets, and pods.

technosophos:
@arschles I don't understand the issue you are raising. We've never seen that happen, as services are always installed prior to anything that would use them. It could happen if you were _not_ using Helm... but with Helm it should not.

arschles:
nevermind, I forgot about the ordering point that you raised in https://github.com/deis/workflow/pull/108#discussion_r48189231.

docs/src/installing-deis/configuring-object-storage.md
arschles: ah, I forgot about the ordering. I'll remove that language

PR #109, opened 2015-12-17
feat(*): update django from 1.8 to 1.9

helgi:
Notable changes:
* ditch deis context_processor for https://docs.djangoproject.com/en/1.9/ref/contrib/sites/#site-middleware
* `registry` no longer treated as an app as it was causing AppRegistry prepared problems
* django-json-field swapped out for jsonfield for django 1.9 compatibility

### DIFF
--- a/rootfs/deis/settings.py
+++ b/rootfs/deis/settings.py
@@ -99,8 +99,7 @@ TEMPLATES = [
                 "django.template.context_processors.request",
                 "django.template.context_processors.static",
                 "django.template.context_processors.tz",
-                "django.contrib.messages.context_processors.messages",
-                "deis.context_processors.site"
+                "django.contrib.messages.context_processors.messages"
             ],
         },
     },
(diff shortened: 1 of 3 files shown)


rootfs/deis/settings.py
```diff
@@ -112,6 +111,7 @@
     'django.contrib.sessions.middleware.SessionMiddleware',
     'django.contrib.auth.middleware.AuthenticationMiddleware',
     'django.contrib.messages.middleware.MessageMiddleware',
+    'django.contrib.sites.middleware.CurrentSiteMiddleware',
```
bacongobbler: +1

rootfs/deis/settings.py
```diff
@@ -140,8 +140,7 @@
     'rest_framework.authtoken',
     'corsheaders',
     # Deis apps
-    'api',
-    'registry',
```
bacongobbler: woah. Why are we removing the registry module here?

bacongobbler: just saw your comment in the OP. Disregard.

slack:
Tested this, works!

PR #110, opened 2015-12-17
feat(*): use latest python (3.5 right now) for API

helgi:
`from __future__ import unicode_literals` is not needed anymore, see http://python-future.org/unicode_literals.html

This piggybacks on the Django 1.9 bump.

On OSX you can `brew install python3` and then do the following

```
virtualenv -p python3 deis-py3
source deis-py3/bin/activate
python --version
pip install --upgrade -r rootfs/requirements.txt
make test
```

Note I haven't looked at what is required on the travis side of things

### DIFF
--- a/rootfs/api/models.py
+++ b/rootfs/api/models.py
@@ -4,7 +4,6 @@
 Data models for the Deis API.
 """
 
-from __future__ import unicode_literals
 import base64
 from datetime import datetime
 import etcd
@@ -18,19 +17,16 @@ from threading import Thread
 from django.conf import settings
 from django.core.exceptions import ValidationError, SuspiciousOperation
 from django.db import models
-from django.db.models import Count
-from django.db.models import Max
+from django.db.models import Count, Max
 from django.db.models.signals import post_delete, post_save
 from django.dispatch import receiver
-from django.utils.encoding import python_2_unicode_compatible
 from jsonfield import JSONField
 from OpenSSL import crypto
 import requests
 from rest_framework.authtoken.models import Token
 
-from api import utils
 from registry import publish_release
-from utils import dict_diff, dict_merge, fingerprint
+from api.utils import dict_diff, dict_merge, fingerprint, generate_app_name
 
 
 User = settings.AUTH_USER_MODEL
@@ -82,9 +78,9 @@ def validate_id_is_docker_compatible(value):
 def validate_app_structure(value):
     """Error if the dict values aren't ints >= 0."""
     try:
-        if any(int(v) < 0 for v in value.viewvalues()):
+        if any(int(v) < 0 for v in value.values()):
             raise ValueError("Must be greater than or equal to zero")
-    except ValueError, err:
+    except ValueError as err:
         raise ValidationError(err)
 
 
@@ -141,7 +137,6 @@ class UuidAuditedModel(AuditedModel):
         abstract = True
 
 
-@python_2_unicode_compatible
 class App(UuidAuditedModel):
     """
     Application used to service requests on behalf of end-users
@@ -159,9 +154,9 @@ class App(UuidAuditedModel):
     @property
     def select_app_name(self):
         """Select a unique randomly generated app name"""
-        name = utils.generate_app_name()
+        name = generate_app_name()
         while App.objects.filter(id=name).exists():
-            name = utils.generate_app_name()
+            name = generate_app_name()
 
         return name
 
@@ -174,9 +169,9 @@ class App(UuidAuditedModel):
 
     def save(self, **kwargs):
         if not self.id:
-            self.id = utils.generate_app_name()
+            self.id = generate_app_name()
             while App.objects.filter(id=self.id).exists():
-                self.id = utils.generate_app_name()
+                self.id = generate_app_name()
 
         return super(App, self).save(**kwargs)
 
@@ -265,7 +260,7 @@ class App(UuidAuditedModel):
                 raise EnvironmentError(
                     'Container type {} does not exist in application'.format(container_type))
         msg = '{} scaled containers '.format(user.username) + ' '.join(
-            "{}={}".format(k, v) for k, v in requested_structure.items())
+            "{}={}".format(k, v) for k, v in list(requested_structure.items()))
         log_event(self, msg)
         # iterate and scale by container type (web, worker, etc)
         changed = False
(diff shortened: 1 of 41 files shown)


helgi:
I had to patch `simpleflock` to be python 3 compatible. Submitted a patch upstream but haven't heard anything. Pulling in my own branch for now

rootfs/registry/dockerclient.py
```diff
@@ -55,8 +54,8 @@ def publish_release(self, source, config, target, deis_registry):
     def build(self, source, config, repo, tag):
         """Add a "last-mile" layer of environment config to a Docker image for deis-registry."""
         check_blacklist(repo)
-        env = ' '.join("{}='{}'".format(
-            k, v.encode('unicode-escape').replace("'", "\\'")) for k, v in config.viewitems())
+        env = ' '.join("{}='{}'".format(k, v) for k, v in config.items())
```
helgi: This one worries me a tiny bit but so far I haven't seen issues but ... unicode :)

rootfs/requirements.txt
```diff
@@ -10,7 +10,8 @@ gunicorn==19.4.1
 psycopg2==2.6.1
 python-etcd==0.3.2
 PyYAML==3.11
-requests==2.8.1
-simpleflock==0.0.2
+requests==2.9.1
+#simpleflock==0.0.2
+git+https://github.com/helgi/python-simpleflock.git@python3#egg=simpleflock
```
helgi: waiting on upstream to merge, doubt they will tho - couldn't find any proper alternatives yet. Best idea is to remove `simpleflock` and related code if at all feasible

rootfs/dev_requirements.txt
```diff
@@ -2,7 +2,4 @@
 coverage>=4.0
 
 # Run "make flake8" to check python syntax and style
-flake8==2.4.1
-
-# Used for mocking endpoints
-mock==1.0.1
```
helgi: `mock` is built in to newer python `unittest`

rootfs/build.sh
```diff
@@ -17,11 +17,15 @@ apk add --update-cache \
...
   libpq \
   postgresql-dev \
-  python \
-  python-dev
+  python3 \
+  python3-dev \
+  git
+
+# Symlink python3 binary to python
+ln -s /usr/bin/python3 /usr/bin/python
```
helgi: For some odd reason the alpine package does not take care of this ... so odd

bacongobbler: perhaps it conflicts with the `python` package, so they just want to install it to `/bin/python3`. Either way this works :)

rootfs/api/models.py
```diff
@@ -1046,7 +1036,7 @@ def save(self, *args, **kwargs):
             config['domains'] = ''
 
         # convert from string to list to work with and filter out empty strings
-        domains = filter(None, config['domains'].split(','))
```
bacongobbler: is filter slower than iterating through the entire list, or is it gone in python 3.5?

rootfs/api/tests/test_healthcheck.py
```diff
@@ -11,8 +10,7 @@ def setUp(self):
     def test_healthcheck(self):
         # GET and HEAD (no auth required)
         resp = self.client.get(self.url)
-        self.assertEqual(resp.status_code, 200)
-        self.assertEqual(resp.content, "OK")
+        self.assertContains(resp, "OK", status_code=200)
```
bacongobbler: -1 LOC! :+1:

rootfs/registry/dockerclient.py
bacongobbler: I recall it was some ugly json parsing error... Might wanna do a `git blame` on that and figure out what the history behind this ugly code fixes

rootfs/requirements.txt
bacongobbler: We can remove the file locks once Docker fixes their concurrency issues ;)

rootfs/dev_requirements.txt
```diff
@@ -2,7 +2,4 @@
...
+flake8==2.5.1
```
bacongobbler: is the version bump here necessary? If not would you mind cutting this into a separate commit just for separation's sake?

rootfs/api/models.py
helgi: `pylint` and other things seem to recommend against the usage of it and `2to3` even converted it when I tried to run that for the lols.

`filter` in 3+ returns an iterator so I had to do a `list()` around it  if I wanted to keep it. Figured might as well go with comprehensions since we anyway use them in other places for similar purposes

rootfs/api/tests/test_healthcheck.py
helgi: `assertEqual` on a it kept telling me it was `b"OK"` - this seemed nicer ;)

rootfs/dev_requirements.txt
helgi: It's required - `2.5.0` adds Python 3.5 support

> Improvement Raise cap on PyFlakes for Python 3.5 support

rootfs/registry/dockerclient.py
helgi: I'm no closer. https://github.com/deis/deis/commit/55b587fc8cdb817da51242bf7d1aab83f2564e56 - @mboersma did something whacky there :)

mboersma: The unicode-escape encoder handles quotes and slashes and such perfectly for this case, except for single-quote. There's a lot of lore about this on StackOverflow and elsewhere. The `ENV` line in a Dockerfile needs to be escaped. Without explicit encoding we'll have problems using env vars that contain quotes and other such characters, but honestly I don't know how Python 3.5 treats this.

helgi: I was having issues in the test cases where I saw `b'DEIS'` inside the expected quotes, could be a function of the test framework tho

rootfs/requirements.txt
mboersma: The current Kubernetes 1.1.3 has Docker 1.8.2, at least when I deploy it according to the official vagrant instructions. And I don't think the "concurrent pull" bug was truly fixed until Docker 1.9, so Deis v2 will have to keep working around this.

(You can probably tell just by watching your k8s pods start up that Kubernetes itself had to implement the same workaround: all `docker pull` operations are serialized on a node.)

bacongobbler:
some flake8 errors in the test run, but everything else looks good :+1:

mboersma:
This PR makes Guido and Brett Cannon very happy. :snake:

Issue #112, opened 2015-12-17
docs: Document adding domain annotation before helm install


technosophos:
Right now the current docs suggest running a straight Helm install.

https://github.com/deis/workflow/blob/master/docs/src/installing-deis/installing-the-deis-platform.md

In practice, the Deis router is not ready out of the box. So the current workflow that I have to use is this:

```
$ helm fetch deis/deis
$ vi $(helm home)/workspace/deis/manifest/deis-router-rc.yaml
$ helm install deis
```

In the second step, I have to do something like this:

```yaml
apiVersion: v1
kind: ReplicationController
metadata:
  name: deis-router
  namespace: deis
  labels:
    heritage: deis
  annotations:
    deis.io/routerConfig: |
      {
        "domain": "10.245.1.3.xip.io",
        "useProxyProtocol": false
      }
```

Unless there is a better workaround, this seems to be required to get _any_ deis cluster working.

/cc @krancour

arschles:
fwiw, it's not necessary to edit _before_ helm install. you can run `kubectl --namespace=deis edit rc deis-router` and make the change there.

technosophos:
Is that a better user experience?

bacongobbler:
I've been doing the following on my hosts in order to get the router working. From my host:

```
$ cat /etc/hosts
10.245.1.3     example.com
10.245.1.3     deis.example.com
10.245.1.3     go.example.com
10.245.1.3     wooden-ferryman.example.com
```

One could do the same with dnsmasq to use wildcards, but this is what I've been using to test the router :)

arschles:
@technosophos not necessarily, just making note here of another option for editing the router manifest. Also, @bacongobbler IMO your strategy is the easiest way to get started on a local cluster, because no router configs necessary (afaik, router ships configured with `example.com`).

krancour:
Editing up front is nicer, imo, but it's good that you're not totally screwed if you forget.

The steps are correct @technosophos laid out look good.  And what @bacongobbler suggests also works and is also what router's own docs suggest.  (Which I am happy to change if we don't like it.)

I know it's not great that this needs to be customized/configured for each installation, but I see no getting around at ATM.  Even in deis v1.x, we required `deisctl config platform set domain=<whatever>` prior to installation.  Platform domain is info we can't do without.  :-/

technosophos:
Here's what I am going to suggest in the docs:

- Set up example.com by default. Focus on this.
[...]

PR #116, opened 2015-12-17
docs(installing-deis): add domain configuration

technosophos:
This addresses issue #112.


arschles:
@technosophos is there still more to #112 after this?

technosophos:
@arschles I wasn't planning on it. Once I started writing, I realized that simpler is probably better here. But if people feel like there should be more, I'd do that.

arschles:
cool, I was asking because I'm +1 on this finishing #112. ok if we close that out with this?

PR #95, opened 2015-12-16
feat(workflow): identify app type and move envs to rc templates

smothiki:
K8s selects respective RC template based on app type.
Dockerfile app types wont go through publish_release function
Moved environment variables to RC templates.

The following was added by @arschles on 12/17/2015:
Fixes #47

PR #117, opened 2015-12-17
fix(api): delete new_release only if it exists

mboersma:
When debugging #95, there was a case where the `except` handler failed since `new_release` hadn't been defined yet in the preceding `try` block. This guards against that potential problem in the three similar cases I could find.


bacongobbler:
nice. This certainly fixes up a few edge cases.

PR #124, opened 2015-12-17
doc(routing/ingress): improve docs

krancour:
Many, many apologies... this is a rewrite of doc that @technosophos added just this morning.  After seeing @sgoings stumble over a few issues, I realized the docs had to be a bit more descriptive in terms of the possible paths on can take to set this up (or may be forced to take, depending on the capabilities of their cluster and underlying infrastructure).

Essentially, this now documents three paths:

1.  Running where the "cloud provider" was able to provision an external load balancer connected to the router service.

2. Running where the above is _not_ true, but you are still able to provision your own load balancer.

3. Where 1 and 2 both don't work... i.e. Vagrant.

Don't anyone be shy about feedback on this PR.  These changes are extensive and I want to be super certain that the directions would be crystal clear to the average Deis operator.

### DIFF
--- a/docs/src/installing-deis/installing-the-deis-platform.md
+++ b/docs/src/installing-deis/installing-the-deis-platform.md
@@ -35,23 +35,119 @@ $ kubectl get pods --namespace=deis
 
 Once you see all of the pods ready, your Deis platform is running on a cluster!
 
-## Mapping a Default Domain
+## Routing Traffic
 
-Deis will route traffic from Kubernetes nodes to Deis components, but to
-do so it needs a domain. By default, Deis is configured to use the
-domain `example.com`. For basic Deis testing, simply add an entry to
-your host machine's `/etc/hosts` file.
+The Deis router component will use an incoming request's HTTP `Host`
+header to direct traffic to Deis components and to applications deployed
+via Deis. It is therefore important that the router be configured
+properly and that DNS and/or your local `/etc/hosts` file are as well.
+
+### Configuring the Platform Domain
+
+The Deis router requires a "platform domain" be set.  Deis components
+that respond to HTTP requests and applications deployed via Deis will
+all live on subdomains of the platform domain.  When installing Deis
+via helm, the platform domain is set to `example.com` by default.  This
+can be changed by editing Deis' helm chart prior to installation.
+Alternatively, the change can be made after Deis has been installed
+by editing the router's replication controller like so:
+
+```
+$ kubectl edit rc deis-router --namespace=deis
+```
+
+Find `example.com` in the manifest, change it to a domain of your
+choosing, and save changes.  Kubernetes will apply the changes and
+the router will dynamically reconfigure itself accordingly.
+
+### Getting Traffic to the Cluster
+
+With the platform domain configured, the only remaining concern
+is to ensure traffic for that domain reaches the cluster.  There
+are a few ways of doing this, and which you use may depend on the
+capabilities of your infrastructure and your Kubernetes cluster.
+
+In general, the goal is to connect traffic bound for
+`*.the-domain-you.picked` to your Kubernetes cluster.
+
+#### Using an Automatically Provisioned Load Balancer
+
+If your Kubernetes cluster and the underlying infrastructure
+support it, the Deis router's service (which is of `type:
+LoadBalancer`) will automatically provision an "external"
+(external to the Kubernetes cluster) load balancer.  On AWS,
+for instance, this takes the form of an ELB, while on Google
+Container Engine, it takes the form of a routing rule.
+
+If it is supported, you will be able to find the DNS name or
+IP(s) of the external load balancer like so:
+
+```
+$ kubectl describe service deis-router --namespace=deis
+```
+
+Examine the `LoadBalancer Ingress` field to find the DNS
+name or IP(s).  The former can be added to DNS as a CNAME,
+while the latter can be added as an A record.
+
+#### Creating Your Own Load Balancer
+
+On Kubernetes clusters and underlying infrastructure that
+do not support the automated provisioning of an external
+load balancer for the Deis router, you may provision your
+own and add _all_ minion nodes of the Kubernetes cluster to
+the pool.
+
+The load balancer must pass incoming traffic on the
+following ports to the _same_ ports on the Kubernetes
+worker nodes:
+
+* 80
+* 443
+* 2222
+* 9090
+
+The Deis router component listens to those ports on any
+host it runs on.  Since not _all_ nodes are likely to
+be running the router, a health check should be used to
+determine which node(s) can serve requests.  The router's
+healthcheck should use the HTTP protocol, port `9090`, and
+the request path `/healthz`.
+
+Load balancer idle timeouts should be set to at least `1200`
+seconds so connections do not close prematurely during longer
+running requests such as application deployments.
+
+#### Without a Load Balancer
+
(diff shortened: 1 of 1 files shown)


krancour:
Tentatively tagging as alpha.  We an change that is people disagree.

docs/src/installing-deis/installing-the-deis-platform.md
```diff
@@ -35,78 +35,156 @@ $ kubectl get pods --namespace=deis
...
+header to direct traffic to Deis components and to applications deployed
+via Deis. It is therefore important that the router be configured
+properly and that DNS and/or your local `/etc/hosts` file are as well.
+
+### Configuring the Platform Domain
+
+The Deis router requires a "platform domain" be set.  Deis components
+that respond to HTTP requests and applications deployed via Deis will
+all live on subdomains of the platform domain.  When installing Deis
+via helm, the platform domain is initially set to `example.com`.  This
```
arschles: Nit: I think this should say "set by default to `example.com`"

technosophos: To really be nit picky, that would be a split infinitive, @arschles

arschles: haha, I was just going for the "default" part, but go nuts

docs/src/installing-deis/installing-the-deis-platform.md
```diff
@@ -35,78 +35,156 @@ $ kubectl get pods --namespace=deis
...
+Examine the `LoadBalancer Ingress` field to find the DNS
+name or IP(s).  The former can be added to DNS as a CNAME,
+while the latter can be added as an A record.
+
+#### Creating Your Own Load Balancer
+
+On Kubernetes clusters and underlying infrastructure that
+do not support the automated provisioning of an external
+load balancer for the Deis router, you may provision your
+own and add _all_ worker nodes of the Kubernetes cluster to
```
arschles: add all _minion_ nodes?

arschles:
:100: I like that we now have a "tl;dr" section on networking and also this.

arschles:
I had 2 nits, but still gave an LGTM. I like these docs

docs/src/installing-deis/installing-the-deis-platform.md
bacongobbler: +1

docs/src/installing-deis/installing-the-deis-platform.md
```diff
@@ -35,78 +35,156 @@ $ kubectl get pods --namespace=deis
...
 
 You can find the IP address of a node using `kubectl`:
 
 ```
-$ kubectl get nodes
-NAME         LABELS                              STATUS    AGE
-10.245.1.3   kubernetes.io/hostname=10.245.1.3   Ready     19h
+$ kubectl get pods --namespace=deis | grep deis-router
+NAME                        READY     STATUS    RESTARTS   AGE
+deis-router-ih25q           1/1       Running   0          2h
```
sgoings: The header field won't show up with this grep. :smile:

docs/src/installing-deis/installing-the-deis-platform.md
```diff
@@ -35,78 +35,156 @@ $ kubectl get pods --namespace=deis
...
+The Deis router requires a "platform domain" be set.  Deis components
+that respond to HTTP requests and applications deployed via Deis will
+all live on subdomains of the platform domain.  When installing Deis
+via helm, the platform domain is initially set to `example.com`.  This
+can be changed by editing Deis' helm chart prior to installation.
+Alternatively, the change can be made after Deis has been installed
+by editing the router's replication controller like so:
+
+```
+$ kubectl edit rc deis-router --namespace=deis
```
michelleN: Should we be using `helm` to edit the manifest here and then apply the changes using kubectl instead of directly using kubectl so that the chart will have the most updated version of the manifests?

technosophos: And use `kubectl apply`?

docs/src/installing-deis/installing-the-deis-platform.md
```diff
@@ -35,78 +35,156 @@ $ kubectl get pods --namespace=deis
...
+use `kubectl` to determine what node the pod is running on:
+
+```
+$ kubectl describe pod describe pod deis-router-ih25q --namespace=deis
+```
+
+The `Node` field of the response should provide an IP:
+
+```
+Node:				172.17.8.100/
```
technosophos: Why did we change the IP from the one that `kube-up` generates to something else?

docs/src/installing-deis/installing-the-deis-platform.md
mboersma: It's not a split infinitive if there's no verb after `to`. (Also, split infinitives are fine these days according to most style guides.)

docs/src/installing-deis/installing-the-deis-platform.md
krancour: I actually knew that... I deliberately put it there because I thought it added valuable context and didn't think anyone would notice.  You win a prize @sgoings!

technosophos:
I'm marking this a showstopper because I'd _really_ like to see it get into the alpha docs. If @krancour thinks we can't get it done tomorrow, we can remove it from the showstopper list.

docs/src/installing-deis/installing-the-deis-platform.md
krancour: Changing to:

> When installing Deis via helm, the platform domain is set to `example.com` by default.

docs/src/installing-deis/installing-the-deis-platform.md
krancour: Is minion a more proper term to use than worker?  I feel I see both terms uses interchangeably.  I do think it's useful if we all stick to a common lexicon, so if minion is prefered, I will change this.

docs/src/installing-deis/installing-the-deis-platform.md
krancour: @michelleN I think editing the chart _before_ installation is ideal, and that was the first option mentioned:

> This can be changed by editing Deis' helm chart prior to installation.

Do you think more detail on that process is required?

As for the use of `kubectl edit` to change this post-install, I was trying to document the lowest friction option for correcting the mistake if one forgot to edit the chart before installation.

In both cases, I was trying to not go too deep into helm workflow in this doc (partly because it feels to me as if it's still evolving).  I can add it though if we all think it's warranted.

krancour:
@technosophos I see no reason this can't be done tomorrow.  It sound like just some minor edits are needed.

docs/src/installing-deis/installing-the-deis-platform.md
krancour: I hadn't used kube-up much, so `10.245.1.3` initially struck me as an IP one would more likely see as a private IP on a compute instance in some cloud provider.  The change was because I was going for was an IP that _looked_ Vagranty.  I realize, of course, that what I actually did was picked one that looked more Vagrant _to me_.

I've just looked at the at kube-up's default `NODE_IP_BASE ` and it appears to be `10.245.1.`.

So I'm changing this back.

docs/src/installing-deis/installing-the-deis-platform.md
krancour: But I'll fix it.

krancour:
I have addressed and/or acted on most of the feedback.

PR #126, opened 2015-12-17
feat(_tests): add config tests

arschles:
also, refactor the random app name generation functionality

this PR contains all pending config tests except for those which load and unload config to/from `.env` files


arschles:
@mboersma does the CI here not run functional tests (in _tests)?

technosophos:
@arschles it does not yet run them.

arschles:
hrm, well since I haven't yet run integration tests to successful completion locally, I'm gonna hold off on merging until I can

technosophos:
Well, the advantage of committing is that then the rest of us can run the tests, too, and fix anything broken. But it's your call.

arschles:
I see. I was thinking of it more in the light of "breaking the build," but I guess since we don't have CI set up for them yet...

_tests/config_test.go
```diff
@@ -2,44 +2,77 @@ package _tests_test
...
+			Expect(err).NotTo(HaveOccurred())
+			Expect(out).To(ContainSubstring("POWERED_BY      the Deis team"))
 		})
 
-		XIt("can set a multi-line environment variable", func() {
-			// `deis config:set FOO="This is a
-			//multiline string" --app={{.AppName}}`
-			// "deis run env --app=%s"
+		It("can set a multi-line environment variable", func() {
+			out, err := execute(`deis config:set FOO=This\ is\ a\
```
mboersma: Are the slashes necessary here? I thought in Go that the tick-quotes made a literal string with no escaping needed.

_tests/util_test.go
```diff
@@ -0,0 +1,16 @@
+package _tests_test
+
+import (
+	"fmt"
+	"math/rand"
+
+	"github.com/onsi/ginkgo/config"
+)
+
+func init() {
```
mboersma: I started putting this type of common code in the test suite file itself, `tests_suite_test.go`. Is there a logical split between these funcs and those? Otherwise we should probably keep them together in one or the other file.

_tests/config_test.go
arschles: No escaping needed in the Go string, but they are necessary to run the command. It outputs a failure message and ultimately sets the value of `FOO` to `This` otherwise

_tests/util_test.go
arschles: No logical split, I'll move it into there.

arschles:
@mboersma addressed your comments

PR #130, opened 2015-12-18
ref(k8s): simplify _create_rc with dict_merge to deal with RC templates

helgi:


helgi:
This PR is not viable anymore given other changes that have gone into the RC around slugbuilder

PR #136, opened 2015-12-18
Small test fixes

mboersma:
Removes the `example-go` app that was accidentally checked in to version control. Also stops checking for optional `DEIS_WORKFLOW_SERVICE_PORT` and doesn't hard-code the location of the `deis` CLI.


arschles:
thanks @mboersma

PR #142, opened 2015-12-18
fix(makefile): Update makefile and deploy.sh to work like the other repos

jchauncey:


mboersma:
I thought `VERSION` was preferred to `BUILD_TAG`, since that's what we have in the [protoptype-repo](https://github.com/deis/prototype-repo/blob/master/Makefile#L14).

jchauncey:
I think yesterday with the change to make the etcd makefile more in line with other repos we decided to change it to BUILD_TAG. I saw that this was set to version and figured we might as well make them all consistent. I dont care either way (I actually had VERSION in the logger and fluentd makefiles) but we should pick and make them all consistent

iancoffey:
We just added [BUILD_TAG to etcd](https://github.com/deis/etcd/blob/master/Makefile#L6), fwiw.

mboersma:
I don't care either way either--just trying to keep things consistent like you are. :-) This is fine with me, but we should PR the same change against the prototype-repo too.

arschles:
and the other repos as well. +1 on consistency

bacongobbler:
Since this isn't up to date with master yet, you'll also need to modify the `prep-bintray-json` target as well.

jchauncey:
@bacongobbler done

bacongobbler:
:+1:

jchauncey:
everyone ok with merging this?

arschles:
:+1: from me

I'm a little concerned about the `VERSION` vs. `BUILD_TAG` decision, but feel like it's worth putting off a final standard until next month earliest

PR #151, opened 2015-12-18
docs(README): update client dev build instructions

bacongobbler:
relies on deis/deis.io#133


bacongobbler:
I'm assuming you both approve of the script at deis/deis.io#133 as well?

PR #155, opened 2015-12-19
doc(installing-the-deis-platform.md): change instructions to edit router RC manifest before helm install

### LINKED ISSUE #149: docs: Fix installation instructions to _not_ use `kubectl edit`
We should not be recommending `kubectl edit` for two big reasons:

1. It does not work with all editors. With `vi` on Mac it silently fails OOTB.
2. If you do it, your remote state differs from your local charts, and every time you update you have to re-set that value or else routing breaks.

The correct method is to edit the chart in the workspace and use `kubectl apply` to load it back in.

Options for editing are:

- Use `helm edit`, though only the newest version of Helm works well
- Edit `$(helm home)/workspace/charts/deis/manifest/deis-router-rc.yaml`. This is what the first version said, but somebody removed this and I don't know why.

arschles:
Fixes #149

I had to move more text around than I would have liked here. The major change is introducing the edit step before `helm install`, which required me to explain (as minimally as I could) what the platform domain should be.

I'd like suggestions particularly on how to improve that text. Restructuring suggestions are also welcome.

cc/ @technosophos @krancour

### DIFF
--- a/docs/src/installing-deis/installing-the-deis-platform.md
+++ b/docs/src/installing-deis/installing-the-deis-platform.md
@@ -44,9 +44,10 @@ Now that you have the Deis chart, you'll need to make a minor edit to the
 incoming traffic on a domain of your choice. Follow the following steps to make
 the edit:
 
-1. Open the `$HELM_HOME/workspace/charts/deis/manifests/deis-router-rc.yaml`
-file in your favorite editor. Note that if `HELM_HOME` is not in your environment,
-it defaults to `~/.helm`
+1. Open the `$(helm home)/workspace/charts/deis/manifests/deis-router-rc.yaml`
+file in your favorite editor. Note that `$(helm home)` executes the command to get
+the directory that helm uses to store its config and all its charts. See
+http://helm.readthedocs.org/en/latest/workspace/ for details.
 2. Replace `example.com` under `metadata.annotations.deis.io/routerConfig` with
 the domain of your choice. This will be called the _platform domain_. You may choose
 any domain, but the `Routing Traffic` section below details how to configure DNS or your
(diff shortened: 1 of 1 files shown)


docs/src/installing-deis/installing-the-deis-platform.md
```diff
@@ -22,54 +24,65 @@ properly by running
...
+incoming traffic on a domain of your choice. Follow the following steps to make
+the edit:
 
-The Deis router component will use an incoming request's HTTP `Host`
-header to direct traffic to Deis components and to applications deployed
-via Deis. It is therefore important that the router be configured
-properly and that DNS and/or your local `/etc/hosts` file are as well.
+1. Open the `$HELM_HOME/workspace/charts/deis/manifests/deis-router-rc.yaml`
+file in your favorite editor. Note that if `HELM_HOME` is not in your environment,
+it defaults to `~/.helm`
```
krancour: At runtime, the helm tool may consider its home to be `~/.helm` if the `HELM_HOME` env var is not set, but that assumption on the part of the tool isn't applicable to the imperative to "Open the `$HELM_HOME/workspace/charts/deis/manifests/deis-router-rc.yaml` file in your favorite editor."

It would, however, work if it read: "Open the `$(helm home)/workspace/charts/deis/manifests/deis-router-rc.yaml` file in your favorite editor."

slack: Either `~/.helm/...` (assuming a default install) or `helm home` subshell version works for me.

sgoings:
I like the move to a simpler getting started guide. :smile:

arschles:
@krancour fixed the `$HELM_HOME` language

Issue #161, opened 2015-12-21
v1.x docs are diluting the value of the v2 docs


krancour:
Lots and lots of inapplicable Deis v1.x documentation is still included in this repo.  While maintainers and active contributors may easily distinguish what's old from what's accurate and current, this is probably not the case for most users.

arschles:
@krancour can you give a few examples, as pointers to getting started? I have a feeling we can fix this one in smaller chunks

krancour:
There's a lot.  Some examples:

Mentions of deisctl:

https://github.com/deis/workflow/blob/master/docs/src/faq.md#whats-the-difference-between-the-deis-and-deisctl-clients

https://github.com/deis/workflow/blob/master/docs/src/installing-deis/installing-deisctl.md

Pre-2.0 versions of the workflow API:

https://github.com/deis/workflow/tree/master/docs/src/reference-guide/controller-api

Mentions of ceph and implicit references to Fleet as scheduler:

https://github.com/deis/workflow/blob/master/docs/src/understanding-deis/failover.md

CoreOS (which we're no longer confined to):

https://github.com/deis/workflow/blob/master/docs/src/understanding-deis/concepts.md#coreos

References to logspout, publisher, and store (which are all gone):

https://github.com/deis/workflow/blob/master/docs/src/understanding-deis/components.md

Architecture diagrams mention Fleet:

https://github.com/deis/workflow/blob/master/docs/src/understanding-deis/DeisSystemDiagram.png

These examples are non-exhaustive.

krancour:
And I agree that it can be accomplished in small chunks.

arschles:
@krancour thanks, that helps a lot as a starting point

krancour:
This is really a meta-issue of sorts that encompasses quite a variety of outstanding doc changes.  But since we've determined that the need for more accurate docs (or at least the elimination of _inaccurate_ docs) is a showstopper, tagging as such.

bacongobbler:
sounds like once #163 is closed we can close this one out as well.

arschles:
:+1:

PR #165, opened 2015-12-21
doc(*): strike references to logger

krancour:
Another PR that partially addresses #161 

Logger is neither included in nor a dependency of workflow ATM.  It's an add-on that requires a specially configured k8s cluster.


krancour:
Note that reference to logger can be added back in in the future, but I expect the docs will change significantly between now and then, so I submit that it's better to lose these refs for now.

Issue #140, opened 2015-12-18
Edgecases for secrets and outside storage


smothiki:
Right now secrets are must to run buildpack type apps. we are mounting secrets to workflow in deis namespace and reading secrets from file and recreating secrets in app namespace . This is not good if the secrets change. we have to read secrets from api server and recreate the secrets in app name space.

Also when there is a new release for an existing app need to create new secret if there is a change in secrets

bacongobbler:
This is also a bit of a security concern by mounting the platform's access and secret keys into the user's container. We should probably step back and rethink this process a bit.

arschles:
I have an incomplete but maybe promising solution - S3's API includes the ability the generate [presigned URLs](http://docs.aws.amazon.com/AmazonS3/latest/dev/PresignedUrlUploadObject.html). Minio also supports them.

The tl;dr of these URLs is that they allow anyone who has one to download a pre-specified object w/o creds, within a duration after which the URL expires. If we go this route we'd obviously have to build in some kind of renewal process.

arschles:
Actually, if each slugrunner only has to download the slug once on startup then pre-signed URLs would work fine. Is that assumption right?

jchauncey:
ok what are the secretes and what do we use them for?

krancour:
I believe they're used for slugrunners to fetch slugs from minio.  Buuut... this is going to be an issue all over again when we look to add end-to-end SSL.

smothiki:
No. It's about the logic how we handle cases. Let's say user has set outside storage and deployed secrets for minio which are not meant for outside storage . We will check the URL and then minio secret mounts and follow that condition try to add minio host configuration for outside storage using minio secretes. Which would fail

arschles:
@krancour regarding SSL, let's keep that to a separate conversation. I've heard some ideas about setting up an intermediate CA, which would change the conversation.

@smothiki that's why I suggested pre-signed URLs. _If_ the condition I listed above holds, then slugrunner would not have to mount the secret and this would become a non-issue. Otherwise, it may be worth our time to investigate a renewal protocol so that we wouldn't have to transport secrets across namespaces.

smothiki:
I'm ok with it . If it solves the issue

technosophos:
Okay, this issue is _way_ too generic. You need to elaborate all of the secrets that we're talking about here.

[...]

PR #176, opened 2015-12-22
feat(secrets): read secrets from apiserver

smothiki:
```
$ kubectl get pods --namespace=atomic-electron
NAME                           READY     REASON    RESTARTS   AGE
atomic-electron-v2-web-c8ftn   1/1       Running   0          48s
$ deis ps
=== atomic-electron Processes
--- web:
web.1 up (v2)
[vagrant@kubernetes-minion-1 example-go]$ curl http://atomic-electron.10.245.1.4.xip.io
Powered by Deis
Release v2 on atomic-electron-v2-web-c8ftn
```
This PR addreses change of secrets problem in the issue https://github.com/deis/workflow/issues/140

### DIFF
--- a/manifests/deis-workflow-rc.yml
+++ b/manifests/deis-workflow-rc.yml
@@ -37,13 +37,7 @@ spec:
           volumeMounts:
             - mountPath: /var/run/docker.sock
               name: docker-socket
-            - name: minio-user
-              mountPath: /var/run/secrets/deis/minio/user
-              readOnly: true
       volumes:
         - name: docker-socket
           hostPath:
             path: /var/run/docker.sock
-        - name: minio-user
-          secret:
-            secretName: minio-user
--- a/rootfs/scheduler/k8s.py
+++ b/rootfs/scheduler/k8s.py
@@ -5,7 +5,6 @@ import re
 import string
 import time
 import urlparse
-import base64
 
 from django.conf import settings
 from docker import Client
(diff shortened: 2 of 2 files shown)


arschles:
so, this reads the secrets on each slugrunner launch instead of mounting them, correct?

smothiki:
yes

arschles:
ok, gave an LGTM. can you take off the reference to #140 though so we can keep the discussion about presigned URLs? an alternative to that, if the issue has gone out of scope, is to create a new one. your call

arschles:
nevermind, I just saw that you didn't write `Fixes`

technosophos:
I may not be understanding this fix, but it doesn't seem to me that this is the right approach to solving the problem.

First, this doesn't fix any part of the initial #140 issue, which is that if the secret in the `deis` namespace changes, those changes have to be replicated in every single app namespace. As far as I can tell, this patch doesn't do anything to fix that.

Second, what this fixes is sort of a non-issue. By design, if a secret changes, all services that use it should be restarted (12 Factor: release = configuration + build). And since the secrets are explicitly declared in a queryable way, we can see using kubectl which services depend on those.

I'm pondering whether this sort of solution would be good to implement in slugrunner (since it _could_ query the apiserver for credentials). That's actually what I was thinking we were doing here. But I don't actually think that's a particularly great solution, either. I don't like the idea of reading secrets from the API server, especially over a namespace boundary.

I think you were right, @smothiki. We should pull this out of the showstoppers list and do it the right way with signed URLs as you and @arschles mentioned the other day.

arschles:
@smothiki I removed my LGTM. I just talked to @technosophos and I believe that it's acceptable, at least for alpha, to say that operators must restart their deis cluster if these secrets change.

also related, but not directly applicable to this PR, is that I've opened https://github.com/deis/workflow/issues/182 to track a a future enhancement which will eliminate the copying of secrets across k8s namespaces.

smothiki:
reading from file system is still efficient than reading over the network.

PR #180, opened 2015-12-22
doc(dev-env): update dev env docs

krancour:
This is a massive rewrite of the docs for setting up the development environment.

Note, I expect these to shrink significantly as we make more headway in transitioning each component's development environment to something container-based.  For now, this is necessary...

This does away with mentions of defunct things like `make dev-cluster` and `deisctl` and replaces them with information about building and deploying to k8s.

This massive update also re-orders docs a bit and pays down some outstanding inaccuracies.

Please note that this doc tries really hard to document the _general_ procedures for hacking on each of our _many_ components.  Until we have made our Makefiles uniform, your mileage from these instructions may vary from one component to the next.  I don't see this as a flaw in the docs; rather I see adherence to the docs as something for each component to aspire to.

### DIFF
--- a/docs/src/contributing/development-environment.md
+++ b/docs/src/contributing/development-environment.md
@@ -1,280 +1,281 @@
 # Development Environment
 
-[DigitalOcean][] is the recommended development environment for Deis project contributors. [Provision a new DO cluster](../installing-deis/digitalocean.md) and then continue to follow the instructions below to get started hacking.
+This document is for developers who are interested in working directly on the Deis codebase. In this guide, we walk you through the process of setting up a development environment that is suitable for hacking on most Deis components.
 
-!!! important
-    Are you a new contributor to Deis? Your first [Pull Request][pr] could earn you credit at [DigitalOcean][]! Submit your changes and then email <deis@engineyard.com>. When your PR is merged, the maintainer team will send you a DigitalOcean credit based on the value of your contribution.
+We try to make it simple to hack on Deis components. However, there are necessarily several moving pieces and some setup required. We welcome any suggestions for automating or simplifying this process.
 
-This document is for developers who are interested in working directly on the Deis codebase. In this guide, we walk you through the process of setting up a local development environment. While there are many ways to set up your Deis environment, this document covers a specific setup:
-
-- Developing on **Mac OSX** or **Linux**
-- Managing virtualization with **Vagrant/Virtualbox**
-- Hosting a docker registry with **docker-machine** (Mac)
-
-We try to make it simple to hack on Deis. However, there are necessarily several moving pieces and some setup required. We welcome any suggestions for automating or simplifying this process.
+!!! note
+    The Deis team is actively engaged in containerizing Go and Python based development environments tailored specifically for Deis development in order to minimize the setup required.  This work is ongoing.  Refer to the [deis/router][router] project for a working example of a fully containerized development environment.
 
 If you're just getting into the Deis codebase, look for GitHub issues with the label [easy-fix][]. These are more straightforward or low-risk issues and are a great way to become more familiar with Deis.
 
 ## Prerequisites
 
-You can develop on any supported platform including your laptop, cloud providers or on bare metal. We strongly recommend a minimum 3-node cluster. We strongly suggest using Vagrant and VirtualBox for your virtualization layer during development.
+In order to successfully compile and test Deis binaries and build Docker images of Deis components, the following are required:
 
-At a glance, you will need:
+- [git][git]
+- Go 1.5 or later, with support for compiling to `linux/amd64`
+- [glide][glide]
+- [golint][golint]
+- [shellcheck][shellcheck]
+- [Docker][docker] (in a non-Linux environment, you will additionally want [Docker Machine][machine])
+
+For [deis/workflow][workflow], in particular, you will also need:
 
 - Python 2.7 or later (with `pip`)
 - virtualenv (`sudo pip install virtualenv`)
-- Go 1.5 or later, with support for compiling to `linux/amd64`
-- [Godep](https://github.com/tools/godep)
-- VirtualBox latest
-- Vagrant 1.5 or later
-
-On Mac, you will also want [Docker Machine][machine].
-
-Additionally, you may need:
-
-- [shellcheck](https://github.com/koalaman/shellcheck)
-- [golint](https://github.com/golang/lint)
 
 In most cases, you should simply install according to the instructions. There are a few special cases, though. We cover these below.
 
 ### Configuring Go
 
-If your local workstation does not support the `linux/amd64` target environment, you will have to install Go from source with cross-compile support for that environment. This is because some of the components are built on your local machine and then injected into a docker container.
+If your local workstation does not support the `linux/amd64` target environment, you will have to install Go from source with cross-compile support for that environment. This is because some of the components are built on your local machine and then injected into a Docker container.
 
 Homebrew users can just install with cross compiling support:
 
-
-    $ brew install go --with-cc-common
+```
+$ brew install go --with-cc-common
+```
 
 It is also straightforward to build Go from source:
 
-    $ sudo su
-    $ curl -sSL https://golang.org/dl/go1.5.src.tar.gz | tar -v -C /usr/local -xz
-    $ cd /usr/local/go/src
-    $ # compile Go for our default platform first, then add cross-compile support
-    $ ./make.bash --no-clean
-    $ GOOS=linux GOARCH=amd64 ./make.bash --no-clean
+```
+$ sudo su
+$ curl -sSL https://golang.org/dl/go1.5.src.tar.gz | tar -v -C /usr/local -xz
+$ cd /usr/local/go/src
+$ # compile Go for our default platform first, then add cross-compile support
+$ ./make.bash --no-clean
+$ GOOS=linux GOARCH=amd64 ./make.bash --no-clean
+```
 
-Once you can compile to `linux/amd64`, you should be able to compile Deis' components as normal.
+Once you can compile to `linux/amd64`, you should be able to compile Deis components as normal.
 
 ### Configuring Docker Machine (Mac)
 
-Deis needs a Docker registry running independently of the Deis cluster. On OS X, you will need [Docker Machine][machine] to run the registry inside of a VirtualBox image.
+Deis needs Docker for building images.  Docker utilizes a client/server architecture, and while the Docker client is available for Mac OS, the Docker server is dependent upon the Linux kernel.  Therefore, in order to use Docker on Mac OS, [Docker Machine][machine] is used to facilitate running the Docker server within a [VirtualBox][vbox] VM.
+
+Install Docker Machine according to the normal installation instructions, then use it to create a new VM:
+
+```
+$ docker-machine create deis-docker \
+    --driver=virtualbox \
+    --virtualbox-disk-size=100000 \
+    --engine-insecure-registry 10.0.0.0/8 \
+    --engine-insecure-registry 172.16.0.0/12 \
+    --engine-insecure-registry 192.168.0.0/16 \
+    --engine-insecure-registry 100.64.0.0/10
+```
+
+This will create a new virtual machine named `deis-docker` that will take up as much as 100,000 MB of disk space. The images you build may be large, so allocating a big disk is a good idea.
+
+Once the `deis-docker` machine exists, source its values into your environment so your docker client knows how to use the new machine. You may even choose to add this to your bash profile or similar.
+
+```
+$ eval "$(docker-machine env docker-deis)"
+```
+
+After following these steps, some Docker Machine users report a slight delay (30 - 60 seconds) before the Docker server is ready.
 
 !!! note
-    Previously, Deis used boot2docker to run the registry. However, Docker has deprecated boot2docker in favor of Docker Machine.
+    In subsequent steps, you may run a Docker registry within the `deis-docker` VM. Such a registry will not have a valid SSL certificate and will use HTTP instead of HTTPS. Such registries are implicitly untrusted by the Docker server (which is also running on the `deis-docker` VM).  In order for the Docker server to trust the insecure registry, `deis-docker` is explicitly created to trust all registries in the IP ranges that that are reserved for use by provate networks.  The VM (and therefore the registry) will exist within such a range.  This will effectively permit Docker pulls and pushes to such a registry.
 
-Install Docker Machine according to the normal installation instructions. Then create a new image for hosting your Deis Docker registry:
+## Fork the Repository
 
-    $ docker-machine create --driver virtualbox --virtualbox-disk-size=100000 \
-      --engine-insecure-registry=192.168.0.0/16 deis-registry
+Once the prerequisites have been met, we can begin to work with Deis components.
 
-This will create a new virtual machine named `deis-registry` that will take up as much as 100,000 MB of disk space. Registries tend to be large, so allocating a big disk is a good idea.
+Begin at Github by forking whichever Deis project you would like to contriubte to, then clone that fork locally.  Since Deis is predominantly written in Go, the best place to put it is under `$GOPATH/src/github.com/deis/`.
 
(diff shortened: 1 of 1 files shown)


docs/src/contributing/development-environment.md
```diff
@@ -1,280 +1,266 @@
...
-This document is for developers who are interested in working directly on the Deis codebase. In this guide, we walk you through the process of setting up a local development environment. While there are many ways to set up your Deis environment, this document covers a specific setup:
+We try to make it simple to hack on Deis components. However, there are necessarily several moving pieces and some setup required. We welcome any suggestions for automating or simplifying this process.
 
-- Developing on **Mac OSX** or **Linux**
-- Managing virtualization with **Vagrant/Virtualbox**
-- Hosting a docker registry with **docker-machine** (Mac)
-
-We try to make it simple to hack on Deis. However, there are necessarily several moving pieces and some setup required. We welcome any suggestions for automating or simplifying this process.
+!!! note
+    The Deis team is actively engaged in containerizing Go and Python based development environments tailored specifcally for Deis development in order to minimize the setup required.  This work is ongoing.
```
arschles: an (optional, of course) idea: link to some or all of the 'containerize the dev environment' issues we have on the repos so people can follow the work as it progresses? some of the repos (router, for example, as you know ;) already are fully, or almost fully, containerized

docs/src/contributing/development-environment.md
```diff
@@ -1,280 +1,266 @@
...
+    --driver=virtualbox \
+    --virtualbox-disk-size=100000 \
+    --engine-insecure-registry=192.168.0.0/16
+```
 
-This will create a new virtual machine named `deis-registry` that will take up as much as 100,000 MB of disk space. Registries tend to be large, so allocating a big disk is a good idea.
+This will create a new virtual machine named `deis-docker` that will take up as much as 100,000 MB of disk space. The images you build may be large, so allocating a big disk is a good idea.
 
-Once the deis-registry machine exists, source its values into your environment so your docker client knows how to use the new machine.
+Once the `deis-docker` machine exists, source its values into your environment so your docker client knows how to use the new machine. You may even choose to add this to your bash profile or similar.
```
arschles: sometimes on my machine, the machine comes up but the docker daemon takes 30-60 seconds more to become responsive. your call if you want to mention it.

docs/src/contributing/development-environment.md
```diff
@@ -1,280 +1,266 @@
...
 
-## Fork the Deis Repository
+## Fork the Repository
 
-Once the prerequisites have been met, we can begin to work with Deis.
+Once the prerequisites have been met, we can begin to work with Deis components.
 
-To get Deis running for development, first [fork the Deis repository][fork], then clone your fork of the repository. Since Deis is predominantly written in Go, the best place to put it is in `$GOPATH/src/github.com/deis/`.
+Begin at Github by forking whichever Deis project you would like to contriubte to, then clone that fork locally.  Since Deis is predominantly written in Go, the best place to put it is in `$GOPATH/src/github.com/deis/`.
```
arschles: nit: I'd say "the best place to put it is _under_..."

docs/src/contributing/development-environment.md
```diff
@@ -1,280 +1,266 @@
...
 
-    $ mkdir -p  $GOPATH/src/github.com/deis
-    $ cd $GOPATH/src/github.com/deis
-    $ git clone git@github.com:<username>/deis.git
-    $ cd deis
+```
+$ mkdir -p  $GOPATH/src/github.com/deis
+$ cd $GOPATH/src/github.com/deis
+$ git clone git@github.com:<username>/<component>.git
+$ cd <component>
```
arschles: maybe add instructions to set up a git remote for the deis repo (in case they have to rebase)?

arschles: nevermind, I read below...

docs/src/contributing/development-environment.md
krancour: Yup.  I've seen that.  I'll mention it.

docs/src/contributing/development-environment.md
krancour: Yes.  That sounds better.  That bit of verbiage carried over from the old doc, as you can see, but it's worth fixing now.

docs/src/contributing/development-environment.md
```diff
@@ -1,280 +1,266 @@
...
-    $ make dev-registry
+!!! important
+    Are you a new contributor to Deis? Your first [Pull Request][pr] could earn you credit at [DigitalOcean][]! Submit your changes and then email <deis@engineyard.com>. When your PR is merged, the maintainer team will send you a DigitalOcean credit based on the value of your contribution.
 
-    To configure the registry for local Deis development:
-        export DEIS_REGISTRY=192.168.59.103:5000
+### Installing Deis Components
 
-It is important that you export the `$DEIS_REGISTRY` environment variable as instructed.
+With a functioning Kubernetes cluster, procede with Deis installation using [these instructions](../installing-deis/digitalocean.md) and then continue to follow the instructions below.
```
arschles: spelling: proceed

docs/src/contributing/development-environment.md
```diff
@@ -1,280 +1,266 @@
...
 
-Running `deisctl list` should display all of the services that your Deis cluster is currently running.
+If your development cluster runs on a cloud provider such as DigitalOcean, a local registry such as the one above will not be accessible to your Kubernetes nodes.  In such cases, a public registry such as [DockerHub][dh] or [quay.io][quay] will suffice.
 
-You can now use your Deis cluster in all of the usual ways.
+To use DockerHub for this purpose, for instance:
 
-At this point, you are running Deis from the code in your Git clone. But since rebuilding like this is time consuming, Deis has a simplified developer workflow more suited to daily development.
+```
+$ export DEIS_REGISTRY=docker.io
```
arschles: iirc, pushing to `docker.io/...` won't work properly, and so `DEIS_REGISTRY` must be empty for that case

arschles:
@krancour a few nits/spelling notes and comments. ping me when fixed and I'll give this one more read. looks great in general

krancour:
@carmstrong when you look, can I ask you to comment on my changed naming of the Docker Machine VM to `deis-docker`?  I thought this sounded more appropriate since that VM is used to support Deis development in many way _beyond_ just hosting a dev registry.  For instance, even a simple `docker pull` or `docker build` is dependent upon that VM.

docs/src/contributing/development-environment.md
krancour: Nice catch.

docs/src/contributing/development-environment.md
krancour: Yup.  You're right.  This error was spotted in another PR, which erroneously led me to believe that might work.  It does not.  I will fix this.

docs/src/contributing/development-environment.md
krancour: Perhaps I'll just reference the router as a working example of such a containerized dev environment?

arschles: yea, that's perfect

docs/src/contributing/development-environment.md
```diff
@@ -1,280 +1,278 @@
...
+Deis needs Docker for building images.  Docker utilizes a client/server architecture, and while the Docker client is available for Mac OS, the Docker server is dependent upon the Linux kernel.  Therefore, in order to use Docker on Mac OS, [Docker Machine][machine] is used to facilitate running the Docker server within a [VirtualBox][vbox] VM.
 
-!!! note
-    Previously, Deis used boot2docker to run the registry. However, Docker has deprecated boot2docker in favor of Docker Machine.
+Install Docker Machine according to the normal installation instructions, then use it to create a new VM:
 
-Install Docker Machine according to the normal installation instructions. Then create a new image for hosting your Deis Docker registry:
+```
+$ docker-machine create deis-docker \
+    --driver=virtualbox \
```
carmstrong: Are we assuming that users have VirtualBox installed? I think it was removed as a dependency above.

docs/src/contributing/development-environment.md
```diff
@@ -1,280 +1,278 @@
...
+After following these steps, some Docker Machine users report a slight delay (30 - 60 seconds) before the Docker server is ready.
 
 !!! note
-    Because the registry that we create will not have a valid SSL certificate, we run the local registry as an insecure (HTTP, not HTTPS) registry. Each time Docker Machine reboots, the registry will get a new IP address somewhere in the `192.168.0.0/16` range. We must declare that explicitly when configuring Docker Machine.
-
-At this point, our `deis-registry` VM can now serve as a registry for Deis' Docker images. Later we will return to this.
+    In subsequent steps, you may run a Docker registry within the `deis-docker` VM. Such a registry will not have a valid SSL certificate and will use HTTP instead of HTTPS. Such registries are implicitly untrusted by the Docker server (which is also running on the `deis-docker` VM).  In order for the Docker server to trust the insecure registry, `deis-docker` is explicitly created to trust all registries in the `192.168.0.0/16` range - the range in which the VM (and therefore the registry) will exist.  This will effectively permit Docker pulls and pushes to such a registry.
```
carmstrong: Is the IP range deterministic? From memory, VirtualBox will create a new interface/bridge with an available range, which may not be the same across hosts.

docs/src/contributing/development-environment.md
```diff
@@ -1,280 +1,278 @@
...
-
-    $ cd $DEIS/client
-    $ make build
-    $ make install  # optionally
-    $ ./deis
-    Usage: deis <command> [<args>...]
+## Test Your Changes
 
-## Start Up a Development Cluster
+Deis components each include a comprehensive suite of automated tests, mostly written in Go. See [testing][] for instructions on running the tests.
```
carmstrong: Need a link target here

carmstrong: Ah, wait... maybe that's shorthand to mean the `testing` target? Cool.

krancour: Yup.

docs/src/contributing/development-environment.md
krancour: If that's the case, I was unaware.  That range was just lifted from the existing docs.  How do we test this to find out?

carmstrong: I would just add a note/caveat that this range may not be the one used in all cases, and to confirm with the machine's IP.

krancour: There might be a slicker way around this.  We could change the line used for creating the VM to mention all the IP ranges reserved for private networks by RFC 1918.  Then it doesn't matter if VirtualBox puts the VM somewhere besides 192.168.0.0/16.

docs/src/contributing/development-environment.md
```diff
@@ -1,280 +1,278 @@
...
-
-This document is for developers who are interested in working directly on the Deis codebase. In this guide, we walk you through the process of setting up a local development environment. While there are many ways to set up your Deis environment, this document covers a specific setup:
+We try to make it simple to hack on Deis components. However, there are necessarily several moving pieces and some setup required. We welcome any suggestions for automating or simplifying this process.
 
-- Developing on **Mac OSX** or **Linux**
-- Managing virtualization with **Vagrant/Virtualbox**
-- Hosting a docker registry with **docker-machine** (Mac)
-
-We try to make it simple to hack on Deis. However, there are necessarily several moving pieces and some setup required. We welcome any suggestions for automating or simplifying this process.
+!!! note
```
arschles: I missed this typo before - `specifically`

docs/src/contributing/development-environment.md
```diff
@@ -1,280 +1,278 @@
...
-You can also use the same tasks on the root `Makefile` to operate on all components at once.  For example, `make deploy` will build, dev-release, and restart all components on the cluster.
+Once your customized Deis component has been deployed, here are some helpful commands that will allow you to inspect your cluster and to troubleshoot, if necessary:
 
-!!! important
-    In order to cut a dev-release, you must commit changes using `git` to increment the SHA used when tagging Docker images.
+### See All Deis Pods
 
-## Test Your Changes
+```
+$ kubectl get pods --namespace=deis
```
arschles: optional: talk about creating an `alias` to `kubectl --namespace=deis`

docs/src/contributing/development-environment.md
krancour: Fixed.

arschles:
@krancour gave LGTM. I don't see any other issues than the one you're already talking about with @carmstrong

docs/src/contributing/development-environment.md
carmstrong: :+1: This is what we do for CoreOS hosts and the builder. https://github.com/deis/deis/blob/f46b967e9fd7c2260cddad656393d7817bcd0897/contrib/coreos/user-data.example#L54

krancour:
@arschles thanks for the review and LGTM.  Just to double-check one thing... you're not aware of any dev tools used for any of the projects that aren't sufficiently covered here, are you?  This doc endeavors to generically cover what is required to hack on _all_ the components and admittedly, I have not gone deep into each and every one.

carmstrong:
Overall, this is a great improvement! Excellent work, @krancour. LGTM after my comments are addressed.

arschles:
@krancour nope, I don't know of any other dev tools except vagrant, which @carmstrong brought up already

krancour:
Wait... what about vagrant?  I don't see a comment about that.

arschles:
@krancour gah, sorry meant vbox: https://github.com/deis/workflow/pull/180#discussion_r48292497

arschles:
however, you do mention vagrant in a few places, so maybe it'd be a good idea to mention how that fits into the picture. not a blocker for this PR

krancour:
They're just sort of passing mentions of Vagrant.  There's nothing that we've documented here that _requires_ Vagrant.  Running a k8s cluster within Vagrant is just one possibility.  It's mentioned as such, but I see no reason to go into greater depth than that.  If someone chooses that route, the procedure and prereqs are documented elsewhere.

docs/src/contributing/development-environment.md
krancour: Ya.  I was originally going to do that here to, and then started to think it wasn't necessary.  But if you're right that vobx might put the VM in a different IP range, this is the best way to account for that possibility.  Doing it now.

krancour: This is fixed.

docs/src/contributing/development-environment.md
krancour: I'll document it as a conditional dependency-- required _if_ you're also in need of Docker Machine.  Folks on Linux or developing using a public dev registry won't need either of those.

krancour: Actually, it's mentioned and linked to on line 55.  Is that sufficient?

docs/src/contributing/development-environment.md
krancour: Ah... whoops.  I fixed the command, but not the note.  Fixing the fix...

krancour: Fixed for realsies now.

docs/src/contributing/development-environment.md
krancour: It's a clever trick and I think we're all using it, but I'm inclined to say we should stick to what's required for development and not embellish too much.

krancour:
@carmstrong all of your feedback is now accounted for if you want to take another gander at your convenience.

docs/src/contributing/development-environment.md
carmstrong: See it now. :+1:

docs/src/contributing/development-environment.md
carmstrong: Agreed.

carmstrong:
:shaved_ice:

carmstrong:
Also, :ship:

slack:
You are an :angel:

arschles:
looks like this needs a rebase again :cry:

carmstrong:
> looks like this needs a rebase again

I think you and @helgi are bugging out - this is mergeable to me

arschles:
k. guess GH doesn't like docs ;)

Issue #183, opened 2015-12-22
[meta] Ginkgo Test Goals


technosophos:
![Moving the Goal Posts](https://warosu.org/data/ck/img/0054/34/1400084863723.jpg)

For the alpha release, these are our test goals:

- [x] translate the full suite of `deis` tests to Ginkgo
- [x] be able to run all of those tests on the existing codebase
- [x] tests that are _part of alpha_ must pass (e.g. `deis log` tests needn't pass, since logging isn't part of alpha1)

The following are not required:
- fully automated builds
- porting unit tests to Ginkgo

arschles:
This issue is now complete to the best of my knowledge, so closing. See details below.

- https://github.com/deis/workflow/pull/189 (which was migrated to deis/workflow in https://github.com/deis/workflow-e2e/pull/3) has all alpha related tests passing
- https://github.com/deis/workflow-e2e/pull/4 has healthcheck tests running and passing
- https://github.com/deis/workflow-e2e now houses all of our end to end tests, and all tests in this repository have been moved there.
- https://github.com/deis/workflow/pull/232 is the last step of the aforementioned migration

PR #188, opened 2015-12-22
fix(releases_test.go): uncomment integration tests for releases

arschles:
makes progress on #183


arschles:
note that it's expected that these pass before merge

PR #191, opened 2015-12-23
ref(_tests): use gexec for executing commands

arschles:
This PR represents a partial migration to use [gexec](https://godoc.org/github.com/onsi/gomega/gexec), Gomega's built-in facilities for executing commands. The main benefits are:

- Use built-in facilities for executing commands
- Ability to capture stdout and stderr in a centralized location (should be done in a later PR)
- Error reporting is standard and much clearer

Note that this PR does not completely migrate the test suite to use the new gexec-based code. Further migrations should be done incrementally and in future patches.

I've marked this a showstopper because the the current test case failure messages do not show the output of most failures, and the person running the test usually will need to insert their own debugging statements to proceed.

cc/ @gabrtv @michelleN 

Ref #183


_tests/apps_test.go
```diff
@@ -14,37 +14,34 @@ var _ = Describe("Apps", func() {
...
 		It("can create an app", func() {
-			output, err := execute("deis apps:create %s", app1Name)
-			Expect(err).NotTo(HaveOccurred())
-			Expect(output).To(SatisfyAll(
+			Expect(execute("deis apps:create %s", app1Name)).To(BeASuccessfulCmdWithOutput(
 				ContainSubstring("Creating Application... done, created %s", app1Name),
 				ContainSubstring("Git remote deis added"),
-				ContainSubstring("remote available at ")))
-			output, err = execute("deis apps:destroy --confirm=%s", app1Name)
-			Expect(err).NotTo(HaveOccurred())
```
bacongobbler: Shouldn't we still be checking if there's an error running these commands? The new commands just seem to check that the output is what's expected. What about the command exiting non-zero?

_tests/auth_test.go
```diff
@@ -12,10 +12,10 @@ var _ = Describe("Auth", func() {
 		})
 
 		It("won't print the current user", func() {
-			output, err := execute("deis auth:whoami")
-			Expect(err).To(HaveOccurred())
-			Expect(output).To(ContainSubstring("Not logged in."))
-			Expect(output).NotTo(ContainSubstring(testUser))
+			Expect(execute("deis auth:whoami")).To(BeASuccessfulCmdWithOutput(
```
bacongobbler: `BeASuccessfulCmdWithOutput` seems odd here, considering that the command should expect to have exited non-zero with the expected substring.

arschles:
@bacongobbler made significant changes to this PR, which incidentally address both of your comments.

bacongobbler:
*much* better :)

michelleN:
:+1:

PR #192, opened 2015-12-23
fix(tests_suite_test.go): seed the rng, use true random numbers for users, passwords and emails

arschles:
not strictly necessary for `v2.0-alpha1`, but the additional randomness would help with concurrent test runs on the same deis cluster

### DIFF
--- a/_tests/tests_suite_test.go
+++ b/_tests/tests_suite_test.go
@@ -37,12 +37,13 @@ func TestTests(t *testing.T) {
 }
 
 var (
-	testAdminUser     = fmt.Sprintf("test-admin-%d", rand.Intn(1000))
+	randSuffix        = rand.Intn(1000)
+	testAdminUser     = fmt.Sprintf("test-admin-%d", randSuffix)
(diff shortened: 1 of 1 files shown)


mboersma:
I intended to use one random number for both app, user, and test-user suffixes. This helps to correlate results after a test run. And our past experience shows that having a fake password such as "asdf1234" will be valuable if a run fails and you want to investigate the cluster state interactively.

arschles:
@mboersma modified to use the same random number for all suffixes

michelleN:
getting this error: 
```
_tests [rand-seed]$ make test-integration
go test -v ./...
# github.com/deis/workflow/_tests_test
./tests_suite_test.go:13: imported and not used: "github.com/onsi/ginkgo/config"
FAIL    github.com/deis/workflow/_tests [build failed]
?       github.com/deis/workflow/_tests/example-go    [no test files]
make: *** [test-integration] Error 2
```

arschles:
@michelleN fixed, thanks

PR #197, opened 2015-12-23
feat(k8s): use dotted format instead of JSON for router domain configuration

### LINKED ISSUE #196: Move from JSON annotations to dotted format for router configuration
See deis/router#63 for the main discussion around this

Format will be as follow:

```
router.deis.io/domains = 'testing,helgi.com'
router.deis.io/gzip.things = 'vary'
```

helgi:
Fixes #196

Router needs to have a PR that understands this format

### DIFF
--- a/rootfs/api/models.py
+++ b/rootfs/api/models.py
@@ -12,8 +12,8 @@ import importlib
 import logging
 import re
 import time
-import json
 import uuid
+import morph
 from threading import Thread
 
 from django.conf import settings
@@ -1001,7 +1001,7 @@ class Domain(AuditedModel):
                                    settings.SCHEDULER_AUTH,
                                    settings.SCHEDULER_OPTIONS)
 
-    def _load_service_config(self, app):
+    def _fetch_service_config(self, app):
         # Get the service from k8s to attach the domain correctly
         svc = self._scheduler._get_service(app, app).json()
         # Get minimum structure going if it is missing on the service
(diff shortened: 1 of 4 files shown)


helgi:
```
metadata:
  annotations:
    router.deis.io/domains: helgi.com,helgi2.com
```

krancour:
Just remembered... using comma delimited strings in place of actual lists is also applicable for app IP/CIDR whitelists.

helgi:
yeah, it will have to be dealt with on case by case basis, for now at least. load / save is generic and so is flatten / unflatten but not the list portion

When I deal with `config:set` I may special case it or try to do some sort of a detection logic around it all

krancour:
Is `deis config:set` still used for managing whitelists?  I'm not a big fan of that.  I know this was v1.x behavior, but I didn't like it then either.  My argument against it is that setting _application_ environment variables is the wrong way to affect _platform_ behavior.  i.e. Setting the `WHITELIST` env var on the app does not affect the application's configuration / behavior at all; rather it affects the router.

I would very much love to see `whitelist` actually added to workflow's app model, with complimentary API endpoints and CLI subcommands for managing it.

helgi:
Not sure? I'm just assuming that's one way of applying `annotations` in general tho - I build this PR to be very generic for the most part so there isn't much stopping us from keeping `config:set` that does it on `platform` or `router`, as in no specific application defined.

As far as the `whitelist` part it self, that's more a "where is the data stored and associated with" type of question. And it should probably be on the app model and it probably should go into the app specific whitelist. A tangent: I also believe there needs to be a whitelist that is global and not just app level. Basically we need both, but that's really a diff ticket... Feel free to open up a ticket somewhere so we can start a proper discussion on where `whitelist` lives and behaves :)

krancour:
@helgi I've tested this against deis/router#72 and it all checks out.  LGTM

rootfs/api/models.py
```diff
@@ -1009,32 +1009,53 @@ def _load_service_config(self, app):
...
-        if 'domains' not in config:
-            config['domains'] = []
+    def _load_service_config(self, app, component):
+        # fetch setvice definition with minimum structure
+        svc = self._fetch_service_config(app)
+
+        # always assume a .deis.io/ ending
+        component = "%s.deis.io/" % component
+
+        # Filter to only include values for the component andstrip component out of it
```
arschles: nit: space between `and` and `strip`

helgi: added

arschles:
LGTM (the nit I added is obviously not blocking for this PR), but I haven't tested myself, so relying on @krancour's testing

Issue #185, opened 2015-12-22
Lots of docs still reference deis/deis


krancour:
There are many points in the documentation where references or links to the deis/deis project should be replaced either with references or links to deis/workflow _or_ (as appropriate) with language that acknowledges the existence of additional projects and repositories for individual Deis components.

Relates to #161

bacongobbler:
I think the jury's still out if we're going to merge this back into deis/deis at some point before 2.0 so we should figure that out first :)

krancour:
That's easy enough to fix later.  For now though, this is a "where there's smoke, there's fire" sort of thing.  References to deis/deis probably coincide with other stale bits of doc.  I've been on a doc bender, but these would be pieces that probably still need love.

bacongobbler:
is this good to close now?

krancour:
Let me check on this.

krancour:
Most remaining references to deis/deis are links to specific issues.  Some of those may be relevant still and others, not so much.  Apart from that, there looks to be just a few remaining references to deis/deis.  It is the roadmap that is most in need of cleanup.

bacongobbler:
sounds good. I found the ones you're referencing by just scrubbing with good ol' `git grep`, so I'll clean those up/

krancour:
@bacongobbler, @technosophos has suggested that we let @slack handle the roadmap changes.  I'm going to close this and open a separate issue for that.

PR #199, opened 2015-12-23
doc(contributing): update contributing.md

krancour:
Partially addresses both #161 and #185 

`contributing.md` was a bit stale.  It needed updates to reference additional support channels (multiple issue queues for individually repo'ed Deis components).  While editing, I noticed tremendous overlap with other documentation in `docs/src/contriubting`, so I simplified `contributing.md` to be a quicker read that references the more detailed docs where applicable.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,101 +1,43 @@
 # How to Contribute
 
-The Deis project is Apache 2.0 licensed and accepts contributions via Github pull
-requests. This document outlines some of the conventions on commit message formatting,
-contact points for developers and other resources to make getting your contribution
-accepted.
+Deis components are Apache 2.0 licensed and accept contributions via Github pull requests. This document outlines resources and conventions useful for anyone wishing to contribute either by reporting issues or submitting pull requests.
 
 # Certificate of Origin
 
-By contributing to this project you agree to the
-[Developer Certificate of Origin (DCO)][dco]. This document was created by the Linux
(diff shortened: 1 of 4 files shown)


CONTRIBUTING.md
```diff
@@ -1,101 +1,40 @@
...
-has already reported the issue you're facing, or it's a known issue that we're already aware of.
+Before opening a new issue or PR against any Deis project, it's helpful to search that project's issue queue - it's likely that another user has already reported the issue you're facing, or it's a known issue that we're already aware of.
 
-Additionally, see the [Troubleshooting Deis][troubleshooting] documentation for common issues.
+Additionally, see the [Troubleshooting][] documentation for common issues.
 
 Our official support channels are:
 
-- GitHub issues: https://github.com/deis/deis/issues/new
+- GitHub issue queues:
```
helgi: I would throw in https://github.com/pulls?utf8=%E2%9C%93&q=is%3Aopen+user%3Adeis+user%3Ahelm as a "one view to see all these issues" kind of dealio - Tho it includes `deis/deis` and I don't know how to remove it

krancour: I'll add that.  I think that's useful.  But I think the list is also useful since if someone wants to _open_ an issue, they need to do it from within the right repo.

helgi: both is good

krancour: @helgi I just added it.  Would you just spot-check that one change?

helgi:
:+1:

PR #202, opened 2015-12-23
ref(web): remove web 'app' from django

helgi:
There is no practical use for this module and in later Django updates (see my other django PRs) there are some slight hoop jumping to do.

Any web UI on top of Deis should be built via the APIs

This is only a proposal but given we do not touch it much or pay attention to it then I think we are best served without it, even if we may lose the potential admin UI


krancour:
> Any web UI on top of Deis should be built via the APIs

:+1:

PR #204, opened 2015-12-23
doc(planning-proces): update planning process docs

krancour:
Partially addresses #185 

Requires #201 to be merged first.


krancour:
This is failing CI, but it will pass after #201 is merged.

Issue #207, opened 2015-12-24
Remove LDAP in the initial few releases of 2.x


helgi:
Been looking at our code base to see what components (especially within workflow) need to be worked on and rewritten due to the removal of `etcd` in the future and noticed that the LDAP integration uses `etcd` heavily to get various configuration elements, some of which are rather sensitive, such as `password`.

I'd like us to drop LDAP support from 2.0 (and maybe keep it that way for a few point releases) *temporarily* to simplify the `etcd` surgery and allow us to rethink the auth plugin strategy in relation to how we want to expose configuration of additional auth backends.

Something to be aware of is that I had to remove `django-ldap-auth` (see #101) from the Django 1.8 upgrade branch as it seems to not be compatible with newer Django's so LDAP is also preventing us from moving forward with Django upgrade.

Note that LDAP is described as user contributed and not supported by the deis team in the documentation.

slack:
I'm ok with dropping LDAP support. We can look at re-adding in a point release with a broader look at identity and federation.

PR #208, opened 2015-12-26
ref(ldap): remove LDAP temporarily

helgi:
LDAP poses a problem with the upgrade path to Django 1.9 as the modules involved are not compatible.
Removing until we have a better identity and federation story. And the django plugin works again

Ref #207


arschles:
makes sense. :+1:

PR #210, opened 2015-12-28
docs(quickstart): document how to configure deis client

### LINKED ISSUE #102: Document how to configure deis and git clients
There should be a document that outlines how to configure the `deis` and `git` clients for at least the following Deis installation kinds:

1. When Deis is `helm install`ed onto a k8s cluster running on a load balancer capable cloud provider
2. When Deis is `helm install`ed onto a k8s cluster running on infrastructure where external load balancer creation is a no-op (e.g. bare metal or virtualbox)
3. When only workflow and builder are running (this will require app creation against the workflow cluster IP, then setting up a git remote directly to the builder cluster IP)

This issue follows from https://github.com/deis/router/issues/58.

bacongobbler:
closes #102

### DIFF
--- a/docs/src/installing-deis/quickstart.md
+++ b/docs/src/installing-deis/quickstart.md
@@ -22,8 +22,14 @@ See [Configuring DNS][] for more information on properly setting up your DNS rec
 
 Now that you've finished provisioning a cluster, please [Install the Deis Platform][install deis].
 
+## Register a User
 
+Once your cluster has been provisioned and the Deis Platform has been installed, you can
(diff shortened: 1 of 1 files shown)


bacongobbler:
okay I filed deis/router#67 which will address the rest of the concerns in the docs. Back and ready for review

arschles:
simple enough, thanks @bacongobbler

arschles:
@bacongobbler I added a `needs rebase` label earlier, but have discovered that I might be getting wrong info from GH...

PR #212, opened 2015-12-28
fix(tests_suite_test.go): cancel accounts gracefully in the AfterSuite block

arschles:
This patch makes the code in `AfterSuite` attempt to delete both users, even if the first fails. Combined with https://github.com/deis/workflow/pull/191, this will at least help with error reporting and graceful cleanup

cc/ @lanej @michelleN

### DIFF
--- a/_tests/tests_suite_test.go
+++ b/_tests/tests_suite_test.go
@@ -3,7 +3,6 @@ package _tests_test
 import (
 	"bytes"
 	"fmt"
-	"io"
 	"math/rand"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


michelleN:
sweeet. I hope this helps with weird states relating to after steps not running. I also wonder what the root cause of the user deletion error was in the first place.

arschles:
update on this: after following [instructions for resetting apps & users in the README](https://github.com/deis/workflow/blob/master/_tests/README.md#resetting-cluster-state), I got this result:

```console
FAIL! -- 23 Passed | 15 Failed | 14 Pending | 0 Skipped --- FAIL: TestTests (23.92s)
```

arschles:
after, https://github.com/deis/workflow/commit/fade18873539c6ec1a70d45d24d8a57ccdc8fe74, the tests clean up more consistently

PR #215, opened 2015-12-28
ref(docs): clean up quickstart documentation

bacongobbler:
Moving some stuff around. Specifically:

 - "Configuring DNS" should now occur after the platform is installed so users can
   deal with it once the routers are up
 - the load balancer configuration was all duplicated information (Managing Deis
   covers this), so it was moved/removed
 - DNS configuration was moved over to "Configuring DNS"
 - removed "Next Steps" from "Installing the Deis Platform" as the next steps are
   described in the Quickstart doc
 - removed the documentation around "Set up the Deis Router". This is no longer necessary
   for a "Quick Start" and should go into a "Configuring Router" document in the future (see deis/router#67)


bacongobbler:
also dependent on https://github.com/deis/charts/pull/52

arschles:
:+1: for the greatly simplified setup (and lots of red)

docs/src/installing-deis/installing-the-deis-platform.md
```diff
@@ -71,137 +55,6 @@ $ kubectl get pods --namespace=deis
 
 Once you see all of the pods in the `READY` state, your Deis platform is running on a cluster!
 
-## A Note on Routing Traffic
```
slack: This should have made it to deis/router#67, no?

docs/src/managing-deis/configuring-dns.md
```diff
@@ -18,16 +18,67 @@ Apps can then be accessed by browsers at `appname.myapps.com`, and the controlle
...
+This will get you started, though you may find that you have to manually maintain this file as you
+create and deploy more applications to the cluster.
+
+## Testing
+
+To test that traffic reaches its intended destination, a request can be
+sent to the Deis controller like so:
+
+```
+curl http://deis.<platform domain>/v2/
```
slack: With this change, this is the first time `platform domain` is referenced in this doc. Should be `example.com` I think.

docs/src/installing-deis/installing-the-deis-platform.md
bacongobbler: couldn't. Two separate projects = two separate PRs

docs/src/managing-deis/configuring-dns.md
bacongobbler: this was a tie-over from the old docs (see https://github.com/deis/workflow/pull/215/files#diff-3122c85f28ce9c670a6d843bc29638feL191). I'm more than happy to change it to example.com though :)

bacongobbler:
okay I addressed your comments, @slack. Should be good to go!

Issue #111, opened 2015-12-17
deis open opens http://<app name>.localhost in browser


sgoings:
After creating a cluster, registering a deis client, and doing a `deis pull` type deployment... when I run:

```
../workflow/client/deis create --no-remote
Creating Application... done, created calmer-handbill
remote available at ssh://git@deis.$IP.xip.io:2222/calmer-handbill.git
 ~/go/src/github.com/deis/example-go  ⑂ master    ../workflow/client/deis pull deis/example-go -a calmer-handbill
Creating build... done
 ~/go/src/github.com/deis/example-go  ⑂ master    ../workflow/client/deis open -a calmer-handbill
```

My browser opens with the URL: http://calmer-handbill.localhost/

It should open to http://calmer-handbill.$IP.xip.io/

mboersma:
Is this code still using the `/deis/platform/domain` key? That defaults to `localhost`. If so, we need either to update that mechanism not to use `etcd`, or document how to set that key correctly.

bacongobbler:
Huh, I was under the assumption that it'd use the base domain that you used when you initially ran `deis login`, but good catch!

mboersma:
Confirmed: if you `etcdct set /deis/platform/domain 10.245.1.3.xip.io` (or the appropriate IP), then `deis open` works correctly again, so the workflow controller is using the etcd key for this purpose.

krancour:
So... my $0.02... you have to tell the router the platform domain abd the router doesn't use etcd at all... so making people set an etcd key is a DRY exception we'd be forcing on our users.  I like the idea of the CLI deriving the platform domain automatically by scraping the `deis.` off the front of the controller's base URL and using what's left.

helgi:
Looked at this a little bit, problem is that we use `ALLOWED_HOST = *` in Django so using `get_host` or similar mechanisms are spoofable, depends if we care about that. That'd be determine if we do any sort of scraping or not

For now I'm going to make it so you can set a `DEIS_CONTROLLER_DOMAIN` and `DEIS_CONTROLLER_SUBDOMAIN` env var as a stop gap measure until we decide if we want people to add the `platform domain` on BOTH router and workflow

krancour:
I really think there's no need for that.  Workflow does not need to know the platform domain.  Only the router and the client do.  The router knows the platform domain already (maybe; @bacobgobbler made it optional recently) but workflow and builder shouldn't need to know it...

[...]

PR #229, opened 2015-12-30
ref(*): remove the notion of a platform domain known to workflow

helgi:
Addresses the workflow side of #111 but doesn't do the client part so that it opens up the right domain

### DIFF
--- a/client/controller/api/apps.go
+++ b/client/controller/api/apps.go
@@ -6,7 +6,7 @@ type App struct {
 	ID      string `json:"id"`
 	Owner   string `json:"owner"`
 	Updated string `json:"updated"`
-	URL     string `json:"url"`
+	URL     string `json:"-"`
 	UUID    string `json:"uuid"`
 }
 
--- a/client/controller/models/apps/apps.go
+++ b/client/controller/models/apps/apps.go
@@ -23,6 +23,12 @@ func List(c *client.Client, results int) ([]api.App, int, error) {
 		return []api.App{}, -1, err
 	}
 
+	for name, app := range apps {
+		// Add in app URL based on controller hostname, port included
+		app.URL = fmt.Sprintf("%s.%s", app.ID, c.ControllerURL.Host)
+		apps[name] = app
+	}
+
 	return apps, count, nil
 }
 
@@ -51,6 +57,9 @@ func New(c *client.Client, id string) (api.App, error) {
 		return api.App{}, err
 	}
 
+	// Add in app URL based on controller hostname, port included
+	app.URL = fmt.Sprintf("%s.%s", app.ID, c.ControllerURL.Host)
+
 	return app, nil
 }
 
@@ -69,6 +78,9 @@ func Get(c *client.Client, appID string) (api.App, error) {
 	if err = json.Unmarshal([]byte(body), &app); err != nil {
 		return api.App{}, err
 	}
+
+	// Add in app URL based on controller hostname, port included
+	app.URL = fmt.Sprintf("%s.%s", app.ID, c.ControllerURL.Host)
 
 	return app, nil
 }
--- a/client/controller/models/apps/apps_test.go
+++ b/client/controller/models/apps/apps_test.go
@@ -21,7 +21,6 @@ const appFixture string = `
     "owner": "test",
     "structure": {},
     "updated": "2014-01-01T00:00:00UTC",
-    "url": "example-go.example.com",
     "uuid": "de1bf5b5-4a72-4f94-a10c-d2a3741cdf75"
 }`
 
@@ -37,7 +36,6 @@ const appsFixture string = `
             "owner": "test",
             "structure": {},
             "updated": "2014-01-01T00:00:00UTC",
-            "url": "example-go.example.com",
             "uuid": "de1bf5b5-4a72-4f94-a10c-d2a3741cdf75"
         }
     ]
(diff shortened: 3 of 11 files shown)


helgi:
Anyone know if the build hook `domains` portion is used anywhere? and if so by what?

helgi:
The is ready from the workflow side of things so it could be merged and the `client` would just crap out until that is fixed.

Would be good to know what the build hook is doing tho

bacongobbler:
the build hook is [called from the builder](https://github.com/deis/builder/blob/20ad31ba8b058aebcbed3efc03b43d71a7cde9e8/rootfs/etc/confd/templates/builder#L177-L179) so that it knows [what URL to display](https://github.com/deis/builder/blob/20ad31ba8b058aebcbed3efc03b43d71a7cde9e8/rootfs/etc/confd/templates/builder#L188-L191)

helgi:
@bacongobbler Can it get that from any other way now that we are on k8s? or would the changed messaging be enough to resolve that?

bacongobbler:
I think the changed messaging would be enough. We'd have to know how they set up their routing information on the router, but this is a quick and easy change so long as we update the client (`deis open`) to work with this. +1

arschles:
+1 on changing the messaging for now. However, shouldn't the client be updated in this PR as well?

helgi:
Probably, maybe today is the day I write my first `go` code... :)

krancour:
I'm not so good with the python... but I am very much in favor of this change.

bacongobbler:
let's add in the client change as well and we should be good to go. Controller side of things LGTM though.

arschles:
controller side looks good to me as well

helgi:
First stab at the `go` code for having the app url internally generated based on the app name and the controller host

.travis.yml
```diff
@@ -18,8 +18,8 @@ sudo: required
 addons:
   postgresql: "9.3"
 before_install:
-  - wget "https://github.com/Masterminds/glide/releases/download/0.8.1/glide-0.8.1-linux-amd64.tar.gz"
-  - sudo tar -vxz -C /usr/local/bin --strip=1 -f glide-0.8.1-linux-amd64.tar.gz
+  - wget "https://github.com/Masterminds/glide/releases/download/0.8.3/glide-0.8.3-linux-amd64.tar.gz"
+  - sudo tar -vxz -C /usr/local/bin --strip=1 -f glide-0.8.3-linux-amd64.tar.gz
```
krancour: Is this strictly required as part of this PR?  There may be a reason for it, which is why I ask.  Otherwise, this should be a separate chore.

helgi: Couldn't get it installing glide.yaml without it. Potentially because I have `0.8.3` locally, it could have messed that up but reverting glide to `0.8.1` seemed like an overkill

krancour: I understand, but while this fixes it for you, it might break it for someone else.  This seems like impetus to containerize the build / test of the go client using deis/docker-go-dev.  That would insulate all contributors from these sort of environmental differences.

krancour: But that would be a separate task, obviously.

helgi: #249 addresses `glide`

krancour:
Go client changes LGTM.  And the Python changes are simple enough that I looked at those as well.

bacongobbler:
Just a few more things need to be done before we can merge this:

 - [ ] language change in deis/builder
 - [ ] amend your second commit to follow our style guide

After that, it's good to :shipit:

bacongobbler:
okay deis/builder#77 addresses the language changes in the builder :)

helgi:
fixes #111

client/controller/api/apps.go
```diff
@@ -6,7 +6,7 @@ type App struct {
 	ID      string `json:"id"`
 	Owner   string `json:"owner"`
 	Updated string `json:"updated"`
-	URL     string `json:"url"`
+	URL     string
```
arschles: @helgi can you either name this field with a struct tag (basically, add the `json:"url"` back) or add a `json:"-"` if it's supposed to be ignored?

arschles:
left 1 comment on client changes

client/controller/api/apps.go
helgi: This one I wasn't too sure about - Since the `URL` is no longer coming from the API and I am overwriting/setting it in the models within the client, do I even need to keep it in the original struct?

arschles: Yea, keep the struct for now. You're decoding JSON into all of the struct except for `URL`. The `json:"-"` is a [struct tag for the JSON encoder](https://golang.org/pkg/encoding/json/#Marshal) (and decoder) that tells the marshaller/unmarshaler to ignore `URL` completely.

So, in short, `json.Marshal` and `json.Unmarshal` will ignore `URL` completely, and you can set it afterward.

helgi: Done - Why does it have to be defined as `json` to be ignored right away oppose to a straight up`string` ?

arschles: the tl;dr is that since some of the struct fields have tags and `URL` doesn't, the [marshaler rules](https://godoc.org/encoding/json#Marshal) state that `URL` will be ignored.

Here's the longer version, from the docs. The first rule below is in action here:

```
The Go visibility rules for struct fields are amended for JSON when deciding which field to marshal or unmarshal. If there are multiple fields at the same level, and that level is the least nested (and would therefore be the nesting level selected by the usual Go rules), the following extra rules apply:

1) Of those fields, if any are JSON-tagged, only tagged fields are considered, even if there are multiple untagged fields that would otherwise conflict. 2) If there is exactly one field (tagged or not according to the first rule), that is selected. 3) Otherwise there are multiple fields, and all are ignored; no error occurs.
```

arschles: Also, see [this code sample](http://play.golang.org/p/Qzfal52SJD) for an illustration of the case when `"URL"` is present and not present in the incoming JSON

Issue #154, opened 2015-12-19
deis limits doesn't work


smothiki:
running 3 minions with 2GB each.
```
$ deis limits:set web=2GB -a quacky-waterbed
$ deis scale web=6
Scaling processes... but first, coffee!
done in 97s
=== quacky-waterbed Processes
--- web:
web.1 up (v3)
web.2 up (v3)
web.3 up (v3)
web.4 up (v3)
web.5 up (v3)
web.6 up (v3)
```
should be able to scale 3 or revert scale operation

bacongobbler:
I believe `deis limits` was off the table for alpha. `config` and `domains` are OK though.

smothiki:
Yes. I was just creating issues to keep track of things

PR #243, opened 2016-01-06
feat(workflow): add support to set memory and cpu configs

smothiki:
fixes https://github.com/deis/workflow/issues/154

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -622,8 +622,8 @@ class KubeHTTPClient(AbstractSchedulerClient):
         containers[0]['args'] = args
         loc = locals().copy()
         loc.update(re.match(MATCH, container_fullname).groupdict())
-        mem = kwargs.get('memory', {}).get(loc['c_type'])
-        cpu = kwargs.get('cpu', {}).get(loc['c_type'])
+        mem = kwargs.get('memory', {}).get(app_type)
+        cpu = kwargs.get('cpu', {}).get(app_type)
         env = kwargs.get('envs', {})
         if env:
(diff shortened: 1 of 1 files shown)


rootfs/scheduler/__init__.py
```diff
@@ -622,23 +622,22 @@ def _create_rc(self, name, image, command, **kwargs):
         containers[0]['args'] = args
         loc = locals().copy()
         loc.update(re.match(MATCH, container_fullname).groupdict())
-        mem = kwargs.get('memory', {}).get(loc['c_type'])
-        cpu = kwargs.get('cpu', {}).get(loc['c_type'])
```
bacongobbler: we can just interchange this with

```
kwargs.get('memory', {}).get(app_type)
```

It's best to set it here rather than in the `if` block so future rewrites have the correct data.

smothiki: sure

rootfs/scheduler/__init__.py
```diff
@@ -622,23 +622,22 @@ def _create_rc(self, name, image, command, **kwargs):
...
+        mem = kwargs.get('memory', {})
+        cpu = kwargs.get('cpu', {})
         env = kwargs.get('envs', {})
         if env:
             for k, v in env.iteritems():
                 containers[0]["env"].append({"name": k, "value": v})
         if mem or cpu:
             containers[0]["resources"] = {"limits": {}}
-
```
bacongobbler: would you mind keeping these formatting changes out of this PR? I'm sure @helgi has a few words to say about these changes ;)

PR #245, opened 2016-01-06
ref(*): remove etcd now that all state is being handled by db/k8s

helgi:
This switches `confd` to source in `env` var for `gunicorn` and a few other minor things, for now

Following items need to be figured out and #9 touches on some of that

* [x] etcd_set_default protocol "${DEIS_PROTOCOL:-http}"
* [x] etcd_set_default secretKey "${DEIS_SECRET_KEY:-$(openssl rand -base64 64 | tr -d '\n')}"
* [x] etcd_set_default builderKey "${DEIS_BUILDER_KEY:-$(openssl rand -base64 64 | tr -d '\n')}"
* [x] Builder reading app names, users and keys from an alternative location

Depends on #242 as it ditches one of the `etcd` requirement that is embedded in the models

### DIFF
--- a/rootfs/api/models/__init__.py
+++ b/rootfs/api/models/__init__.py
@@ -3,7 +3,6 @@
 """
 Data models for the Deis API.
 """
-import etcd
 import importlib
 import logging
 import uuid
@@ -16,8 +15,8 @@ from django.db.models.signals import post_delete, post_save
 from django.dispatch import receiver
 from django.core.exceptions import ValidationError
 from rest_framework.authtoken.models import Token
-from api.utils import fingerprint, dict_merge
 
+from api.utils import dict_merge
 
 logger = logging.getLogger(__name__)
 
@@ -40,20 +39,6 @@ def validate_label(value):
     match = re.match(r'^[a-z0-9-]+$', value)
     if not match:
         raise ValidationError("Can only contain a-z (lowercase), 0-9 and hypens")
-
-
-def get_etcd_client():
-    if not hasattr(get_etcd_client, "client"):
-        # wire up etcd publishing if we can connect
-        try:
-            get_etcd_client.client = etcd.Client(
-                host=settings.ETCD_HOST,
-                port=int(settings.ETCD_PORT))
-            get_etcd_client.client.get('/deis')
-        except etcd.EtcdException:
-            logger.log(logging.WARNING, 'Cannot synchronize with etcd cluster')
-            get_etcd_client.client = None
-    return get_etcd_client.client
 
 
 class AuditedModel(models.Model):
@@ -184,48 +169,6 @@ def _log_cert_removed(**kwargs):
     logger.info("cert {} removed".format(cert))
 
 
-def _etcd_publish_key(**kwargs):
-    key = kwargs['instance']
-    _etcd_client.write('/deis/builder/users/{}/{}'.format(
-        key.owner.username, fingerprint(key.public)), key.public)
-
-
-def _etcd_purge_key(**kwargs):
-    key = kwargs['instance']
-    try:
-        _etcd_client.delete('/deis/builder/users/{}/{}'.format(
-            key.owner.username, fingerprint(key.public)))
-    except KeyError:
-        pass
-
-
-def _etcd_purge_user(**kwargs):
-    username = kwargs['instance'].username
-    try:
-        _etcd_client.delete(
-            '/deis/builder/users/{}'.format(username), dir=True, recursive=True)
-    except KeyError:
-        # If _etcd_publish_key() wasn't called, there is no user dir to delete.
-        pass
-
-
-def _etcd_publish_app(**kwargs):
-    appname = kwargs['instance']
-    try:
-        _etcd_client.write('/deis/services/{}'.format(appname), None, dir=True)
-    except KeyError:
-        # Ignore error when the directory already exists.
-        pass
-
-
-def _etcd_purge_app(**kwargs):
-    appname = kwargs['instance']
-    try:
-        _etcd_client.delete('/deis/services/{}'.format(appname), dir=True, recursive=True)
-    except KeyError:
-        pass
-
-
 # Log significant app-related events
 post_save.connect(_log_build_created, sender=Build, dispatch_uid='api.models.log')
 post_save.connect(_log_release_created, sender=Release, dispatch_uid='api.models.log')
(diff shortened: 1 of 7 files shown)


helgi:
`SECRET_KEY` needs to be set (https://docs.djangoproject.com/en/1.9/ref/settings/#secret-key) but we shouldn't be publishing it anywhere really

helgi:
Will read the secrets from https://github.com/deis/charts/pull/67

helgi:
Was told the protocol is hard-coded to `http` right now so no need to try to solve for that until we tackle it platform wide

rootfs/api/models/__init__.py
```diff
@@ -16,8 +15,11 @@
 from django.dispatch import receiver
 from django.core.exceptions import ValidationError
 from rest_framework.authtoken.models import Token
+<<<<<<< 34ff500564254a1f010f39e60bd7e4dd29967143
```
slack: O_o

helgi: Nothing to see here

rootfs/templates/confd_settings.py
```diff
@@ -16,6 +16,8 @@
 
 {{ if exists "/deis/controller/registrationMode" }}
 REGISTRATION_MODE = '{{ getv "/deis/controller/registrationMode" }}'
+{{ else }}
+REGISTRATION_MODE = 'enabled'
```
slack: We really need to move this into an API within workflow.

helgi: Agreed - `admin:foo` ahoy

rootfs/deis/settings.py
```diff
@@ -296,13 +292,6 @@
 
 APP_URL_REGEX = '[a-z0-9-]+'
 
-# Unit Hostname handling.
-# Supports:
-#  default      - Docker generated hostname
-#  application  - Hostname based on application unit name (i.e. my-application.v2.web.1)
-#  server       - Hostname based on CoreOS server hostname
-UNIT_HOSTNAME = 'default'
```
bacongobbler: nice catch on the dead code. This was a fleet-specific feature.

arschles:
Will this PR allow us to remove etcd entirely?

helgi:
Yeah

rootfs/templates/confd_settings.py
bacongobbler: +1. Either that or make this configurable in deis/settings.py via `REGISTRATION_MODE = os.getenv('REGISTRATION_MODE', 'enabled')`

arschles:
that's pretty :bomb:.com

rootfs/templates/confd_settings.py
helgi: I switched `confd` to `env` mode so what you proposed and what's there are pretty much the same. I didn't want to try to replace `confd` yet for the `gunicorn` config

rootfs/bin/boot
```diff
@@ -13,62 +13,11 @@ echo system information:
...
 python --version
 
-# configure etcd
-export ETCD_PORT=${DEIS_ETCD_1_SERVICE_PORT_CLIENT:-4001}
-export ETCD_HOST=${DEIS_ETCD_1_SERVICE_HOST:-$HOST}
-export ETCD="$ETCD_HOST:$ETCD_PORT"
-export ETCD_PATH=${ETCD_PATH:-/deis/controller}
-export ETCD_TTL=${ETCD_TTL:-20}
+# spawn confd in the background to update services based on changes
+confd -backend env --confdir /app --log-level error --interval 5 &
```
arschles: what's confd watching on now?

bacongobbler:
+1 on everything except for the reliance on confd to template out envvars. We should just remove confd entirely and rely on `os.getenv` in settings.py much like how we do with the database credentials/config.

rootfs/bin/boot
helgi: the `env` backend

rootfs/bin/boot
```diff
@@ -78,18 +27,18 @@ else
...
 echo "Database Migrations:"
-sudo -E -u deis ./manage.py migrate --noinput
+sudo -E -u deis ./app/manage.py migrate --noinput
 
 # spawn a gunicorn server in the background
-sudo -E -u deis gunicorn -c deis/gconf.py deis.wsgi &
+sudo -E -u deis gunicorn -c /app/deis/gconf.py deis.wsgi &
 
-./manage.py load_db_state_to_etcd
+./app/manage.py load_db_state_to_k8s
```
slack: Should all of these refs be absolute paths? Looks like we're using both absolute and relative. Seems brittle.

helgi:
That's not as friendly when it comes to `gunicorn` - You'd have to cycle the full RC to set new worker count oppose to just dropping on the pod and setting a new env var. While that isn't exactly the cleanest you can edit the RC and do that without any real downtime, in theory, if confd works

rootfs/bin/boot
bacongobbler: envvars. `--backend env` switches it from reading etcd to reading from local environment variables: https://github.com/kelseyhightower/confd/blob/master/docs/quick-start-guide.md#environment-variables

arschles: Ah, I echo what @bacongobbler said right below then. seems like it'd be cleaner (even if more work) to rely on `os.getenv`. As an added bonus, the image would also get smaller.

rootfs/bin/boot
helgi: Before it was doing a `cd app` and then running that. I mostly did it since the `cd` was one more line in the boot script :D

bacongobbler: agreed, especially since we set the WORKDIR to /app in the Dockerfile.

helgi: Just note that `/app/manage.py` is an absolute path. The whole app lives in `/app`. At this point it's just nitpicking

slack: Yeah, but `./app/manage.py` isn't absolute. It is relative to your current working directory. If that changes due to `WORKDIR` or some shenanigans upstream you break.

arschles:
@helgi got it. sounds like it'd be pretty hard to take out confd then. could take it out after we rewrite the controller in Go...

*Aaron runs away*

rootfs/bin/boot
helgi: Yeah, true. It's only like that to run the `pythong`. I'll just change it up a bit

bacongobbler:
In theory, yes. In practice, is there actually a way we can inject environment variables into the running processes through kubectl? If not then I'd still be in favour of removing confd since we wouldn't be able to do live reloads, simplifies the codebase and reduces the image size by a few megs.

helgi:
I'll leave that for another PR - Right now confd also writes out a specific file with "prod" configuration, which doesn't happen on testing so more work would be required than to simply ripping out confd and making things use env vars via pythong

aledbf:
Code LGTM.
Tested with `example-go` and node app.

PR #246, opened 2016-01-06
ref(settings): remove SECRET_KEY from etcd as it does not appear to be used anywhere

helgi:
relates to #9


helgi:
Apparently I didn't do proper googling yesterday... https://docs.djangoproject.com/en/1.9/ref/settings/#secret-key

helgi:
But that should not be part of `etcd` tho ... just needs to be around somehow :p

Issue #13, opened 2015-11-16
Set env vars in rc or pod manifest


mboersma:
Instead of doing a programmatic `docker build` to add on last-mile environment variables, the k8s.py scheduler code could instead add them to the configuration / manifest it creates. `docker.sock` would still be necessary to pull images and push them to deis-registry, but this could be simpler and more efficient for the last-mile layer.

bacongobbler:
We considered this but thought of it as a security concern at one point, but I'm not sure why. It doesn't matter where we store that metadata internally in the cluster, just as long as it *stays* internal. I'm +1 on this idea, since it will dramatically simplify the "magic" behind the scenes.

helgi:
Would be fine with just dumping it all into `env` for now and then later we can look into combining `secrets` into it somehow? There is a GH issue on k8s to expose `secrets` as `env` but it's still in the works. We could do a little scripty injection to achieve that if we wanted to support that adhoc.

Thinking `deis config:set hello=secret --secret` kind of dealio and it would know what to do it it correctly

Would spin that off into a separate ticket

bacongobbler:
yes, I think that's fine for now.

helgi:
https://github.com/kubernetes/kubernetes/issues/4710 is the ticket I was referring to upstream

helgi:
Looks like we are already seeding ENV Vars into k8s RC templates (the pod portion of them). Happened as part of another ticket. That work also specifically added `DEIS_APP` and `DEIS_RELEASE`.

I did a PR that drops those parts (needs proper testing)

helgi:
Created a new ticket, #347, to deal with the secret portion

Issue #141, opened 2015-12-18
Better logic for differentiating apps


smothiki:
The current logic assumes all procfile type apps are web and dockerfile type apps are web.
We have a way to identify whether an app has dockerfile or not. Need to pass this as an extra argument while creating an RC. So that users can create Procfile or Dockerfile type apps of any type.

arschles:
@smothiki I suggested the below in #258. I think it removes all ambiguity between dockerfile and buildpack apps.

- for buildpack builds, builder sends `quay.io/deis/slugrunner` as the image, and the URL of the slug as another param
- for dockerfile builds (not yet implemented), builder sends the URL of the image (which would be in the internal registry) and no URL for the slug (since it's N/A). It would no longer send "dockerfile": true

smothiki:
I was thinking the same. But Embedding images and envs into a singel template is more maintenance than choosing two different templates. I'm thinking of having single RC template and two container templates which make life easier

helgi:
I'm waiting for a k8s PR to land (and get released) that separate the Pod Template from RC. That'd help here in the future :)

smothiki:
there is not requirement for a separate POD template as only containers are being changed

arschles:
that's fine. my suggestion was only about the API (which, to my knowledge, is only used by builder). to illustrate, here's a prototype of the request that workflow could take for a buildpack app:

```json
{
  "sha": "abc....",
  "receive_user": "arschles",
  "receive_repo": "myapp",
  "app": {
    "type": "buildpack",
    "procfile": {
      "web": "example_go"
    }
  }
}
```

and here's a prototype for a dockerfile app:

```json
{
  "sha": "abc....",
  "receive_user": "arschles",
  "receive_repo": "myapp",
  "app": {
    "type": "dockerfile",
    "image": "quay.io/somerepo/someimage:sometag"
  }
}
```

smothiki:
As of now builder sends whether an apptype is dockerfile or not. If we can define App type we have to change the  workflow release model which is a good start and builder can send directly the container JSON object which can be embedded directly into the RC template .

smothiki:
The above mentioned model will simplify k8s scheduler code alot in choosing type of apps. Having builder send the templated container object to workflow makes k8s scheduler code agnostic of app type.

arschles:
what do you mean by "templated container object"

smothiki:
Apologies the above described theory is not possible as builder has no idea about releases from workflow. Please ignore

[...]

PR #248, opened 2016-01-07
feat(registry): stop adding last mile ENV Vars into slugs as k8s handles that

helgi:
I am assuming we need to keep the `FROM` portion of the `build` section? We can not just pull the source and push to our own registry with a new name?

ref #13

### DIFF
--- a/rootfs/registry/dockerclient.py
+++ b/rootfs/registry/dockerclient.py
@@ -2,7 +2,6 @@
 """Support the Deis workflow by manipulating and publishing Docker images."""
 
 from __future__ import unicode_literals
-import io
 import logging
 
 from django.conf import settings
@@ -22,7 +21,7 @@ class DockerClient(object):
         self.client = docker.Client(version='auto')
         self.registry = settings.REGISTRY_HOST + ':' + str(settings.REGISTRY_PORT)
 
-    def publish_release(self, source, config, target, deis_registry):
+    def publish_release(self, source, target, deis_registry):
         """Update a source Docker image with environment config and publish it to deis-registry."""
         # get the source repository name and tag
         src_name, src_tag = docker.utils.parse_repository_tag(source)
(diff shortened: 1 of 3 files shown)


bacongobbler:
Perhaps the faster way to do a build would be to just pull, tag then push since that's really all this is doing now.

rootfs/registry/dockerclient.py
```diff
@@ -46,7 +46,6 @@ def publish_release(self, source, config, target, deis_registry):
         self.tag(image, src_name, tag=src_tag)
 
         # build a Docker image that adds a "last-mile" layer of environment
-        config.update({'DEIS_APP': name, 'DEIS_RELEASE': tag})
```
bacongobbler: I think we'll need to inject these into the environment on `create` [here](https://github.com/deis/workflow/blob/15df775432a4d7f87c8d69805e7bab6ce857b838/rootfs/api/models.py#L221) for DEIS_APP and [here](https://github.com/deis/workflow/blob/15df775432a4d7f87c8d69805e7bab6ce857b838/rootfs/api/models.py#L485) for DEIS_RELEASE.

smothiki:
we hit that function if we hit deis pull use case 
https://github.com/deis/workflow/blob/master/rootfs/api/models.py#L880-L881

smothiki:
deis pull apps doesn't fall into buildpack or dockerfile, so we will set the environment variables. Might be pulling up dockerfile template and end up having two environment variables. https://github.com/deis/workflow/blob/master/rootfs/scheduler/__init__.py#L606-L608. Which I'll fix while dealing with issue.
https://github.com/deis/workflow/issues/141

bacongobbler:
@smothiki if that's the case then that's a bug. All three deployment types should have DEIS_APP and DEIS_RELEASE present in their runtime environment. The proposed changes that I suggested in https://github.com/deis/workflow/pull/248#discussion_r49103251 should revert and fix that back to the way things were.

smothiki:
DEIS_RELEASE and DEIS_APP are being set in the replication controller for the rest of the app types.

helgi:
I cut out the build step all together and instead we are doing a pull / push to get it into the deis-registry with all the tagging

rootfs/registry/dockerclient.py
```diff
@@ -45,26 +45,9 @@ def publish_release(self, source, config, target, deis_registry):
...
         self.push("{}/{}".format(self.registry, name), tag)
 
-    def build(self, source, config, repo, tag):
-        """Add a "last-mile" layer of environment config to a Docker image for deis-registry."""
-        check_blacklist(repo)
-        env = ' '.join("{}='{}'".format(
-            k, v.encode('unicode-escape').replace("'", "\\'")) for k, v in config.viewitems())
-        dockerfile = "FROM {}\nENV {}".format(source, env)
-        f = io.BytesIO(dockerfile.encode('utf-8'))
-        target_repo = "{}/{}:{}".format(self.registry, repo, tag)
```
bacongobbler: By removing this line, we are no longer tagging an image called `$REGISTRY_URL/appname:v2`, which will break the push operation on line 49. From what I see, line 46 doesn't really do much other than re-tag the incoming image from `quay.io/mycustomimage:latest` to just `mycustomimage:latest`, stripping the external registry URL. In order to fix this, I think we need to change line 46 to something like:

```
self.tag(image, "{}/{}".format(self.registry, name), tag=tag)
```

That way the image should be properly bagged and tagged for shipping to the registry. I'll verify my thoughts manually tomorrow, but if you have more insight on how this works then I'm prepared to be wrong :)

helgi: Doneeeeee and it works on `deis pull`

bacongobbler:
:shipit:

helgi:
I tried to use https://github.com/docker/docker-py/blob/master/docker/api/image.py#L45 to see if that could simplify anything but didn't have any luck (yet). Doesn't seem to do the kind of tagging we need nor does it appear to block but I may have just used it wrong :D

PR #258, opened 2016-01-09
fix(scheduler): support both long and short sha being passed in for builds

helgi:


rootfs/scheduler/__init__.py
```diff
@@ -601,11 +601,14 @@ def _create_rc(self, name, image, command, **kwargs):
...
         imgurl = self.registry + "/" + image
         TEMPLATE = RCD_TEMPLATE
-        shalen = len(image[image.index(":")+5:])
-        git = image[image.index(":")+1:image.index(":")+4]
-        if git == "git" and shalen == 8 and app_type == 'web':
-            imgurl = "http://"+settings.S3EP+"/git/home/"+image+"/push/slug.tgz"
+
+        # Check if it is a slug builder image.
+        # Example format: golden-earrings:git-5450cbcdaaf9afe6fadd219c94ac9c449bd62413s
+        vcs, sha = image[image.index(':')+1:].split('-')
```
arschles: @helgi what do you think about the following changes to the API:

- for buildpack builds, builder sends `quay.io/deis/slugrunner` (or other appropriate image as appropriate) as the image, and the URL of the slug as another param?
- for dockerfile builds (not yet implemented), builder sends the URL of the image (which would be in the internal registry) and no URL for the slug (since it's N/A). It would no longer send `"dockerfile": true`

helgi: yeah, we can do it as part of another PR. Create a ticket. That should take care of a few hardcoded points we have in the system as it stands

smothiki:
I'm working on this issue https://github.com/deis/workflow/issues/141. which might completely avoid that condition

rootfs/scheduler/__init__.py
arschles: sounds good @helgi

arschles: @helgi @smothiki I copied these suggestions to #141

smothiki:
I have no worries about merging this PR but soon this code will get obsolete

PR #260, opened 2016-01-11
fix(scheduler): python 3 b64 needs bytes but k8s api needs string passed

helgi:
base64encode returns a bytes object which is doesn't do so well with json serialization so I decode() it to string first

Causes a failure in all new apps

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -720,19 +720,18 @@ class KubeHTTPClient(AbstractSchedulerClient):
     # SECRETS #
 
     def _create_secret(self, namespace):
-        secretId, secretKey = '', ''
-        with open("/var/run/secrets/deis/minio/user/access-key-id") as the_file:
+        with open("/var/run/secrets/deis/minio/user/access-key-id", "rb") as the_file:
             secretId = the_file.read()
-        with open("/var/run/secrets/deis/minio/user/access-secret-key") as the_file:
+        with open("/var/run/secrets/deis/minio/user/access-secret-key", "rb") as the_file:
             secretKey = the_file.read()
-        Key, Id = base64.b64encode(secretKey), base64.b64encode(secretId)
-        l = {
+
+        template = json.loads(string.Template(SECRET_TEMPLATE).substitute({
             "version": self.apiversion,
             "id": namespace,
-            "secretId": Id,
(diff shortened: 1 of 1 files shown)


rootfs/scheduler/__init__.py
```diff
@@ -721,19 +721,18 @@ def _healthcheck(self, controller, path='/', port=8080, delay=30, timeout=1):
...
-        l = {
+
+        template = json.loads(string.Template(SECRET_TEMPLATE).substitute({
             "version": self.apiversion,
             "id": namespace,
-            "secretId": Id,
-            "secretKey": Key,
-            }
-        template = json.loads(string.Template(SECRET_TEMPLATE).substitute(l))
+            "secretId": base64.b64encode(secretId).decode(),
```
arschles: @helgi I get that `b64encode` returns bytes, but why decode it immediately after encoding it? otherwise, wouldn't `str(secretId)` be more efficient?

helgi: Because that's what the documentation suggests. Originally did `str` but was like "docs must be gooood". I'll change it back to `str` since it is easier to grok it

arschles: I'm no python expert, but at face value, it seems inefficient to round-trip some bytes through a base64 coder just to convert to a string

helgi: That's kinda not the point tho. the `base64` interface only offers `bytes` as input / output and the `json` interface can't deal with being partially `bytes` :)

arschles: so, doesn't `str()` convert `bytes` to strings in a more clear way (and more efficiently I'm guessing)? that's my only point here

helgi: Python 3 split `str` into `bytes` and `str` vs the old `bytestring` and `unicode` concepts in v2. http://www.diveintopython3.net/strings.html might explain some of the differences. Especially the Further Reading section on the various (albeit silly) differences :/

helgi: Clearer yeah (which is more important), not sure about more efficient but it is also a code path that is rarely hit.

arschles: k, got it. thanks for the reading, too, very helpful

PR #263, opened 2016-01-11
feat(logger): Support logger v2

jchauncey:
This PR submits the necessary code changes to support the new v2 logger implementation.

### DIFF
--- a/client/controller/models/apps/apps.go
+++ b/client/controller/models/apps/apps.go
@@ -98,10 +98,21 @@ func Logs(c *client.Client, appID string, lines int) (string, error) {
 	body, err := c.BasicRequest("GET", u, nil)
 
 	if err != nil {
-		return "", err
+		return fmt.Sprintf("Error:%v", err), err
 	}
 
-	return strings.Trim(body, `"`), nil
+	if len(body) < 1 {
+		return fmt.Sprintf(
+			`There are currently no log messages. Please check the following things:
+1) Logger and fluentd pods are running.
+2) The application is writing logs to the logger component.
+You can verify that logs are appearing in the logger component by issuing the following command:
+curl http://<log service ip>:8088/%s on a kubernetes host.
+To get the service ip you can do the following: kubectl get svc deis-logger --namespace=deis`, appID), nil
+	}
+
+	// We need to trim a few characters off the front and end of the string
+	return body[3 : len(body)-2], nil
 }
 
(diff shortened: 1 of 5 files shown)


client/controller/models/apps/apps.go
```diff
@@ -92,7 +92,7 @@ func Logs(c *client.Client, appID string, lines int) (string, error) {
 	u := fmt.Sprintf("/v2/apps/%s/logs", appID)
 
 	if lines > 0 {
-		u += "?log_lines=" + strconv.Itoa(lines)
+		u += "/?log_lines=" + strconv.Itoa(lines)
```
krancour: Do logger URLs absolutely need a slash at the end now?  It seems a generally accepted practice that the final trailing slash in a URL is irrelevant.

jchauncey: yeah for some reason I had to add the slash in the client to get it to talk to the logger component. I'll look into seeing if I can fix this.

krancour: Sorry... I read this wrong.  This is in client code.  Maybe this has to do with recent Django updates?

rootfs/api/models.py
```diff
@@ -475,7 +475,7 @@ def _default_scale(self, user, release):
     def logs(self, log_lines=str(settings.LOG_LINES)):
         """Return aggregated log data for this application."""
         try:
-            url = "http://{}:{}/{}?log_lines={}".format(settings.LOGGER_HOST, settings.LOGGER_PORT,
+            url = "http://{}:{}/{}/?log_lines={}".format(settings.LOGGER_HOST, settings.LOGGER_PORT,
```
krancour: This is the spot I thought I was commenting on before.  I checked and the extra slash isn't needed here.  But it also doesn't hurt, so I won't make a fuss. :)

jchauncey: Yeah I've already made the slash optional but haven't pushed that yet

jchauncey:
Ok so this PR is ready to start having serious reviews. 

* You will need to build a new deis client: `cd client && make build`
* You will need to use the [deis-logger chart](https://github.com/deis/charts/pull/27)
* You will also need to build a custom workflow image and deploy it: `make docker-build docker-push kube-update`

Go through the steps to create and deploy an application. Then you should be able to use `deis logs` like normal.

If you have any questions let me know.

PR #271, opened 2016-01-12
ref(secrets): read generated secrets from k8s for the setings file

### LINKED ISSUE #9: create k8s secrets for DEIS_SECRET_KEY and DEIS_BUILDER_KEY
These two keys are [auto-generated](https://github.com/deis/workflow/blob/8ed73db9cfaea9c48c985e298c0eb26c733b90bc/rootfs/bin/boot#L52-L53) by the controller on boot, then templated out in the controller and the builder using confd. We should plug into the kubernetes secrets API instead, if possible.

helgi:
Not the biggest fan of reading it on every API request but we do have a very low yield API right now and we can somewhat rely on filesystem caches

Maybe later (stable timeframe or after stable) I can output it into the file statically or use confd + env vars like I did in #245 for other things

Fixes #9

### DIFF
--- a/rootfs/templates/confd_settings.py
+++ b/rootfs/templates/confd_settings.py
@@ -1,8 +1,11 @@
 import os
 
 # security keys and auth tokens
-SECRET_KEY = '{{ getv "/deis/controller/secretKey" }}'
-BUILDER_KEY = '{{ getv "/deis/controller/builderKey" }}'
+with open('/var/run/secrets/api/builder/auth/builder-key') as f:
+    BUILDER_KEY = f.read()
+
+with open('/var/run/secrets/api/django/secret-key') as f:
+    SECRET_KEY = f.read()
 
 # scheduler settings
 SCHEDULER_MODULE = 'scheduler'
--- a/rootfs/bin/boot
+++ b/rootfs/bin/boot
@@ -50,8 +50,6 @@ function etcd_safe_mkdir {
 }
 
 etcd_set_default protocol "${DEIS_PROTOCOL:-http}"
-etcd_set_default secretKey "${DEIS_SECRET_KEY:-$(openssl rand -base64 64 | tr -d '\n')}"
-etcd_set_default builderKey "${DEIS_BUILDER_KEY:-$(openssl rand -base64 64 | tr -d '\n')}"
 etcd_set_default registrationMode "enabled"
 etcd_set_default webEnabled 0
 
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -664,8 +664,10 @@ class KubeHTTPClient(AbstractSchedulerClient):
                 continue
             create = True
             rc = self._get_rc(name, app_name)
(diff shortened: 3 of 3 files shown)


rootfs/templates/confd_settings.py
```diff
@@ -1,8 +1,11 @@
 import os
 
 # security keys and auth tokens
-SECRET_KEY = '{{ getv "/deis/controller/secretKey" }}'
-BUILDER_KEY = '{{ getv "/deis/controller/builderKey" }}'
+with open('/var/run/secrets/api/builder/auth/builder-key') as f:
+    BUILDER_KEY = f.read()
+
+with open('/var/run/secrets/api/django/secret-key') as f:
+    SECRET_KEY = f.read()
```
krancour: Would this be better off in `settings.py` instead of in `confd_settings.py`?

helgi: Was mostly keeping consistency - `settings.py` is static, `confd_settings.py` is not. Could totally move to `settings.py` tho since it isn't happening as part of `confd`. I'm indifferent for now

krancour: If I understand correctly, the two get merged settings files get merged to create a single settings object?  I'm only thinking it seems odd to have code in `confd_settings.py` for managing values that are entirely unaffected by confd.  That said, I don't feel strongly enough about it to block this.

helgi:
Note we are losing the ability (for now) to specify custom keys. We can fix that in helm at some point, auto generate if people do not provide `values.toml` or whatever is the right now :)

krancour:
@helgi just realized... should there have been manifest changes in this repo that go along with this?

arschles:
ooooh, @krancour good call - forgot about this. +1

rootfs/templates/confd_settings.py
helgi: Actually that'd make the tests fail since they do not have access to those files

helgi:
Oh yeah... forgot we do this silly duplicate all the charts thing. I can whip up a PR, but it will depend on helm

arschles:
depends on helm 0.3.1, right?

helgi:
`0.3.0` is fine, the `.1` was to fix some versioning I think. So yeah the `0.3` series

arschles:
@helgi cool. IMO not a problem to have it depend on helm

krancour:
@helgi maybe we can avoid it depending on helm?  I think the workflow that seems to be emerging is that people are helm installing deis-dev and then they're using make targets in whatever project they hack on to "side load" the hacked upon bits into the cluster.

With that in mind, you could probably pretend like the secrets in question must already exist.  (Because they probably do-- and if they don't, then the builder probably doesn't exist either and thus the missing secret would be irrelevant.)

Just a thought about how to avoid helm stuff creeping into the repo.

helgi:
We can't avoid it but we surely assume it exists already for the particular repo. 

For everyone else: We are using go templates to auto generate the secrets, which are a helm functionality.

PR #275, opened 2016-01-13
feat(workflow): add build_type in extra args to differentiate build t…

smothiki:
fixes https://github.com/deis/workflow/issues/141
Will open a different PR for abstracting templates into hierarchies.

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -604,15 +604,11 @@ class KubeHTTPClient(AbstractSchedulerClient):
 
         # Check if it is a slug builder image.
         # Example format: golden-earrings:git-5450cbcdaaf9afe6fadd219c94ac9c449bd62413s
-        image_name = image[image.index(':')+1:]
-        if '-' in image_name:
-            vcs, sha = image_name.split('-')
-            if vcs == 'git' and len(sha) in [8, 40] and app_type == 'web':
-                imgurl = 'http://{}/git/home/{}/push/slug.tgz'.format(
-                    settings.S3EP,
-                    image
-                )
-                TEMPLATE = RCB_TEMPLATE
+        if kwargs.get('build_type', {}) == "buildpack":
+            imgurl = 'http://{}/git/home/{}/push/slug.tgz'.format(
+                settings.S3EP,
+                image)
+            TEMPLATE = RCB_TEMPLATE
 
         l = {
             "name": name,
--- a/rootfs/api/models.py
+++ b/rootfs/api/models.py
@@ -26,7 +26,7 @@ import requests
 from rest_framework.authtoken.models import Token
 
 from registry import publish_release
-from api.utils import dict_diff, dict_merge, fingerprint, generate_app_name
+from api.utils import dict_diff, dict_merge, fingerprint, generate_app_name, app_build_type
 
 
 User = settings.AUTH_USER_MODEL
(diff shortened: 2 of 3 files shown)


smothiki:
```
DEBUG create hushed-yeomanry_v2.cmd, img hushed-yeomanry:v2, params {'app_type': 'cmd', 'memory': {}, 'aname': 'hushed-yeomanry', 'num': 1, 'build_type': 'deispull', 'cpu': {}, 'healthcheck': {'port': 8080, 'timeout': 1, 'path': '/', 'delay': 10}, 'envs': {}, 'tags': {}, 'version': 'v2'}, cmd ""
```
```
DEBUG create rubber-zucchini_v2.web, img rubber-zucchini:git-e1590887, params {'app_type': 'web', 'memory': {}, 'aname': 'rubber-zucchini', 'num': 1, 'build_type': 'buildpack', 'cpu': {}, 'healthcheck': {'port': 8080, 'timeout': 1, 'path': '/', 'delay': 10}, 'envs': {}, 'tags': {}, 'version': 'v2'}, cmd "start web"
```
```
$ kubectl get pods --namespace=rubber-zucchini
NAME                           READY     STATUS    RESTARTS   AGE
rubber-zucchini-v2-web-llo6b   1/1       Running   0          10m
i$ kubectl get pods --namespace=hushed-yeomanry
NAME                           READY     STATUS    RESTARTS   AGE
hushed-yeomanry-v2-cmd-z3ud4   1/1       Running   0          8m
```

arschles:
@smothiki this LGTM, but can you wait for deis/builder#72 before merging this?

smothiki:
Sure, but this PR has nothing to do with builder controller interactions. Also I'm testing builder rewrite now. Let me know if there are any issues.

arschles:
isn't `_deploy_app` invoked in response to a `/v2/hooks/build` request?

smothiki:
yes, but unless you change the build model, builder has no affect on this PR.

arschles:
ok, but does this PR have any effect on builder? ;)

smothiki:
Nope

arschles:
ok, if you've tested a `git push` then I'm convinced :)

smothiki:
Yeah . you can check the output I have pasted deployed two apps and both the PODS are running

arschles:
got it, thanks for clarifying!

rootfs/scheduler/__init__.py
```diff
@@ -604,15 +604,11 @@ def _create_rc(self, name, image, command, **kwargs):  # noqa
...
-        image_name = image[image.index(':')+1:]
-        if '-' in image_name:
-            vcs, sha = image_name.split('-')
-            if vcs == 'git' and len(sha) in [8, 40] and app_type == 'web':
-                imgurl = 'http://{}/git/home/{}/push/slug.tgz'.format(
-                    settings.S3EP,
-                    image
-                )
-                TEMPLATE = RCB_TEMPLATE
+        if kwargs.get('build_type', {}) == "buildpack":
```
helgi: Only worry about the defaulting to `{}` here since we are using strings in there. Maybe no defaulting and rather have things blow up hard if somehow `app_build_type` doesn't return anything, not that I see how it would :)

smothiki: Yeah it return "dockerfile" by default

PR #284, opened 2016-01-15
chore(deploy.sh): Stop producing v2-beta images for each PR merge

jchauncey:
Instead we should produce a canary image. This allows us to cut stable release images while providing an image that tracks HEAD. This image should also be in the deis org.

### DIFF
--- a/manifests/deis-workflow-rc.yml
+++ b/manifests/deis-workflow-rc.yml
@@ -65,4 +65,3 @@ spec:
         - name: builder-key-auth
           secret:
             secretName: builder-key-auth
-
--- a/_scripts/deploy.sh
+++ b/_scripts/deploy.sh
@@ -5,7 +5,7 @@
 
 cd "$(dirname "$0")" || exit 1
 
-export IMAGE_PREFIX=deisci VERSION=v2-beta
+export VERSION=canary
 docker login -e="$DOCKER_EMAIL" -u="$DOCKER_USERNAME" -p="$DOCKER_PASSWORD"
 DEIS_REGISTRY='' make -C .. docker-build docker-push
 docker login -e="$QUAY_EMAIL" -u="$QUAY_USERNAME" -p="$QUAY_PASSWORD" quay.io


manifests/deis-workflow-rc.yml
```diff
@@ -16,7 +16,7 @@ spec:
     spec:
       containers:
         - name: deis-workflow
-          image: quay.io/deisci/workflow:v2-beta
+          image: quay.io/deis/workflow:v2-beta
```
sgoings: Pretty sure you want to track canary here, right?

mboersma: And we're using the `deisci` org for *everything* until Deis v2 has an actual release. There shouldn't ever be test images deployed to the `deis` org on Docker Hub or quay.io.

jchauncey: @sgoings no. we still want to track the stable builds.

@mboersma i disagree completely. It makes it extremely confusing that we are producing images in two different orgs. The images we have cut for alpha are in `deis` while images being produced right now are in `deisci`. It makes it difficult to go to the quay ui and see what has been updated and what hasn't. The upkeep will become a nightmare as we produce more versions of the beta and alpha releases. While I agree that jenkins testing images should be in the deisci org. Anything that can be used by us or a customer should be in the deis org.

mboersma:
I think we agree, in that "canary" images are testing images and should go in `deisci/`. They are not supported nor intended for customers to use. Only released, supported tags go into `deis/`.

More specifically, there is no `quay.io/deis/workflow:v2-beta` image to download, so that change will not work.

jchauncey:
No canary is the image that represents the current state of the repo. It is a point in time image that allows us to easily check the state of our system before we cut a new release (v2-beta, alpha, or whatever). You would more than likely load up a deis install using all canary images and verify everything works.

So while we could call it a testing image. It is definitely not a CI image. If thats the case then why not tag each CI run and just not worry about canary.

krancour:
I agree with @jchauncey's POV on this.  The CI process may _produce_ the canary images, and we may, in fact use them for testing, but the canary image of each component is definitely it's own _thing_ (current state) and, in theory (though we may not be there yet), canary image of each component is a release candidate and is merely a tag away from becoming an official semver'ed release.

PR #285, opened 2016-01-15
ref(tests): since Django 1.7 tests are auto detected based on test*.py

helgi:
https://docs.djangoproject.com/en/1.7/topics/testing/overview/

When you run your tests, the default behavior of the test utility is to find all the test cases (that is, subclasses of unittest.TestCase) in any file whose name begins with test, automatically build a test suite out of those test cases, and run that suite.


mboersma:
Maintaining these imports was a pain. It's great that this isn't needed in newer versions of Django. :+1:

PR #292, opened 2016-01-20
ref(router): use the built in router of Django

helgi:
DefaultRouter does all the actions we need and the `@action` and `@link` is not used anywhere in our code base.

`@list_route` and `@detail_route` exist if needed http://www.django-rest-framework.org/api-guide/routers/#defaultrouter


arschles:
:+1: on all the deletions!

Issue #312, opened 2016-01-27
dockerfile apps are not starting


smothiki:
I have deployed a dockerfile based app and the POD kept failing because of readiness probe

```
  Tue, 26 Jan 2016 18:32:06 -0700	Tue, 26 Jan 2016 18:32:06 -0700	1	{kubelet ip-172-20-0-23.us-west-1.compute.internal}	spec.containers{zigzag-traverse-cmd}	Unhealthy	Liveness probe failed: Get http://10.244.1.7:8080/: dial tcp 10.244.1.7:8080: connection refused
  Tue, 26 Jan 2016 18:31:56 -0700	Tue, 26 Jan 2016 18:32:36 -0700	2	{kubelet ip-172-20-0-23.us-west-1.compute.internal}	spec.containers{zigzag-traverse-cmd}	Unhealthy	Readiness probe failed: 
  Tue, 26 Jan 2016 18:32:36 -0700	Tue, 26 Jan 2016 18:32:36 -0700	1	{kubelet ip-172-20-0-23.us-west-1.compute.internal}	spec.containers{zigzag-traverse-cmd}	Killing		Killing with docker id 4dd7f9f0efeb
  Tue, 26 Jan 2016 18:32:36 -0700	Tue, 26 Jan 2016 18:32:36 -0700	1	{kubelet ip-172-20-0-23.us-west-1.compute.internal}	spec.containers{zigzag-traverse-cmd}	Pulled		Container image "10.0.165.104:5000/zigzag-traverse:git-be3be457d1466c5cf6000496179875ab6dd22bdd" already present on machine
  Tue, 26 Jan 2016 18:32:36 -0700	Tue, 26 Jan 2016 18:32:36 -0700	1	{kubelet ip-172-20-0-23.us-west-1.compute.internal}	spec.containers{zigzag-traverse-cmd}	Created		Created with docker id c3a9963281f5
  Tue, 26 Jan 2016 18:32:36 -0700	Tue, 26 Jan 2016 18:32:36 -0700	1	{kubelet ip-172-20-0-23.us-west-1.compute.internal}	spec.containers{zigzag-traverse-cmd}	Started		Started with docker id c3a9963281f5
```

helgi:
Is the app exposed on port `8080`?

smothiki:
Nope. Its exposed on 80.
we have to change the logic for readiness probe https://github.com/deis/workflow/pull/313

helgi:
Seems closed based on that PR

PR #313, opened 2016-01-27
fix(dockerpull): fix logic for image pull for dockerfile apps

smothiki:
to create service we have to know port . 
This fixes logic for pulling deis pull app images and dockerfile type app images for inspecting Port

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -330,8 +330,8 @@ class KubeHTTPClient(AbstractSchedulerClient):
                 data = {'metadata': {'labels': {'routable': 'true'}}}
             self._create_namespace(app_name)
             self._create_secret(app_name)
-            self._create_rc(name, image, command, **kwargs)
             self._create_service(name, app_name, app_type, data, image=image)
+            self._create_rc(name, image, command, **kwargs)
 
         except Exception as e:
             logger.debug(e)
@@ -697,11 +697,13 @@ class KubeHTTPClient(AbstractSchedulerClient):
         if controller['spec']['selector']['type'] not in ['web', 'cmd']:
             return controller
 
+        app_name = controller['spec']['selector']['app']
         # Inspect if a PORT env is already defined, make sure that's the port used
-        for item in controller['spec']['template']['spec']['containers'][0]['env']:
-            if item['name'] == 'PORT':
-                port = int(item['value'])
-
+        try:
+            service = self._get_service(app_name, app_name).json()
+            port = service['spec']['ports'][0]['targetPort']
+        except:
+            pass
         # Only support HTTP checks for now
         # http://kubernetes.io/v1.1/docs/user-guide/pod-states.html#container-probes
         healthcheck = {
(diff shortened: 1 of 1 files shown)


smothiki:
also fixes https://github.com/deis/workflow/issues/312

helgi:
Problem here is that we are falling back on what the user defines as `PORT` and hoping that docker is right. Probably is the best thing but just wanted to point out that now what the user defines is a fallback

smothiki:
in that case we should give user to configure health check ports( I think this feature is already there) and use conditions that use configured port first and if not set then fall back to ports from spec and docker image. Let me know if health checks are configurable using ```deis config``` I'll add logic accordingly.

helgi:
They should be, the `PORT` logic was to get past certain limitations before, less worried about it now.

smothiki:
Also creates service first and then RC. helps in probing service for target port

rootfs/scheduler/__init__.py
```diff
@@ -770,8 +772,9 @@ def _create_service(self, name, app_name, app_type, data={}, **kwargs):
         image = kwargs.get('image')
         try:
             image = self.registry + '/' + image
+            repo = image.split(":")
             # image already includes the tag, so we split it out here
-            docker_cli.pull(image.rsplit(':')[0], image.rsplit(':')[1])
+            docker_cli.pull(repo[0]+":"+repo[1], tag=repo[2], insecure_registry=True)
```
helgi: Why do we need `insecure_registry` here? Because we are generally doing `insecure` in deis right now?

helgi:
Nice, much cleaner overall logic

smothiki:
```
spec:
  containers:
  - env:
    - name: DEIS_APP
      value: loving-jerrycan
    - name: DEIS_RELEASE
      value: v2
    image: 10.0.96.168:5000/loving-jerrycan:git-be3be457
    imagePullPolicy: IfNotPresent
    livenessProbe:
      httpGet:
        path: /
        port: 80
        scheme: HTTP
      initialDelaySeconds: 10
      timeoutSeconds: 1
    name: loving-jerrycan-cmd
    readinessProbe:
      httpGet:
        path: /
        port: 80
        scheme: HTTP
      initialDelaySeconds: 10
      timeoutSeconds: 1
    resources: {}
```

rootfs/scheduler/__init__.py
bacongobbler: The internal registry is insecure and the docker engine demands that you explicitly set this flag when dealing with these, however I'm not sure why we need to set it in the CLI if the engine already has such information.

carmstrong: :+1: I don't like the idea of having this logic live in the component unless it's absolutely necessary

smothiki: I copied this line from publish_release thought a best practice 
https://github.com/deis/workflow/blob/master/rootfs/registry/dockerclient.py#L54

smothiki: @bacongobbler do you want me to remove insecure registry flag?

bacongobbler: yes please. we should remove that flag elsewhere in a followup :)

arschles:
@smothiki what is the manifest that you posted ^^?

Anyway, LGTM

smothiki:
Its the APP manifest but not complete one  Just the health check part

Issue #251, opened 2016-01-07
Expose new healthcheck probe params as soon as they are available in k8s


helgi:
- PeriodSeconds - How often to probe
- SuccessThreshold - Number of successful probes to go from failure to success state
- FailureThreshold - Number of failing probes to go from success to failure state

https://github.com/kubernetes/kubernetes/pull/15967

PR #331, opened 2016-02-02
feat(healthchecks): add DB health check to /healthz

### LINKED ISSUE #305: Update HealthCheck to hit the database
As discussed during HA conversation we should make sure the HealthCheck is deep enough. It should hit the database.

helgi:
On boot see if the DB is connectable *before* hitting the migration step and in `/healthz` connect to the DB and run a simple dumb `SELECT` statement

Note that this will cause k8s to restart the container after a few tries, if the DB is not available then it will go into a crash loop. This is the nature of the beast:
http://kubernetes.io/v1.1/docs/user-guide/production-pods.html#liveness-and-readiness-probes-aka-health-checks and http://kubernetes.io/v1.1/docs/user-guide/walkthrough/k8s201.html#health-checking

> if the Kubelet discovers a failure the container is restarted.

This is is also based on the [Restart Policy](http://kubernetes.io/v1.1/docs/user-guide/pod-states.html#restartpolicy), which we currently define as `Always

To test deploy `deis-workflow` RC and then scale `deis-database` to 0

We will have more control over this when #251 lands

Fixes #305

### DIFF
--- a/rootfs/api/views.py
+++ b/rootfs/api/views.py
@@ -19,14 +19,31 @@ from rest_framework.authtoken.models import Token
 from api import authentication, models, permissions, serializers, viewsets
 
 import requests
+import logging
+
+logger = logging.getLogger(__name__)
 
 
 class HealthCheckView(View):
-    """Simple health check view to determine if the server
-       is responding to HTTP requests.
+    """
+    Simple health check view to determine if the server
+    is responding to HTTP requests and DB connection / query.
     """
 
     def get(self, request):
+        try:
+            import django.db
+            with django.db.connection.cursor() as c:
+                c.execute("SELECT 0")
+        except django.db.Error as e:
+            logger.critical("Database health check failed")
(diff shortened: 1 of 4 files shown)


helgi:
```
Checking if database is alive
There was a problem connecting to the database
server closed the connection unexpectedly
	This probably means the server terminated abnormally
	before or while processing the request.
```

When booting things up with a defunct DB

rootfs/api/views.py
```diff
@@ -19,14 +19,31 @@
...
+
+logger = logging.getLogger(__name__)
 
 
 class HealthCheckView(View):
-    """Simple health check view to determine if the server
-       is responding to HTTP requests.
+    """
+    Simple health check view to determine if the server
+    is responding to HTTP requests.
```
arschles: maybe change these docs to talk about checking if the DB is available

PR #334, opened 2016-02-03
fix(controller): do not require slash at the end of `GET /v2/users`

### LINKED ISSUE #333: port deis/deis#4898 to v2

mboersma:
This ports the fix from deis/deis#4898 to Deis v2 workflow. 

I updated the URL to `/v2/` here relative to the original PR.  Thanks again to @smt116, let me know if you have any concerns or comments.

Closes #333.

### DIFF
--- a/rootfs/api/tests/test_users.py
+++ b/rootfs/api/tests/test_users.py
@@ -11,23 +11,20 @@ class TestUsers(TestCase):
     fixtures = ['tests.json']
 
     def test_super_user_can_list(self):
-        url = '/v2/users'
-
         user = User.objects.get(username='autotest')
         token = Token.objects.get(user=user)
 
-        response = self.client.get(url,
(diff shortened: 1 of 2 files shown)


rootfs/api/tests/test_users.py
```diff
@@ -11,23 +11,21 @@ class TestUsers(TestCase):
     fixtures = ['tests.json']
 
     def test_super_user_can_list(self):
-        url = '/v2/users/'
+        for url in ['/v2/users', '/v2/users/']:
+            user = User.objects.get(username='autotest')
+            token = Token.objects.get(user=user)
```
helgi: You technically don't need to generate those twice or fetch twice

mboersma: Good point--I'll rearrange that loop.

arschles:
@mboersma LGTM. Also, Github is telling me that this needs to be rebased

PR #336, opened 2016-02-03
feat(builder hooks): app and user hooks for builder to get public keys

### LINKED ISSUE #335: Expose user/app checking for Builder in a builder specific hook
Relates to deis/builder#81 and deis/builder#122

TL;DR from deis/builder#122 (has a more detailed discussion)

Expose `GET /v2/hooks/keys/$APP_NAME` and `GET /v2/hooks/keys/$APP_NAME/$USER_NAME` endpoints for Builder Auth Key only where it returns `404` when `APP_NAME` or `USER_NAME` are missing.

Return body will be `{"users": {"aaron": "keysandthings"}}` for the `APP_NAME` specific hook and `{"key": "keysandthings"}` for the per user hook

helgi:
Give builder the ability to query for application or application + user and get the public key associated if both resource exist.
Replaces etcd usage for builder.

Fixes #335

### DIFF
--- a/rootfs/api/tests/test_hooks.py
+++ b/rootfs/api/tests/test_hooks.py
@@ -15,6 +15,24 @@ from rest_framework.authtoken.models import Token
 
 from . import mock_status_ok
 
+RSA_PUBKEY = (
+    "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQCfQkkUUoxpvcNMkvv7jqnfodgs37M2eBO"
+    "APgLK+KNBMaZaaKB4GF1QhTCMfFhoiTW3rqa0J75bHJcdkoobtTHlK8XUrFqsquWyg3XhsT"
+    "Yr/3RQQXvO86e2sF7SVDJqVtpnbQGc5SgNrHCeHJmf5HTbXSIjCO/AJSvIjnituT/SIAMGe"
+    "Bw0Nq/iSltwYAek1hiKO7wSmLcIQ8U4A00KEUtalaumf2aHOcfjgPfzlbZGP0S0cuBwSqLr"
+    "8b5XGPmkASNdUiuJY4MJOce7bFU14B7oMAy2xacODUs1momUeYtGI9T7X2WMowJaO7tP3Gl"
+    "sgBMP81VfYTfYChAyJpKp2yoP autotest@autotesting comment"
+)
+
+RSA_PUBKEY2 = (
+    "ssh-rsa AAAAB3NzaC1yc2EAAAADAQABAAABAQC4xELdubosJ2/bQuiSUyWclVVa71pXpmq"
+    "aXTwfau/XFLgD5yE+TOFbVT22xvEr4AwZqS9w0TBMp4RLfi4pTdjoIK+lau2lDMuEpbF4xg"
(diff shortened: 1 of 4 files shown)


helgi:
@arschles You can still query for apps via etcd using this PR, that way you can do a smaller bump at first

arschles:
Ref https://github.com/deis/builder/issues/81

arschles:
@helgi thanks for this. LGTM

rootfs/api/tests/test_hooks.py
```diff
@@ -27,12 +36,59 @@ def setUp(self):
...
+        public = response.data['public']
+
+        # Make sure 404 is returned for a random app
+        url = '/v2/hooks/keys/doesnotexist'
+        response = self.client.get(url, HTTP_X_DEIS_BUILDER_AUTH=settings.BUILDER_KEY)
+        self.assertEqual(response.status_code, 404)
+
+        # Test app that exists
+        url = '/v2/hooks/keys/{}'.format(app_id)
+        response = self.client.get(url, HTTP_X_DEIS_BUILDER_AUTH=settings.BUILDER_KEY)
```
smothiki: Keys are associated to USER not APP

arschles: I believe that this endpoint returns a list of user/key pairs

helgi: Correct, if you look at the code I am then using the *app* to query for all users permitted on it and then outputting all those users and their keys.

smothiki: this is the file that we might change  https://github.com/deis/builder/blob/master/rootfs/etc/confd/templates/authorized_keys.

This file is concerned about USER and key associated with the USER . Not the App. just query User and get associated keys is easier than querying APP and then getting USER and Keys.

helgi: Given the API request is going to happen on `git push` where the app is the known factor (read the connected builder tickets) then this is appropriate as is.

smothiki: Okay got it.

helgi:
The key section is now a `list` to account for people having many keys

rootfs/api/views.py
```diff
@@ -397,6 +398,53 @@ class BaseHookViewSet(BaseDeisViewSet):
...
+        if not permissions.is_app_user(request, app):
+            raise PermissionDenied()
+
+        data = {request.user.username: []}
+        keys = models.Key.objects \
+                     .filter(owner__username=kwargs['username']) \
+                     .values('public') \
+                     .order_by('created')
+        if not keys:
+            raise Http404("No Keys matche the given query.")
```
arschles: `s/matche/match` - just caught this

aledbf:
@helgi can you add the fingerprint? (it's used to lookup the username)

helgi:
In what part of it are you needing the fingerprint? the GET or the response of the GET

aledbf:
@helgi in the response of the GET

helgi:
Can you make an issue for it? I will change it so that the response adds a `dict` inside the `list` as such `{'key: '2jk234jk234j', 'fingerprint': '12:34:56:78'}` for each `list` item

PR #338, opened 2016-02-03
feat(certs): add support for wildcard certificates and domains with new cert cli commands

### LINKED ISSUE #254: Certificates in k8s secrets
Make certificates discoverable by the router.

Format will be:
name: `<domain>-cert` (248 char limit on the domain)
contents: `cert` (public key and chain) and `key` in base64

There will be no need to mount the secret to the applications for router to work

ref #50

helgi:
Add support for wildcard certificates that also contain SAN, and wildcard domains. The certificate domain model records far more information now.

New CLI commands added: `certs:info`, `certs:attach`, and `certs:detach`

* `certs:list` output has drastically changed to account for all the new pretty information.
* `certs:create` dropped `--san` and `--common-name` as those are automagically discovered by the API. It also now requires a `name` to identify a certificate which is then used in other commands.
* `certs:attach` and `certs:detach` are used to tie a single certificate to one or more domains. No need to duplicate a certificate to achieve that.

Certificates are stores in k8s as secrets with the certificate name, which get attached to the given application namespace they are in (via `attach`) and router can query for a domain to certificate name mapping

TODO:

- [ ] e2e tests
- [x] router changes

Fixes #254


helgi:
To test it go into `client` directory and run `make build`

helgi:
Note on the e2e tests, there are **no** cert e2e tests as it stands, no prior ones

client/cmd/certs.go
```diff
@@ -93,48 +118,117 @@ func CertAdd(cert, key, commonName, sans string) error {
...
-	certFile, err := ioutil.ReadFile(cert)
+// CertRemove deletes a cert from the controller.
+func CertRemove(name string) error {
+	c, err := client.New()
+	if err != nil {
+		return err
+	}
 
+	fmt.Printf("Removing %s... ", name)
+	quit := progress()
```
arschles: does `progress()` run something in a goroutine?

client/glide.lock
```diff
@@ -18,21 +18,21 @@ imports:
 - name: bitbucket.org/ww/goautoneg
   version: 75cd24fc2f2c
 - name: code.google.com/p/go-charset
-  version: ebbeafdc430e
+  version: ""
```
arschles: not sure why these are empty. did you run `glide up` or `glide install`?

helgi: `glide up` iirc - I was as surprised as the next person

client/cmd/certs.go
helgi: Don't know - Not my original code, even tho it looks like it for some reason in the diff. https://github.com/deis/workflow/blame/master/client/cmd/certs.go#L135

arschles:
@helgi think you need to run `bootstrap` before `build` when you build the client [in the `Makefile`](https://github.com/deis/workflow/blob/master/.travis.yml#L38)

helgi:
Let me give that a go. I was just peaking at `.travis.yml` anyway - Hold on

client/cmd/certs.go
arschles: k, I got no problem leaving it as-is then

arschles:
@helgi can you put the following `ignore`s at the top level of the glide.yaml file. That will help clean up some of the empty `version` fields. The others that will remain are from code.google.com repositories, which we can deal with later.

```yaml
ignore:
- appengine
- appengine/datastore
- appengine/memcache
- appengine/user
```

client/cmd/certs.go
mboersma: [`progress`](https://github.com/deis/workflow/blob/master/client/cmd/utils.go#L15) is a goroutine I wrote to do the text animation when the CLI is waiting.

arschles: :100: @mboersma !

krancour:
The `-a` / `--app` flags are not working with `cert` subcommands.

helgi:
That has never been supported. Certs are independant resources.

krancour:
Ahhhh.  Right.

rootfs/api/views.py
```diff
@@ -346,14 +346,38 @@ def get_object(self, **kwargs):
...
+        try:
+            # TODO how to validate domain is set in the body?
+            kwargs['domain'] = request.data['domain']
+            self.get_object().attach(*args, **kwargs)
+        except Http404:
+            raise
+        except AlreadyExists as e:
+            return Response({'detail': str(e)}, status=status.HTTP_409_CONFLICT)
+        except Exception as e:
+            return Response({'detail': str(e)}, status=status.HTTP_503_SERVICE_UNAVAILABLE)
```
mboersma: Is 503 "Service Unavailable" really the catchall error that should be returned here and in `detach`?

helgi: It was in line with the rest of the API error "handling" so I was mostly just keeping with convention. It can be a different return value if we figure out what's more appropriate for unrecoverable issue

mboersma: > mostly just keeping with convention

I see that now on re-reading views.py. Works for me. :+1:

helgi: to be honest we should revisit that for the whole API and get a little bit more clarity on how we want to handle catch all error vs how / when we go more granular :)

rootfs/api/urls.py
```diff
@@ -45,7 +45,8 @@
     url(r"^apps/(?P<id>{})/containers/?".format(settings.APP_URL_REGEX),
         views.ContainerViewSet.as_view({'get': 'list'})),
     # application domains
-    url(r"^apps/(?P<id>{})/domains/(?P<domain>[-\._\w]+)/?".format(settings.APP_URL_REGEX),
+    # FIXME make sure * is only matched at the start
+    url(r"^apps/(?P<id>{})/domains/(?P<domain>[-\*\._\w]+)/?".format(settings.APP_URL_REGEX),
```
mboersma: This would probably work. Matches `*.example.com` and `foo.example.com` but not `foo.*.com`:
```python
 url(r"^apps/(?P<id>{})/domains/(?P<domain>\**\.?[-\._\w]+)/?".format(settings.APP_URL_REGEX),
```

helgi: I might be wrong but if I pass in `foo.*.helgi.com` then it would match `foo.` oppose to not matching anything. Got more regex magic?

rootfs/api/serializers.py
```diff
@@ -256,10 +253,11 @@ def validate_domain(self, value):
         if 'xip.io' in value:
             return value
 
+        # Let wildcards through by not trying to validate it
         if labels[0] == '*':
```
bacongobbler: I think we'll also need to add a check here if the length of labels is == 1 along with a unit test. If that returns true, it means the user tried `deis domains:add *`, which if I'm reading right will be accepted and the router would use it as the wildcard domain (in other words, all "failed" requests would fall back to this application). We don't want that to be an accepted value. :)

rootfs/api/urls.py
mboersma: I think bounding the overall regex with slashes ensures that case won't match. I linked to my test expressions below.

    /(?P<domain>\**\.?[-\._\w]+)/

Regular expression visualization

[Debuggex Demo](https://www.debuggex.com/r/b-lamaeKEw0ZrsV8)

helgi: Ah neat - I just realised I had tried it in my regex program using a Ruby type matching so that's probably telling :)

Going to add this regex in and see how it goes!

rootfs/api/serializers.py
helgi: That's fair. I addressed that by adding in a validation if no label is left after the popping, and added it to the "bad domain" test

bacongobbler:
code LGTM, will manually test once I get back home in a few hours (at family's so don't have nearly the amount of bandwidth I normally have)

krancour:
fwiw, I have not gone over this code carefully, but I _have_ tested this hands-on with the complementary changes to the router (deis/router#106) and I am satisfied.

mboersma:
Tests are complaining about an error in the Django migrations now:
```console
CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0005_auto_20160114_0313, 0005_auto_20160208_2156 in api).
To fix them run 'python manage.py makemigrations --merge'
```

helgi:
Yeah, new migrations snuck in via other PRs. I'll adjust

bacongobbler:
Manually tested and all looks good. :+1:

slack:
:magic: :key: :bear:

carmstrong:
:confetti_ball: Excellent work, @helgi!

PR #340, opened 2016-02-04
fix(Makefile,glide.yaml,glide.lock): build client code inside a Docker container

arschles:
Also, ignore appengine packages in the glide.yaml file

Fixes #339

cc/ @helgi

PR #341, opened 2016-02-04
fix(Makefile): Restore local build/install capability

krancour:
I am a big fan of #340 (because I am a fan in general of containerizing our builds), but it had the unfortunate side-effect of removing any possibility of building the client from source on a Mac and then _using_ that client on a Mac-- this is something I do frequently and I suspect I'm not the only one.

This PR restores that possibility through:

```
$ make binary-build install
```

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -30,10 +30,13 @@ glideup:
 	${DEV_ENV_CMD} glide up
 
 build:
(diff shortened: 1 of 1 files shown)


arschles:
@krancour LGTM, sorry about the breakage

carmstrong:
LGTM3

PR #342, opened 2016-02-04
feat(tests): enable branch coverage information

helgi:


rootfs/.coveragerc
```diff
@@ -4,12 +4,12 @@ omit =
     */virtualenv/*
     *tests*
     api/__init__.py
-    api/docker.py
```
helgi: Doesn't exist anymore

PR #346, opened 2016-02-05
feat(builder): add the ability to optionally pass in slug_url in build hook

helgi:
/cc @arschles


rootfs/deis/settings.py
```diff
@@ -289,12 +289,6 @@
 REGISTRY_PORT = os.environ.get('DEIS_REGISTRY_SERVICE_PORT', 5000)
 REGISTRY_URL = '{}:{}'.format(REGISTRY_HOST, REGISTRY_PORT)
 
-BUILDER_HOST = os.environ.get('DEIS_BUILDER_SERVICE_HOST', '127.0.0.1')
-MINIO_HOST = os.environ.get('DEIS_MINIO_SERVICE_HOST', BUILDER_HOST)
-MINIO_PORT = os.environ.get('DEIS_MINIO_SERVICE_PORT', 3000)
```
helgi: Removed these 3 since there is no other usage of them then to populate `S3EP`

rootfs/scheduler/__init__.py
```diff
@@ -605,11 +605,11 @@ def _create_rc(self, name, image, command, **kwargs):  # noqa
         TEMPLATE = RCD_TEMPLATE
 
         # Check if it is a slug builder image.
-        # Example format: golden-earrings:git-5450cbcdaaf9afe6fadd219c94ac9c449bd62413s
         if kwargs.get('build_type', {}) == "buildpack":
```
bacongobbler: can we remove the second argument here? We don't need it to default to anything since it's not evaluated after the `if` statement.

helgi: we can but seems out of scope given that was recently added by @smothiki

mboersma:
@smothiki can you review this since you've been working with the related code?

rootfs/scheduler/__init__.py
bacongobbler: true enough. I'll open a separate issue for this

smothiki:
do we have to add slug_url as a seperate field ? why can't we just pass it as an image?

helgi:
Did it as a separate since it allowed me to turn it into a `URLField`. It could be combined into one but would lose out on the more integrated validation. That could happen in a separate step of course.

rootfs/scheduler/__init__.py
smothiki: If we are checking for slug_url we don't require a check for buildtype here . Just a check to see if slug_url is present or not is sufficient

helgi:
@smothiki What if I turn `image` into a `TextField` and try to do optional URL validation? Less model fields, tho wonder if that's extra complication for little benefit?

smothiki:
Im okay with both the solutions. But prefer less complexity on the build class model. So more into tending towards  changing the image field to text.

smothiki:
this make coding effort on the builder side less. cc @arschles

helgi:
@smothiki changed it to reuse `image` - I do not have any URL validation right now

```
from django.core.validators import URLValidator

# in model
field = models.TextField(validators=[URLValidator()])
```

Would be too aggressive, could write a wrapper around that to be somewhat smarter if we feel it's needed

smothiki:
That's fie even for custom buildpack URL we are not doing a validation.

PR #368, opened 2016-02-13
feat(tz): add pytz as a requirement to get better timezone data (Olsen DB)

helgi:
https://docs.djangoproject.com/en/1.9/topics/i18n/timezones/ recommended


mboersma:
Sounds good--I see we already have `USE_TZ = True` in settings.py.

PR #379, opened 2016-02-16
feat(healthcheck): remove default healthcheck for apps

smothiki:
timeouts and delays defaults to 50 seconds

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -652,7 +652,8 @@ class KubeHTTPClient(AbstractSchedulerClient):
             containers[0]["resources"]["limits"]["cpu"] = cpu
 
         # add in healtchecks
-        js_template = self._healthcheck(js_template, **kwargs['healthcheck'])
+        if kwargs.get('healthcheck'):
+            js_template = self._healthcheck(js_template, **kwargs['healthcheck'])
 
         url = self._api("/namespaces/{}/replicationcontrollers", app_name)
         resp = self.session.post(url, json=js_template)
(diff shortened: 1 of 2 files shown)


helgi:
@smothiki 50 seconds for timeout seems a tad high here. For initial delay it sounds pretty decent, that's what is used as the "warm up cache" type of delay the timeouts will wait from the first time the pod comes up but then the timeout is 50 seconds for hitting `/healthz` or whatever URL is defined.

I understand that timeout would be within the 1 minute connection timeout window but gives me a pause

smothiki:
This PR doesn't fix the failed ruby build pack error . We have to take a deeper look at it.
@helgi I think we should leave setting timeout thing to USER and 50 like close to a min is reasonable.

smothiki:
Let me know if you want me to change delays to a lesser value

helgi:
We'll keep it like this for now at least

rootfs/scheduler/__init__.py
```diff
@@ -652,7 +652,8 @@ def _create_rc(self, name, image, command, **kwargs):  # noqa
             containers[0]["resources"]["limits"]["cpu"] = cpu
 
         # add in healtchecks
-        js_template = self._healthcheck(js_template, **kwargs['healthcheck'])
+        if kwargs.get('healthcheck', {}):
```
bacongobbler: don't we want this to be just `if kwargs.get('healthcheck'):` so it'll fail if the `healthcheck` kwarg doesn't exist?

bacongobbler:
Nice implementation :+1:

rootfs/scheduler/__init__.py
smothiki: I'm just extra cautious. Let me know if you want me to change

bacongobbler: yes please :)

PR #384, opened 2016-02-17
feat(Makefile): organize Makefiles to simplify/unify VERSION

vdice:
VERSION was previously being defined in multiple Makefiles.
This change creates includes.mk to store the canonical version
for this and other values.

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -15,7 +15,7 @@ GO_PACKAGES_REPO_PATH = $(addprefix $(repo_path)/,$(GO_PACKAGES))
 GOFMT = gofmt -e -l -s
 GOTEST = go test --cover --race -v
 
-VERSION := $(shell git describe --tags --abbrev=0 2>/dev/null)+$(shell git rev-parse --short HEAD)
+VERSION ?= $(shell git rev-parse --short HEAD)
 
 define check-static-binary
   if file $(1) | egrep -q "(statically linked|Mach-O)"; then \


client/Makefile
```diff
@@ -15,8 +17,6 @@ GO_PACKAGES_REPO_PATH = $(addprefix $(repo_path)/,$(GO_PACKAGES))
 GOFMT = gofmt -e -l -s
 GOTEST = go test --cover --race -v
 
-VERSION := $(shell git describe --tags --abbrev=0 2>/dev/null)+$(shell git rev-parse --short HEAD)
```
helgi: Note that the `VERSION` in client vs workflow are different. This one combines two git outputs

vdice: Ah, so we wish to maintain this difference?

helgi: @bacongobbler or @mboersma would know better but client and API versioning is slightly different. Especially if we split it out into diff repos like planned

bacongobbler: I'm not sure why the client uses a different version scheme than the rest of the components. For safety's sake let's keep it out of includes.mk for now though :)

includes.mk
```diff
@@ -0,0 +1,12 @@
+COMPONENT ?= workflow
```
bacongobbler: I'm not sure this makes sense in includes.mk, since COMPONENT typically in this case should be set to `deis` for the client

client/Makefile
vdice: Okay, I'll leave the two VERSION... versions. :)  Was there a preference to maintaining the previous client VERSION above or can we transition client's to just the `$(shell git rev-parse --short HEAD)` as used by other components?

bacongobbler: I don't know the history on it, but I don't see why not

includes.mk
vdice: Gotcha.  In that case, since we are currently only building/pushing all of workflow as a Docker image, with no immediate plans to build/push a Docker image of the client (right?), and since COMPONENT and VERSION wish not to be shared, perhaps the `includes.mk` is a bit premature at this point?

bacongobbler: Perhaps. DEIS_REGISTRY, IMAGE and IMAGE_PREFIX are also only used by workflow so they're not shared as well. `make info` makes sense to keep in, but at the same time the client doesn't deploy an image or uses a dev reigstry so `make -C client info` would be displaying "useless" information as well.

I'm all for this PR if workflow and the client shared more variables together, but it seems like VERSION is the only candidate at this time, and even then they differ (for some unknown reason). Maybe we should revisit this once we figure out why the client version differs from the components? :)

bacongobbler:
And interestingly enough, this PR is failing due to the following:

```
Makefile:3: ../includes.mk: No such file or directory
```

vdice:
Thanks, @bacongobbler ; yea spotted that but deciding to remove `includes.mk` for the time being.  Travis wasn't seeing the file based on the Docker mount dir...

Makefile
```diff
@@ -2,7 +2,7 @@
 DEIS_REGISTRY ?= $(DEV_REGISTRY)
 IMAGE_PREFIX ?= deis
 COMPONENT ?= workflow
-VERSION ?= git-$(shell git rev-parse --short HEAD)
+VERSION ?= $(shell git rev-parse --short HEAD)
```
bacongobbler: would you mind keeping this as `git-$(shell git rev-parse --short HEAD)` and change the client Makefile to match so it matches https://github.com/deis/prototype-repo/blob/master/Makefile#L12-L14? :)

sgoings: That would create binaries named: `deis-git-<sha>-<os>-<arch>`

We decided the `git` was not necessary and was actually confusing.

bacongobbler: okay let's do that for the client, but let's keep in the `git-` convention for workflow, because that's the current version convention for [the](https://github.com/deis/builder/blob/master/Makefile#L15) [other](https://github.com/deis/postgres/blob/master/Makefile#L9) [components](https://github.com/deis/router/blob/master/Makefile#L4-L5).

client/Makefile
```diff
@@ -15,7 +15,7 @@ GO_PACKAGES_REPO_PATH = $(addprefix $(repo_path)/,$(GO_PACKAGES))
...
+VERSION ?= $(shell git rev-parse --short HEAD)
```
vdice: So, sorry for all the noise -- this is the only change for this PR, to align with what deis.io looks for in bintray:

```
 ~/code/deis.io/deis-cli  ⑂ update-deiscli-bin-name    ./install-v2-alpha.sh
Downloading deis-git-5e08df4-darwin-amd64 from Bintray...
```

bacongobbler:
quite the rollercoaster! :)

PR #385, opened 2016-02-17
feat(client): add -d to gofmt to get more info on style failures

helgi:


mboersma:
Does adding `-d` (show diffs instead of rewrite files) actually change the `gofmt` output on failure? I think `-l` (just list the files) effectively overrides it. That is, I got the same output from `make -C client/ test-style` before and after this change, whether or not the go code has an error that `gofmt` notices.

helgi:
it works but blagh... our makefile isn't doing us any favours:

```
make test
/bin/sh: -c: line 0: syntax error near unexpected token `('
/bin/sh: -c: line 0: `if [ cmd/ps.go diff cmd/ps.go gofmt/cmd/ps.go --- /var/folders/cb/xyx1dpd527z951dzq3f80r7r0000gp/T/gofmt219126809	2016-02-17 12:45:53.000000000 -0800 +++ /var/folders/cb/xyx1dpd527z951dzq3f80r7r0000gp/T/gofmt557976740	2016-02-17 12:45:53.000000000 -0800 @@ -118,7 +118,7 @@    	if len(processes) == 0 {  		fmt.Println("Could not find any processes to restart") -	} else{ +	} else {  		fmt.Printf("done in %ds\n", int(time.Since(startTime).Seconds()))  		printProcesses(appID, processes)  	} ]; then 	echo "gofmt check failed:"; gofmt -e -l -s -d deis.go deis_test.go cmd controller/api controller/client controller/models/apps controller/models/auth controller/models/builds controller/models/certs controller/models/config controller/models/domains controller/models/keys controller/models/perms controller/models/ps controller/models/releases controller/models/users parser pkg/git pkg/webbrowser; exit 1; fi'
```

PR #391, opened 2016-02-18
fix(auth): return a 409 if a user is cancelled that has apps

helgi:
```
deis auth:cancel --username=slack
cancel account slack at http://deis.172.17.8.100.xip.io? (y/N): y
Error:
409 Conflict
detail: slack still has applications assigned. Delete or transfer ownership
```

Can't do anything about the `Error` info there until we refactor the client

### DIFF
--- a/client/cmd/auth.go
+++ b/client/cmd/auth.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"net/url"
 	"os"
+	"reflect"
 	"strings"
 	"syscall"
 
@@ -241,8 +242,11 @@ func Cancel(username string, password string, yes bool) error {
 	}
 
 	err = auth.Delete(c, username)
-
-	if err != nil {
+	cleanup := fmt.Errorf("\n%s %s\n\n", "409", "Conflict")
+	if reflect.DeepEqual(err, cleanup) {
+		fmt.Printf("%s still has application associated with it. Transfer ownership or delete them first\n", username)
+		return nil
+	} else if err != nil {
 		return err
 	}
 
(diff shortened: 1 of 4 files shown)


sgoings:
You even wrote unit tests! WOOO!

client/cmd/auth.go
```diff
@@ -241,13 +242,16 @@ func Cancel(username string, password string, yes bool) error {
 	}
 
 	err = auth.Delete(c, username)
-
-	if err != nil {
+	cleanup := fmt.Errorf("\n%s %s\n\n", "409", "Conflict")
+	if reflect.DeepEqual(err, cleanup) {
```
arschles: @helgi we talked about this offline, but capturing here. I'd like `auth.Delete` to put this error string in its API contract, or better yet have it return a specific, exported error (like `409Error` or something), so that we can get stronger compile time checks

arschles:
@helgi good to go, pending CI. I made a line comment about this too, but can you create an issue for the client error handling refactor?

client/cmd/auth.go
helgi: I can file an issue, yeah. Given how the error is raised way down in the `http.go` portion then it would be a bit of a change required to pull that off since it wouldn't be a one off.

mboersma:
This broke the e2e tests immediately:
```
Logged in as test-81
Please log in again in order to cancel this account
Logged in as test-81
Error: 
409 Conflict
detail: test-81 still has applications assigned. Delete or transfer ownership

Failure [0.175 seconds]
[AfterSuite] AfterSuite 
/go/src/github.com/deis/workflow-e2e/tests/tests_suite_test.go:140

  No future change is possible.  Bailing out early after 0.085s.
  Expected
      <int>: 1
  to match exit code:
      <int>: 0

  /go/src/github.com/deis/workflow-e2e/tests/tests_suite_test.go:174
```

PR #393, opened 2016-02-18
fix(local_settings.py): read DATABASES from settings.py

bacongobbler:
This fix renames and moves confd_settings.py to local_settings.py
within the container. This allows us to import the generic settings.py
and overwrite existing values, allowing the database settings to be
overwritten.


bacongobbler:
this in tandem with https://github.com/deis/charts/pull/101 fixes the issues seen in master

PR #396, opened 2016-02-18
chore(glide): update to glide 0.9 and use the resolved import lock file

helgi:
Depends on #395


arschles:
:100: on all the red. good to go pending CI

PR #399, opened 2016-02-19
docs(understanding-deis): document database lifecycle

bacongobbler:
closes deis/postgres#34

### DIFF
--- a/docs/src/understanding-deis/components.md
+++ b/docs/src/understanding-deis/components.md
@@ -11,8 +11,9 @@ The `deis` command-line client interacts with this component.
 
 ## Database
 
-The database component is a [PostgreSQL][] server used to store durable
-platform state. Backups and WAL logs are pushed to [Store][].
+The database component is a [PostgreSQL][] server used to store durable platform state. Backups and
+WAL logs are pushed to the [Store][] through [WAL-E][]. When the database is restarted, backups are
+fetched and replayed from Store so no data is lost.
 
 ## Builder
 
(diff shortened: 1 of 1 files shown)


slack:
The backup and restore doc is still TODO in master.

bacongobbler:
I don't know enough about minio to comment on how minio should be properly backed up to persistent disk, so I left that out in this PR.

docs/src/understanding-deis/components.md
```diff
@@ -41,18 +42,22 @@ The router component uses [Nginx][] to route traffic to application containers.
 
 ## Store
 
-The store component uses [Ceph][] to store data for Deis components
-which need to store state, including [Registry][], [Database][].
+The Store component uses [Minio][] to store data for Deis components
```
arschles: this should probably say that the store component uses minio by default - you can configure the cluster to use other object stores

slack:
Object storage should generally be transparent to the process of restoring a postgres backup, no?

docs/src/understanding-deis/components.md
```diff
@@ -41,18 +42,22 @@ The router component uses [Nginx][] to route traffic to application containers.
...
+which need to store state, including the [Registry][] and the [Database][].
+
+It is highly recommended that the Store's data is backed by a persistent volume in the event of
```
arschles: same idea here - if you _do_ use minio, you should back it by a persistent volume

docs/src/understanding-deis/components.md
bacongobbler: good point

bacongobbler:
@slack yes, generally it is completely transparent. The user doesn't need to know anything about the underlying object storage layer we provide unless they swap it out with S3.

PR #402, opened 2016-02-19
fix(logger): Return non-zero exit code when deis logs fails

### LINKED ISSUE #401: deis logs returns wrong exit code
When `deis logs` is performed it swallows the exceptions returned and instead prints out a user friendly  message to help them debug the issue. The problem was that this code just returned a string and a nil error which made the client thing everything was ok. Instead, we should return an empty string with a properly formatted error message so that way we return a non-zero exit code.

jchauncey:
fixes #401
Create a user friendly error message and return that instead of a string.

### DIFF
--- a/client/controller/models/apps/apps.go
+++ b/client/controller/models/apps/apps.go
@@ -98,14 +98,14 @@ func Logs(c *client.Client, appID string, lines int) (string, error) {
 	body, err := c.BasicRequest("GET", u, nil)
 
 	if err != nil || len(body) < 1 {
-		return fmt.Sprintf(
+		return "", fmt.Errorf(
 			`There are currently no log messages. Please check the following things:
 1) Logger and fluentd pods are running.
 2) If you just installed the logger components via the chart, please make sure you restarted the workflow pod.
 3) The application is writing logs to the logger component.
 You can verify that logs are appearing in the logger component by issuing the following command:
 curl http://<log service ip>:8088/logs/%s on a kubernetes host.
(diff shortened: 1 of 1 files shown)


jchauncey:
```
╭─jonathanchauncey at ENG000637 in ~/projects/deis/example-go on master✔ using ‹2.0.0-p645›
╰─± deis logs
There are currently no log messages. Please check the following things:
1) Logger and fluentd pods are running.
2) If you just installed the logger components via the chart, please make sure you restarted the workflow pod.
3) The application is writing logs to the logger component.
You can verify that logs are appearing in the logger component by issuing the following command:
curl http://<log service ip>:8088/logs/foo on a kubernetes host.
To get the service ip you can do the following: kubectl get svc deis-logger --namespace=deis
╭─jonathanchauncey at ENG000637 in ~/projects/deis/example-go on master✔ using ‹2.0.0-p645›
╰─± echo $?
0
╭─jonathanchauncey at ENG000637 in ~/projects/deis/example-go on master✔ using ‹2.0.0-p645›
╰─± deis logs
Error: There are currently no log messages. Please check the following things:
1) Logger and fluentd pods are running.
2) If you just installed the logger components via the chart, please make sure you restarted the workflow pod.
3) The application is writing logs to the logger component.
You can verify that logs are appearing in the logger component by issuing the following command:
curl http://<log service ip>:8088/logs/foo on a kubernetes host.
To get the service ip you can do the following: kubectl get svc deis-logger --namespace=deis
╭─jonathanchauncey at ENG000637 in ~/projects/deis/example-go on master✔ using ‹2.0.0-p645›
╰─± echo $?
1
```

PR #407, opened 2016-02-20
chore(client): bump go-dev to 0.8.0 for glide 0.9.0

slack:
Help speed up the build a bit with newer Glide.


slack:
Old master was old.

PR #409, opened 2016-02-22
fix(logs): Update the array index for determining the color category

### LINKED ISSUE #408: Colorizing of log message is currently broken
While we are correctly parsing off the pod name from the log message the conditional used to setup the color of the message is using the wrong array check.

jchauncey:
fixes #408

### DIFF
--- a/client/cmd/apps.go
+++ b/client/cmd/apps.go
@@ -175,9 +175,7 @@ func printLogs(logs string) error {
 	for _, log := range strings.Split(logs, `\\n\\n`) {
 		category := "unknown"
 		parts := strings.Split(strings.Split(log, " -- ")[0], " ")
-		if len(parts) >= 2 {
-			category = parts[1]
-		}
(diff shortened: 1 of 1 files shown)


client/cmd/apps.go
```diff
@@ -175,8 +175,8 @@ func printLogs(logs string) error {
 	for _, log := range strings.Split(logs, `\\n\\n`) {
 		category := "unknown"
 		parts := strings.Split(strings.Split(log, " -- ")[0], " ")
-		if len(parts) >= 2 {
-			category = parts[1]
+		if len(parts) >= 1 {
```
bacongobbler: I don't think this check is needed any more. `strings.Split()` will always return a []string with a length >= 1. This was just to make sure that the string itself had been split at least once.

jchauncey: Ok I'll try that out and resubmit (need to rebase anyways)

PR #411, opened 2016-02-23
feat(pods): wait until pod / container has reported success on readiness / liveness probe during scale event

helgi:


mboersma:
I tested this locally. It seems to fix the failing test cases in deis/workflow-e2e#53.

slack:
Working well on my cluster.

PR #413, opened 2016-02-23
feat(github): add pull request template

bacongobbler:
This pull request template describes the typical workflow each
pull request goes through. This just formalizes it so users and
core contributors have a checklist to go through when reviewing
pull requests.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -1,0 +1,52 @@
+# Summary of Changes
+
+Please explain what this PR changes, and include what unit tests you've written (we hope you wrote tests!)
+
+# Issue(s) that this PR Closes
+
+Please list the issue(s) that this PR closes, similar to the below:
+
+closes #1234
+closes #5678
+
+# Associated [End To End Test](https://github.com/deis/workflow-e2e) PR(s)
+
+If your change requires any additions or changes to the [end to end tests](https://github.com/deis/workflow-e2e), please submit them as 1 or more pull requests against that repo and refer to them here.
+
+refs deis/workflow-e2e#1234
+refs deis/workflow-e2e#5678
+
+# Associated [Documentation](https://github.com/deis/workflow) PR(s)
+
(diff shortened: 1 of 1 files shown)


jchauncey:
my only suggestion is to make those items into sentences with correct grammar.

arschles:
FWIW, my original reaction to PR templates was that they would be best used for providing a 'skeleton' that has fields for a contributor to fill in, instead of interrogatives. However, I'm fine with seeing how this works. 

Also, this looks like it needs a rebase

bacongobbler:
@jchauncey suggestions on improvements? I tried to make the points as clear as possible while fitting on one line.

@arschles I share your concerns as well. I wanted to shake things out and see if we liked this way of contributing. However, the issue format for docker/docker leaves a bad taste in my mouth when submitting issues, so maybe this is overkill? Open to suggestions/killing this off

arschles:
@bacongobbler I def don't think this is overkill and I'd love to get something like this in!

In case it helps illustrate what I was thinking, here's my proposal, adapted from the changes herein:
```
# Summary of Changes

# Issues that this PR Closes

# Associated [Documentation](https://github.com/deis/docs-v2) PR(s)

# Associated [Design Document](http://docs.deis.io/en/latest/contributing/design-documents.html#design-documents)(s)

# Pull Request Hygiene TODOs

- [ ] Your pull request is concise and to the point (make another PR for refactoring nearby code)
- [ ] Your commits are squashed into logical units of work
- [ ] Your commits follow the [commit style guidelines](http://docs.deis.io/en/latest/contributing/standards.html#commit-style)
```

It might be a little impersonal, so could add back some of the "thank you" sections...

jchauncey:
I was just thinking capitalization and punctuation

bacongobbler:
@arschles needs more flowers :cherry_blossom: 

In all seriousness that was I was trying to target, so thank you for the input. I'll rewrite the PR template to follow that template

bacongobbler:
I changed the template to follow deis/builder's example.

bacongobbler:
merging as this is a template that's not tested by CI.

PR #415, opened 2016-02-24
fix(client): exit 1 with malformed config:set input

### LINKED ISSUE #228: `config:set` complains about formatting of `KEY=VAL` but merrily continues
```console
example-go [master]$ deis2 config:set POWERED_BY pushups
'POWERED_BY' does not match the pattern 'key=var', ex: MODE=test
'pushups' does not match the pattern 'key=var', ex: MODE=test
Creating config... ..o^...^o..done

=== mellow-zaniness Config
$
```

Should most likely exit 1.

bacongobbler:
closes #228


helgi:
Can we gets a test?

bacongobbler:
ask and ye shall receive

PR #416, opened 2016-02-24
ref(scheduler): improve scale operation logging and performance

helgi:
The idea here is to introduce far more comprehensive logging on scale operations to give both better insight into where scaling is slow and generally give an operator an idea of what step it is on.

I've tested a normal scale up and scale down + a deploy using `config:set` but only on docker images

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -1,5 +1,6 @@
 import json
 import logging
+import os
 import re
 import string
 import time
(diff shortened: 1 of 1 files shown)


bacongobbler:
even if we can't get test coverage for this, more debug logging is awesome. :+1:

helgi:
> even if we can't get test coverage for this, more debug logging is awesome. :+1:

yet, we can't get it just yet. I got some plans :microscope:

slack:
This with v2-beta builder shaved roughly 60 seconds from a deploy across 5 replicas. (1m40s to 45s)

PR #421, opened 2016-02-26
feat(config): add pipe suport to config:pull

Joshua-Anderson:
Fixes deis/deis#4697


mboersma:
I tested this locally and it works as advertised. Thanks @Joshua-Anderson!

PR #426, opened 2016-02-27
fix(cmd/apps): simplify URL prefixing

mboersma:
Go's net/url package returns malformed urls when a port is specified to `url.Parse()`.

For example, given "kiddie-armchair.192.168.64.2.xip.io:31258", `deis open` would call this `AppOpen` func and open a browser to "http:31258". The code appears to intend just to ensure a URL scheme prefix, so string manipulation should be adequate.

### DIFF
--- a/client/cmd/apps.go
+++ b/client/cmd/apps.go
@@ -2,7 +2,6 @@ package cmd
 
 import (
 	"fmt"
-	"net/url"
 	"os"
 	"strings"
 	"time"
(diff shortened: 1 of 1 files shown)


Joshua-Anderson:
I'm not sure this is the right fix for the problem.
Commented on deis/deis#4939

bacongobbler:
looks like a reasonable fix considering the downstream conversation in deis/deis#4939.

jchauncey:
@mboersma this is good to go. but needs rebase

arschles:
LGTM3. If it matters in the future, [urlx](https://godoc.org/github.com/goware/urlx) could help solve similar problems to this one.

PR #431, opened 2016-02-29
fix(scheduler): if a pod is in pending there is no start time, instead use now()

### LINKED ISSUE #429: 503 error on ps:scale
Some of the e2e process tests in ps_tests.go fail in this way:
```console
Processes with a deployed app can scale up and down 
  scales to 1
  /Users/matt/Projects/src/github.com/deis/workflow-e2e/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:37
$ deis login http://deis.192.168.64.2.xip.io:32370 --username=test-81 --password=asdf1234
Logged in as test-81
$ deis ps:scale web=1 --app=test-897795369
Scaling processes... but first, Slurm!
done in 0s
Error: 
503 Service Unavailable
detail: 'startTime'


• Failure [1.610 seconds]
Processes
/Users/matt/Projects/src/github.com/deis/workflow-e2e/tests/ps_test.go:155
  with a deployed app
  /Users/matt/Projects/src/github.com/deis/workflow-e2e/tests/ps_test.go:154
    can scale up and down
    /Users/matt/Projects/src/github.com/deis/workflow-e2e/vendor/github.com/onsi/ginkgo/extensions/table/table.go:96
      scales to 1 [It]
      /Users/matt/Projects/src/github.com/deis/workflow-e2e/vendor/github.com/onsi/ginkgo/extensions/table/table_entry.go:37

      No future change is possible.  Bailing out early after 0.131s.
      Got stuck at:
          
          
      Waiting for:
          === test-897795369 Processes

      /Users/matt/Projects/src/github.com/deis/workflow-e2e/tests/ps_test.go:58
```

helgi:
Fixes #429

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -1,3 +1,4 @@
+from datetime import datetime
 import logging
 import re
 import requests
@@ -538,7 +539,11 @@ class App(UuidAuditedModel):
                 item['state'] = self._scheduler.resolve_state(p).name
                 item['release'] = p['metadata']['labels']['version']
                 item['type'] = p['metadata']['labels']['type']
-                item['started'] = p['status']['startTime']
+                if 'startTime' in p['status']:
+                    started = p['status']['startTime']
+                else:
+                    started = str(datetime.utcnow().strftime(settings.DEIS_DATETIME_FORMAT))
+                item['started'] = started
 
                 data.append(item)
 


jchauncey:
is there a way to test this?

helgi:
Best way is to do something like `ps:scale web=10` and then right away in another window have `ps:list -a foo` ready so it catches pods in very early stages.

It happens when you get a pod in Pending state

rootfs/api/models/app.py
```diff
@@ -538,7 +539,11 @@ def list_pods(self, *args, **kwargs):
                 item['state'] = self._scheduler.resolve_state(p).name
                 item['release'] = p['metadata']['labels']['version']
                 item['type'] = p['metadata']['labels']['type']
-                item['started'] = p['status']['startTime']
+                if 'startTime' in p['status']:
+                    started = p['status']['startTime']
+                else:
+                    started = str(datetime.utcnow().strftime('%Y-%m-%dT%H:%M:%SZ'))
```
bacongobbler: we should use `settings.DEIS_DATETIME_FORMAT` so we can change this if necessary: https://github.com/deis/workflow/blob/master/rootfs/deis/settings.py#L148

helgi: done

mboersma:
I tested this with the e2e ps_tests and didn't see any 503s. :+1:

PR #434, opened 2016-02-29
ref(scheduler): logs in scheduler does not appear to be used anymore

helgi:
I can not spot where it is used. Potentially it got accidentally refactored out but can't see where that would be


bacongobbler:
Yeah I think this was refactored out in favor of deis/logger. Probably some tech preview cruft that was left over See https://github.com/deis/workflow/blob/master/rootfs/api/models/app.py#L463-L487 for how it's used now.

:+1: for more code removal

PR #435, opened 2016-02-29
ref(scheduler): move docker port discovery to a function for cleanliness

helgi:


rootfs/scheduler/__init__.py
```diff
@@ -543,6 +543,20 @@ def resolve_state(self, pod):
...
+        try:
+            image = self.registry + '/' + image
+            repo = image.split(":")
+            # image already includes the tag, so we split it out here
+            docker_cli = Client(version="auto")
+            docker_cli.pull(repo[0]+":"+repo[1], tag=repo[2], insecure_registry=True)
+            image_info = docker_cli.inspect_image(image)
+            port = int(list(image_info['Config']['ExposedPorts'].keys())[0].split("/")[0])
+        except:
+            port = 5000
```
mboersma: Maybe add debug output here?
```python
logger.debug("Failed to find port for Docker image {}, defaulting to 5000".format(image))
```

helgi: Yeah, I can do that. I was just moving code around really so that's why it wasn't there. Hold Space Commander

PR #439, opened 2016-03-01
ref(scheduler): use proper k8s naming (namespace vs app_name) and pass that in properly

helgi:
Also update `_get_job_id` to be more like the format we expect so transformations do not need to happen inside the scheduler

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -90,7 +90,7 @@ class App(UuidAuditedModel):
         app = self.id
         release = self.release_set.latest()
         version = "v{}".format(release.version)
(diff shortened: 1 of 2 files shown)


rootfs/api/models/app.py
```diff
@@ -418,7 +418,6 @@ def _deploy_app(self, scale_types, release, existing):
                 'cpu': release.config.cpu,
                 'tags': release.config.tags,
                 'envs': release.config.values,
-                'aname': self.id,
```
mboersma: Excellent. This variable name always made me annoyed.

helgi: You and me both!

PR #449, opened 2016-03-02
fix(models): catch proper kube exceptions in certficates so secrets get created

### LINKED ISSUE #444: certs: attach returns 503 or 404; not reflected in CLI
Consider the following scenario:
```
$ deis domains:add foo.com -a test
Adding foo.com to test... done
$ deis certs:attach foo-com foo.com
Attaching certificate foo-com to domain foo.com... done
$ deis certs
       Name      | Common Name | SubjectAltName |        Expires         |   Fingerprint   | Domains |  Updated   |  Created
+----------------+-------------+----------------+------------------------+-----------------+---------+------------+------------+
  foo-com        | foo.com     |                | 1 Mar 2017 (in 1 year) | 25:64[...]0E:59 | foo.com | 1 Mar 2016 | 1 Mar 2016
```
Although the `foo-com` cert appears to be attached successfully to the `foo.com` domain, router had returned the following 503 error when the `certs:attach` command was executed:

```
10.240.0.9 "POST /v2/certs/foo-com/domain/ HTTP/1.1" 503 304 "Deis Client v2.0.0-dev"
```

This can sometimes also be a 404.  In both cases, the error is silently swallowed up. cc @helgi 

In the end, @helgi and I are thinking this is preventing the proper function of the SSL endpoint on the kubernetes cluster:

```
$ curl -I -H "Host: foo.com" https://deis.104.197.31.94.xip.io
curl: (35) Server aborted the SSL handshake
```

helgi:
Fixes #444

### DIFF
--- a/rootfs/api/models/certificate.py
+++ b/rootfs/api/models/certificate.py
@@ -14,6 +14,7 @@ from django.contrib.postgres.fields import ArrayField
 from api.models import AuditedModel, validate_label, AlreadyExists
 from api.models.domain import Domain
 
+from scheduler import KubeHTTPException
 
 import logging
 logger = logging.getLogger(__name__)
@@ -174,7 +175,7 @@ class Certificate(AuditedModel):
         try:
             # We raise an exception when a secret doesn't exist
             self._scheduler._get_secret(app, name)
-        except RuntimeError:
+        except KubeHTTPException:
             self._scheduler._create_secret(app, name, data)
 
         # get config for the service
@@ -206,7 +207,7 @@ class Certificate(AuditedModel):
             # We raise an exception when a secret doesn't exist
             self._scheduler._get_secret(app, name)
             self._scheduler._delete_secret(app, name)
-        except RuntimeError as e:
+        except KubeHTTPException as e:
             logger.critical(e)
 
         # get config for the service


krancour:
I know the behavior I'm about to comment on wasn't eirectly introduced by this PR, but I am just noticing...

I think it's generally preferred to avoid raising exceptions to handle what are actually not error conditions, where a conditional could be used to direct the flow of execution instead.  So in this case, it would be better if `self._scheduler._get_secret` were to return a null value if the requested secret doesn't exist, and you could check for that and create the secret under that circumstance.

helgi:
Yeah, but that's how the whole scheduler works, and in fact Django. If you query the DB via the ORM you get a `DoesNoExists` Exception originating from the model.

It's on the agenda to figure out the error handling flow for the kube client, just not yet. Especially since I haven't figured out a good plan of attack, not fond of exceptions for this either but a `None` or `False` always felt wonky as well since well I may want to know why the `get` returned those values oppose to something else and then you end up with `get_last_error` and other error introspection methods. Same happens if you start mapping `404` to `None` then what is `409` and actual errors but are soft errors.

Anyway, thoughts of the day

jchauncey:
How should I test this locally?

vdice:
@jchauncey https://github.com/deis/workflow/issues/444

krancour:
> Yeah, but that's how the whole scheduler works, and in fact Django. If you query the DB via the ORM you get a DoesNoExists Exception originating from the model.

I don't know Django (or how it does ORM) too well, but I know other similar frameworks well enough to be a little surprised and turned off by that.  Something not existing isn't inherently an error.  It's only an error when your business logic decides that something that doesn't exist _should have_ and we therefore have a problem on our hands.

I know I'm getting a bit philosophical here.  Sorry.

At any rate, if this is how things are currently done with Django in general, or it is how things are currently done throughout _this_ Django app, then this certainly shouldn't stand in the way of a merge. :+1:

jchauncey:
hibernate and eclipselink/toplink return DoNotExist exceptions if a row isnt found in the database. Using exceptions as flow control is much more graceful than null/nil checks that other langauges like ruby/go use.

It allows you to be way more precise in the errors you want to handle (who knows why you have a nil value).

PR #456, opened 2016-03-02
feat(version): add build version to deis -v output

vdice:


sgoings:
Wicked sweet, @vdice!

client/parser/version.go
```diff
@@ -20,6 +20,6 @@ Use 'deis help [command]' to learn more.
 		return err
 	}
 
-	fmt.Println(version.Version)
+	fmt.Printf("%s-%s\n", version.Version, version.BuildVersion)
```
bacongobbler: We should make this conditional on `version.BuildVersion`'s value. For example, if we wanted to produce an official client with no build version number:

```
v := version.Version
if version.BuildVersion != "" {
    v = fmt.Sprintf("%s-%s", version.Version, version.BuildVersion)
}
fmt.Println(v)
```

vdice: Added; good call @bacongobbler

PR #465, opened 2016-03-03
fix(make): determine host OS for supplying to (container-ized) go build

vdice:
In #456 with a slight re-arrangement in client's Makefile, `go build`, being container-ized, was defaulting to producing a linux binary.  We now need to pass in host OS for building on darwin cc @mboersma


mboersma:
LGTM.

This change in `glide` 0.9.1 seems possibly relevant: Masterminds/glide#267.

PR #466, opened 2016-03-03
feat(make): make deploy to apply the latest RC image for workflow

helgi:
run `make deploy` after a `git commit` to roll the image, apply it and scale to whatever replica count the component was set to.

Does not touch helm, only uses the already applied information from k8s


Makefile
```diff
@@ -53,6 +57,11 @@ update-manifests:
 	sed 's#\(image:\) .*#\1 $(IMAGE)#' manifests/deis-workflow-rc.yml \
 		> manifests/deis-workflow-rc.tmp.yml
 
+deploy: docker-build docker-push
+	sed 's#\(image:\) .*#\1 $(IMAGE)#' /tmp/deis-$(COMPONENT) | kubectl apply --validate=true -f -
+	kubectl scale rc deis-workflow --replicas 0 --namespace deis
+	kubectl scale rc deis-workflow --replicas $(DESIRED_REPLICAS) --namespace deis
+
```
sgoings: You're pooping `deis-$(COMPONENT)` files in my temp directory, @helgi. :laughing:

helgi: Correct!

PR #467, opened 2016-03-03
fix(certs): propogate err in certs:attach/detach/remove

### LINKED ISSUE #462: certs: attach/detach with non-existent cert/domains does not error
Consider the following scenario on a fresh deis app:
```
$ deis domains:add foo.com -a app
Adding foo.com to app... done
$ deis certs:attach non-existent-cert foo.com
Attaching certificate non-existent-cert to domain foo.com... done
$ deis certs:detach non-existent-cert foo.com
Detaching certificate non-existent-cert from domain foo.com... done
$ deis certs:add real-cert server.crt server.key
Adding SSL endpoint... done
$ deis certs:attach real-cert non.existent.domain
Attaching certificate real-cert to domain non.existent.domain... done
$ deis certs:detach real-cert non.existent.domain
Detaching certificate real-cert from domain non.existent.domain... done
```
Anytime a non-existent cert or non-existent domain is involved in `certs:attach` or `certs:detach`, I would expect to see an error, as opposed to the misleading affirmative messages seen above.  

(Will be covered by `certs_test.go` in `workflow-e2e`)

vdice:
...instead of ignoring and reporting misleading affirmation.  Will be covered in e2e certs_test(s)...

So instead of this:
```
$ deis certs:attach foo goo
Attaching certificate foo to domain goo... done
```

We will currently get this:
```
$ deis certs:attach foo goo
Attaching certificate foo to domain goo... Error:
404 Not Found
detail: Not found.
```

(Where `foo` is not a valid cert and/or `goo` is not a valid domain)

Closes #462

### DIFF
--- a/client/cmd/certs.go
+++ b/client/cmd/certs.go
@@ -142,7 +142,7 @@ func CertRemove(name string) error {
 	fmt.Printf("Removing %s... ", name)
 	quit := progress()
 
-	certs.Delete(c, name)
+	err = certs.Delete(c, name)
 
 	quit <- true
 	<-quit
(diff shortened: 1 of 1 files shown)


arschles:
LGTM3

PR #469, opened 2016-03-03
chore(manifests): remove manifests as now helm install works + make deploy for development needs

helgi:


Makefile
```diff
@@ -71,14 +43,6 @@ commit-hook:
 full-clean: check-docker
 	docker images -q $(IMAGE_PREFIX)$(COMPONENT) | xargs docker rmi -f
 
-postgres:
```
mboersma: Can we leave this target in? I use `make postgres` all the time, so I can run `make test-unit`...

helgi: Added back in

README.md
```diff
@@ -55,16 +63,18 @@ registry. The `$DEIS_REGISTRY` environment variable must point to a registry acc
...
+When you want to test changes then commit the changes to your branch and run
+
 ```console
-$ make docker-build docker-push
+$ make deploy
 ```
 
-You'll want to modify the deis chart to use your custom image, then run `helm install` on the
-chart.
+This will build the required docker images and push up to the registry that was configure, update the Replication Controller
```
mboersma: I would reword this a bit:

> This will build the required docker images and push them to the registry that was configured, then update and recreate the Replication Controller.

helgi: Changed it

PR #470, opened 2016-03-03
fix(client/Makefile): don't export GOOS when running gox

### LINKED ISSUE #468: Client: "cannot execute binary file" when following v2 guide
Following the directions to install the deis client (http://docs-v2.readthedocs.org/en/latest/using-deis/installing-the-client/), I get the following:

```
$ curl -sSL http://deis.io/deis-cli/install-v2-alpha.sh | bash
Downloading deis-7ba6255-darwin-amd64 from Bintray...
deis is now available in your current directory.
To learn more about deis, execute:
    $ ./deis --help
```

This provides me a binary that I attempt to execute and get an error, slightly different if I'm on zsh or bash:

```
$ ./deis
zsh: exec format error: ./deis
$ bash
$ ./deis
bash: ./deis: cannot execute binary file
```

From what I can tell, the binary is for the correct platform (OS X El Capitan in this case):

```
$ file deis
deis: ELF 64-bit LSB executable, x86-64, version 1 (SYSV), statically linked, not stripped
$ uname -mps
Darwin x86_64 i386
```

Any ideas on what might be causing this? I ended up building the deis client myself and had success, so that's my recommendation for anybody else running into this issue: https://github.com/deis/workflow/tree/master/client

vdice:
Due to https://github.com/mitchellh/gox/issues/50, if GOOS is set to particular OS,
gox will build all 'cross-compiled' binaries for that GOOS.

Fixes #468

### DIFF
--- a/client/Makefile
+++ b/client/Makefile
@@ -12,7 +12,7 @@ endif
 
 DEV_ENV_IMAGE := quay.io/deis/go-dev:0.9.0
 DEV_ENV_WORK_DIR := /go/src/${repo_path}
-DEV_ENV_PREFIX := docker run --rm -e GO15VENDOREXPERIMENT=1 -e CGO_ENABLED=0 -e GOOS=${GOOS} -v ${CURDIR}:${DEV_ENV_WORK_DIR} -w ${DEV_ENV_WORK_DIR}
+DEV_ENV_PREFIX := docker run --rm -e GO15VENDOREXPERIMENT=1 -e CGO_ENABLED=0 -v ${CURDIR}:${DEV_ENV_WORK_DIR} -w ${DEV_ENV_WORK_DIR}
 DEV_ENV_CMD := ${DEV_ENV_PREFIX} ${DEV_ENV_IMAGE}
 DIST_DIR := _dist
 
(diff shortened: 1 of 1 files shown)


client/Makefile
```diff
@@ -59,7 +59,7 @@ build-all:
 	-output="$(DIST_DIR)/deis-${VERSION}-{{.OS}}-{{.Arch}}" .
 
 binary-build:
-	${DEV_ENV_CMD} go build -a -installsuffix cgo -ldflags "-s -X ${repo_path}/version.BuildVersion=${VERSION}" -o deis .
+	${DEV_ENV_PREFIX} -e GOOS=${GOOS} ${DEV_ENV_IMAGE} go build -a -installsuffix cgo -ldflags "-s -X ${repo_path}/version.BuildVersion=${VERSION}" -o deis .
```
sgoings: Why can't we just put `${DEV_ENV_IMAGE}` in `DEV_ENV_PREFIX`?

sgoings: Oops! I understand this change now. We were using `DEV_ENV_CMD` before, and we can't now because of the addition of geese.

PR #476, opened 2016-03-04
fix(logger): Update the port that workflow uses to find logger

jchauncey:
1. install deis-dev
1. install deis-logger
1. restart workflow
1. create app of your choice
1. trigger log creation
1. run `deis logs` on said app
1. make sure logs show up


rootfs/deis/settings.py
```diff
@@ -270,7 +270,7 @@
 
 # logger settings
 LOGGER_HOST = os.environ.get('DEIS_LOGGER_SERVICE_HOST', '127.0.0.1')
-LOGGER_PORT = os.environ.get('DEIS_LOGGER_PORT_8088_TCP_PORT', 8088)
+LOGGER_PORT = os.environ.get('DEIS_LOGGER_SERVICE_PORT_HTTP', 80)
```
arschles: :+1:

PR #479, opened 2016-03-07
chore(scheduler): remove dead code. loc is used in v1 to get mem/cpu etc

helgi:
Application type information is acquired differently now


bacongobbler:
looks like `container_fullname` should get removed from that func, otherwise LGTM

helgi:
Done and done

PR #481, opened 2016-03-08
fix(serializers): validate tags properly as k8s labels

mboersma:
See https://github.com/kubernetes/kubernetes/blob/release-1.1/docs/user-guide/labels.md#syntax-and-character-set for an explanation of what is legal in a k8s label key and value. Previously Deis was disallowing the slash `/` character which could denote a DNS-style prefix for a label key.

See also deis/workflow-e2e#67, which this unblocks.

### DIFF
--- a/rootfs/api/serializers.py
+++ b/rootfs/api/serializers.py
@@ -14,8 +14,7 @@ from api import models
 PROCTYPE_MATCH = re.compile(r'^(?P<type>[a-z]+)')
 MEMLIMIT_MATCH = re.compile(r'^(?P<mem>[0-9]+(MB|KB|GB|[BKMG]))$', re.IGNORECASE)
 CPUSHARE_MATCH = re.compile(r'^(?P<cpu>[0-9.]+[m]{0,1})$')
-TAGKEY_MATCH = re.compile(r'^[a-z]+$')
-TAGVAL_MATCH = re.compile(r'^\w+$')
+TAGVAL_MATCH = re.compile(r'^(?:[a-zA-Z\d][-\.\w]{0,61})?[a-zA-Z\d]$')
 CONFIGKEY_MATCH = re.compile(r'^[a-z_]+[a-z0-9_]*$', re.IGNORECASE)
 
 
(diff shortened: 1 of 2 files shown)


arschles:
@mboersma code LGTM

helgi:
```
get nodes 172.17.8.100 --template={{.metadata.labels}}
map[kubernetes.io/hostname:172.17.8.100 ssd:true deis.com/fun:yes]
```

```
deis tags:set foo=bar -a oddest-zillions
Applying tags... Error:
400 Bad Request
detail: These tags do not match labels on kubernetes nodes: foo=bar
```

```
deis tags:set ssd=true -a oddest-zillions
Applying tags... done

=== oddest-zillions Tags
ssd     true
```

This is a weird one:
```
deis tags:set deis.com/fun=no -a oddest-zillions
Applying tags... Error:
400 Bad Request
detail: These tags do not match labels on kubernetes nodes: ssd=true, deis.com/fun=no
```
`ssd=true` is mentioned even if it applied fine earlier. It must be some sort of a compound logic gone wrong

Applying the right `deis.com/fun=yes` is all happy tho

```
deis tags:set deis.com/fun=yes -a oddest-zillions
Applying tags... done

=== oddest-zillions Tags
deis.com/fun     yes
ssd              true
```

Up to you if you want to look at the compound problem above or do a ticket on it and punt

helgi:
Forgot this one

```
deis tags:set deis.com-/fun=yes -a oddest-zillions
Applying tags... Error:
400 Bad Request
tags: Tag key prefixes must be DNS subdomains.
```

mboersma:
@helgi good catch. I'll take a look at fixing the validation message that includes all tags.

> `tags: Tag key prefixes must be DNS subdomains.`

It's complaining about `com-` ending with a non-alpha character which isn't allowed in a subdomain part. Should there be a better validation error? It's a bit tricky to be more specific here.

helgi:
Ah yeah, I was just including that last one as a proof your validation worked (= At best we should link to documentations more aggressively but we should just do that later as part of a bigger error -> doc improvements

mboersma:
Sounds good. I'll put an issue in for the `ssd=true` validation bug and merge this.

PR #487, opened 2016-03-09
feat(Makefile): enable immutable (git-based) Docker tag publishing

sgoings:


helgi:
Is this going to work for local development as well? docker-push now does multiple things

sgoings:
@helgi `docker-push` will definitely still work locally. It will push both the canary and the immutable-tagged image to whatever registry/image_prefix you have configured.

helgi:
Is there a way we can not have locally get 2x images of everything, since it is a local disk that fills up quickly. I've had to clean up my registry a few times just because development generates a lot of images

vdice:
This is nice but would be even prettier with a little... [makeup.](https://github.com/sgoings/makeup) :)

sgoings:
Followed up with @helgi via slack:

> How does it fill up your disk faster, @helgi? It’s just a tag that points to the same underlying image...
> right now local development is using that git sha based version...

As for @vdice's suggestion... I'm interested in getting this logic into repositories and then picking the elegant solution that plugs right in.

PR #493, opened 2016-03-11
feat(tests): use codecov.io for visual code coverage reporting

### LINKED ISSUE #442: Add code coverage tooling that augments coverage.py
https://codecov.io/#languages is the top contender for me. coveralls is something I've used before but right now I'm liking codecov.io better

helgi:
See https://codecov.io/github/deis/workflow

Closes #442 

Example of how PRs could start looking https://github.com/pyca/cryptography/pull/2805

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,6 +19,8 @@ install:
   - pip install -r rootfs/dev_requirements.txt
 script:
   - make test
+after_success:
+  - cd rootfs/ && codecov
 deploy:
   - provider: script
     # ensure client/doc builds aren't removed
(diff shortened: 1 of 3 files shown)


helgi:
For the curious there is a small toggle up in the right corner(ish) at https://codecov.io/github/deis/workflow/features/ to find badges (which I added), additional graphs we can embed and how the PR commits and status checks could work if we integrate it all

helgi:
Some problems going on in there it looks like. It's not finding the coverage file, which is odd

mboersma:
> how PRs could start looking

I noticed those coverage comments when looking at the Python cryptography package the other day. Very useful. Deis should definitely add this. Maybe for [dockerbuilder](https://github.com/deis/dockerbuilder) too.

mboersma:
@helgi I changed the line in .travis.yml to `cd rootfs/ && codecov` and it seems to have worked: https://travis-ci.org/mboersma/workflow

Issue #412, opened 2016-02-23
Write a new Chaos Scheduler


helgi:
The current one needs to be replaced since we have started moving more and more logic from the scheduling code and up into the relevant models

Joshua-Anderson:
@helgi I know you've been working on a lot of the scheduler tests in the last few sprints. Is this still relevant?

helgi:
Yes - We have not written this + relevant test cases

Cryptophobia:
This issue was moved to teamhephy/controller#68

Issue #60, opened 2015-12-09
add mock k8s tests


bacongobbler:
Having some library to inspect the requests sent to kubernetes would be helpful for adding unit tests to new features in the kubernetes client. That way we aren't spending all of our time manually testing these features.

[added by @helgi]:
Minimum coverage needs to be 85% per file / class in the scheduler to be considered good coverage

Following need work:

- [ ] `__init__.py`
- [ ] `Deployment`
- [ ] `HPA`
- [ ] `Pod`
- [ ] `RC`

https://codecov.io/gh/deis/controller/tree/master/rootfs/scheduler has the most up to date information

arschles:
:+1:. Such a library could be modeled after the Go on in Helm: https://github.com/helm/helm/tree/master/kubectl.

helgi:
Closed by #497

helgi:
Reopening this since I think we may need to inspect a bit more closely what we add to k8s. Right now we use a very basic mock but in many cases it would at least give us a decent idea of what we are applying based on various configs and otherwise.

Thoughts on how we should go about this?

Joshua-Anderson:
@helgi In the past few sprints I know you've worked on these mocks, are they sufficient enough to close this issue now?

helgi:
I haven't been happy enough with those - They prove certain interactions but not enough introspection on the built payload, et all. When we do more of that then I'd consider this done

helgi:
I updated the core description with some additional information.

One thing to note is functions like `manifest` in each resource we do not test directly but rather rely on `create` and `update` to call it with the right things. As such there is no introspection at that level on what is being sent... That could be a good addition overall, to compare to that directly since that's what is sent to Kubernetes... Thoughts?

helgi:
Going to close this - We have very good coverage, mostly 70% and higher. In many cases 85% will be hard to achieve due to the mocking not being smart enough to create more delays etc to exercise some of the wait code

PR #497, opened 2016-03-11
ref(*): mock the k8s API interactions and remove Container model

### LINKED ISSUE #29: backfill unit tests for k8s.py
This is crucial path code and could use some unit test :green_heart:.

helgi:
This change removes the Container model as it is redundant and changes around the MockScheduler to instead mock out actual k8s API calls and responses to get real testing of the scheduler code.

Chaos scheduler was ripped temporarily but will be readded in the future (see #412)

More logic needs to go into the mock handling to deal with pod transitions

Fixes #29, #60


vdice:
Closes #512

rootfs/api/models/app.py
```diff
@@ -534,23 +398,56 @@ def logs(self, log_lines=str(settings.LOG_LINES)):
...
-                                     release=self.release_set.latest(),
-                                     type='run',
-                                     num=c_num)
         # SECURITY: shell-escape user input
-        escaped_command = command.replace("'", "'\\''")
-        return c.run(escaped_command)
+        command = command.replace("'", "'\\''")
+
+        # if this is a procfile-based app, switch the entrypoint to slugrunner's default
+        # FIXME: remove slugrunner's hardcoded entrypoint
```
arschles: @helgi is there an issue in slugrunner for this?

helgi: Not sure, that's been there as a comment as long as I can remember

arschles: Ref https://github.com/deis/slugrunner/issues/23

rootfs/api/tests/test_build.py
```diff
@@ -114,14 +114,17 @@ def test_build_default_containers(self):
...
+        url = "/v2/apps/{app_id}/pods/cmd".format(**locals())
+        response = self.client.get(url, HTTP_AUTHORIZATION='token {}'.format(self.token))
         self.assertEqual(response.status_code, 200)
         self.assertEqual(len(response.data['results']), 1)
         container = response.data['results'][0]
         self.assertEqual(container['type'], 'cmd')
-        self.assertEqual(container['num'], 1)
+        self.assertEqual(container['release'], 'v2')
+        # pod name is auto generated so use regex
+        self.assertRegex(container['name'], app_id + '-v2-cmd-[a-z0-9]{5}')
```
arschles: worth pulling this regex out into a global?

helgi: Yeah, I'll do it after this PR tho since I have a few test additions i want to do that are not directly affecting this PR

rootfs/api/tests/test_pods.py
```diff
@@ -285,16 +282,21 @@ def test_container_errors(self):
...
-                               app=app,
-                               config=app.config_set.latest(),
-                               build=build)
+        Release.objects.create(
+            version=2,
+            owner=user,
+            app=app,
+            config=app.config_set.latest(),
+            build=build
+        )
```
arschles: does this return anything or raise an exception?

helgi: It returns the `release` object but since we don't use it `flake8` complains if I do `relase = Release.objects.create`

arschles: is that because that value is not being used? maybe there's something to check in it?

rootfs/api/tests/test_pods.py
```diff
@@ -454,48 +502,55 @@ def test_command_good(self):
...
-                                         app=app,
-                                         config=app.config_set.latest(),
-                                         build=build)
-        # create a container
-        c = Container.objects.create(owner=user,
-                                     app=app,
-                                     release=release,
-                                     type='web',
-                                     num=1)
+        Release.objects.create(
```
arschles: same here regarding checking the returned value

arschles: there are a few instances below also, but I won't comment on them for brevity's sake

rootfs/api/tests/test_pods.py
helgi: The value is not used, it's just to create a DB record so we have a release attached to the Build object

rootfs/scheduler/__init__.py
```diff
@@ -960,6 +956,7 @@ def _healthcheck(self, controller, path='/', port=8080, delay=30, timeout=1):
         }
 
         # Because it comes from a JSON template, need to hit the first key
+        # FIXME: assumption that container 0 is the app
```
arschles: I'm assuming this is old code that was copied, but can we get an issue for it?

helgi: https://github.com/deis/controller/issues/518

arschles: thanks

rootfs/scheduler/__init__.py
```diff
@@ -985,6 +979,8 @@ def _get_secret(self, namespace, name):
         if unhealthy(response.status_code):
             error(response, 'get Secret "{}" in Namespace "{}"', name, namespace)
 
+        # FIXME decode data - can it be done without affecting the response object too much???
```
arschles: same here RE an issue in GH for this

helgi: https://github.com/deis/controller/issues/519

arschles:
@helgi made a few "meta" comments, but code LGTM

rootfs/scheduler/__init__.py
```diff
@@ -481,59 +476,62 @@ def run(self, name, image, entrypoint, command, **kwargs):
...
+                    self._delete_pod(namespace, name)
                     return 0, log
+
                 if pod['status']['phase'] == 'Running':
                     if iteration > 28:
-                        duration = duration + 1
-            except:
+                        duration += 1
+            except KubeException:
                 break
```
mboersma: Should this log the exception?

helgi: Likely, @smothiki was using this as a flow mechanism tho I think. I just made the exception catching more explicit

rootfs/api/urls.py
```diff
@@ -29,22 +29,6 @@
         views.ReleaseViewSet.as_view({'post': 'rollback'})),
     url(r"^apps/(?P<id>{})/releases/?".format(settings.APP_URL_REGEX),
         views.ReleaseViewSet.as_view({'get': 'list'})),
-    # restart containers
```
mboersma: This change to the REST API implies an update to [the docs](https://github.com/deis/workflow/blob/master/src/reference-guide/controller-api/v2.0.md). I'll make an issue to track that: deis/workflow#93.

helgi: We also have https://github.com/deis/controller/issues/363 - should that be closed down in favour of the issue you created?

mboersma: They describe slightly different issues, so we should leave them both open for now IMHO.

mboersma:
I've seen failures from `make test-unit` on this PR. One example:
```python

FAIL: test_tags (api.tests.test_config.ConfigTest)
  File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/unittest/mock.py", line 1157, in patched
    return func(*args, **keywargs)
  File "/Users/matt/Projects/src/github.com/deis/controller/rootfs/api/tests/test_config.py", line 472, in test_tags
    self.assertEqual(response.status_code, 201)
AssertionError: 400 != 201

Ran 152 tests in 37.485s
```

vdice:
Fixes #522

bacongobbler:
The approach looks good to me, though the tests need to go green.

PR #508, opened 2016-03-15
fix(scheduler): create RC with scale 0 and then use scale function to have pod monitoring

helgi:


helgi:
Test Plan:

```
deis create --no-remote
deis pull deis/example-go --procfile 'worker: while true; do echo hi; sleep 3; done' -a <app>
deis scale worker=1 -a <app>
```

The scale should wait until the worker one is done scaling up. Before this PR it comes up in the returned list as `starting`

carmstrong:
Code LGTM but I didn't manually test.

PR #509, opened 2016-03-16
ref(apps): iterate over RC for desired pod count when doing an app restart

helgi:
Rationale here is that the RC already has the desired information and we may be iterating over many RCs.
Not using the pod count is helpful in case other operations happen to have interfered with things.

Another option is to use the structure info from the App model but this solutions seems more solid

Also moved some filtering into a shared function

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -177,15 +177,20 @@ class App(UuidAuditedModel):
                 version = "v{}".format(release.version)
                 kwargs['name'] = '{}-{}-{}'.format(kwargs['id'], version, kwargs['name'])
 
-            # Fetch the initial set of pods to work from
-            pods = self.list_pods(**kwargs)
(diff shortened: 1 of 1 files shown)


bacongobbler:
does this fix a specific bug for beta? I'm leaning against reviewing things that are non-critical to the beta milestone so we can get this release out the door.

helgi:
I haven't noticed a specific problem but in my mock testing it made things more stable in restart while the cache around pods was settling

Saw odd things since I was grabbing pod numbers and then running a delete. Quite a few times got the wrong desired number because cache was still being committed or something similar. This made that go away

arschles:
@helgi [associated tests](https://github.com/deis/controller/pull/497/files#diff-132a7f6400fa1f92ad4431c64959fd4d) that you pointed me helped me understand what's going on here. Thanks!

PR #513, opened 2016-03-16
fix(travis): update secure URL

bacongobbler:
The job name at https://ci.deis.io has been changed to
controller-master, so travis needs to be updated to push to that
job.


bacongobbler:
github was being incredibly stupid. Nothing a good ol' `git push origin --delete && git push -f` can't fix.

Issue #135, opened 2015-12-18
Make slugrunner image configurable


technosophos:
The image used by slugrunner should be configurable.

We need to make it possible for the slugrunner image to be configured by a cluster administrator so that any deploys use that slugrunner instead of the default Deis slugrunner.

arschles:
@technosophos should deis/builder have the same capability with slugbuilder?

technosophos:
Probably.

smothiki:
Can you be more specific about what configs should be tunable

technosophos:
I was initially imagining making `image` tunable for slugrunner. And @arschles is suggesting the same for slugbuilder. That was all.

technosophos:
I think we can safely demote this from Beta, since none of the Beta requirements specify or rely upon this behavior.

helgi:
This has not been fixed. The PR in question does slug and docker builder, not slug runner

https://github.com/deis/workflow/blob/master/rootfs/scheduler/__init__.py#L31 is still around

krancour:
> I think we can safely demote this from Beta, since none of the Beta requirements specify or rely upon this behavior.

I would argue for including this.  Once we transition to immutable tags on our images, we're not going to want to be figuring out a way to insert the latest tag for slugrunner and slugbuilder into controller and builder... we'll be wanting to insert it into env variables in the manifests for those components instead.

helgi:
Fixed by #515

PR #515, opened 2016-03-16
feat(slugrunner): make slugrunner configurable via env variables

smothiki:
https://github.com/deis/charts/pull/164


helgi:
Fixes #135

Issue #521, opened 2016-03-17
Swap "minio" for generic "object storage"


krancour:
See here: https://github.com/deis/controller/blob/745f7e1504ae6cb9374a94d77f37eabe48578788/rootfs/scheduler/__init__.py#L46-L66

It would probably be more proper to not use "minio" here and, instead, not assume or hint at the use of any particular object storage provider by using a more generic term like "object storage" instead.

A large benefit to this is cleaner charts.

cc @arschles @helgi

arschles:
This seems related to the following two issues, either of which may change the scope of this one or render it obsolete:

- https://github.com/deis/slugrunner/issues/18 will mean that we'll be using the new `deis-objectstorage` secret, which is a more generic way to specify the object storage endpoint info and credentials
- https://github.com/deis/controller/issues/182 will mean that we won't need to pass slugrunner pods an object storage secret at all. Instead, the controller would be able to pass the slugrunner a pre-signed URL to the object, and the slugrunner will be able to download the slug without any other credentials.

helgi:
Doesn't this duplicate #458

Issue #458, opened 2016-03-02
change secrets to access object storage credentials


smothiki:
In build pack apps template and in creating a secret a object store secret in the App namespace we are still using minio-user . Have to use deis-object-storage secret

arschles:
:+1:

Depends on https://github.com/deis/builder/issues/224

arschles:
does this depend on https://github.com/deis/slugrunner/issues/18?

Issue #97, opened 2015-12-16
[meta] all components should report their 2.0.x version string


mboersma:
In Deis v1.x each component has an `ENV DEIS_RELEASE v1.x` string in its Dockerfile. That was a workaround that may no longer be needed, and requires lots of clerical PRs just to update strings. In Deis v2.x it would be preferable for each component to print its version up-front, on startup.

arschles:
@mboersma should the versions be `2.0.0`?

mboersma:
In workflow we bumped the `deis` client to `2.0.0-dev`. But that's probably not specific enough at this point--it will be more useful to have `2.0.0-beta1` or something similar that contains the pre-release version. (And maybe the git SHA?) And since printing such a version up-front will require changes to components, it's not in scope for alpha.

arschles:
@mboersma I see. I opened a few PRs that I'll leave open for now, without an alpha milestone

helgi:
@mboersma Can we start opening up tickets for the components that are missing and fix up the version strings as needed?

helgi:
@mboersma says this isn't needed anymore

PR #523, opened 2016-03-17
ref(scheduler): fetch object storage credentials from API instead of local secret

helgi:
For now fetch the minio credentials from the API

ref #458
ref #521

Note: I can change it to to object-storage if someone points me to the doc what to use instead

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -968,14 +968,11 @@ class KubeHTTPClient(AbstractSchedulerClient):
     # http://kubernetes.io/v1.1/docs/api-reference/v1/definitions.html#_v1_secret
 
     def _create_minio_secret(self, namespace):
-        with open("/var/run/secrets/deis/minio/user/access-key-id", "rb") as the_file:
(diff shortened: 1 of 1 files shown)


smothiki:
is this going in for beta? object storage secret is not yet implemented in slugrunner

helgi:
This can go into beta if I just keep it as `minio-user` which is what slugrunner expects based on what you said

smothiki:
Also this requires a change in helm charts removing secret volume mount in controller rc

helgi:
https://github.com/deis/charts/pull/168

arschles:
Ref https://github.com/deis/workflow/issues/97

arschles:
@helgi has this been tested?

smothiki:
if this is going in we should hold off the current testing and wait for the new images

arschles:
Looks like it's already merged :)

helgi:
Already merged and master for it was already built https://travis-ci.org/deis/controller/builds/116716881

smothiki:
oops my bad

PR #525, opened 2016-03-17
feat(database): persist DB connection for 10 minutes

helgi:
https://docs.djangoproject.com/en/1.9/ref/databases/#persistent-connections

Maybe the volume to our DB is not enough but majority of our interactions depend on it. It does have a note about dev envs as well


bacongobbler:
related historical issue with CONN_MAX_AGE: https://github.com/deis/deis/issues/1619

bacongobbler:
disregard; that issue seems unrelated to this change. I just did a background check because I thought that CONN_MAX_AGE had bitten us in the past.

helgi:
I believe that was a problem due to `Threading` that was also accessing the database?

PR #535, opened 2016-03-17
feat(slugrunner): pass storage parameters to slugrunner

### LINKED ISSUE #458: change secrets to access object storage credentials

### LINKED ISSUE #521: Swap "minio" for generic "object storage"

kmala:
pass storage parameters to slugrunner

Fixes #458
Fixes #521

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -41,14 +41,14 @@ POD_BTEMPLATE = """\
             "value":"$image"
         },
         {
-            "name": "DOCKERIMAGE",
-            "value":"1"
+            "name": "BUILDER_STORAGE",
+            "value":"$storagetype"
         }
         ],
         "volumeMounts":[
         {
-            "name":"minio-user",
-            "mountPath":"/var/run/secrets/object/store",
+            "name":"objectstorage-keyfile",
+            "mountPath":"/var/run/secrets/deis/objectstore/creds",
             "readOnly":true
         }
         ]
@@ -56,9 +56,9 @@ POD_BTEMPLATE = """\
     ],
     "volumes":[
     {
-        "name":"minio-user",
+        "name":"objectstorage-keyfile",
         "secret":{
-        "secretName":"minio-user"
+        "secretName":"objectstorage-keyfile"
         }
     }
     ],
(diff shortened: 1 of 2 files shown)


arschles:
This requires https://github.com/deis/slugrunner/pull/24, right?

kmala:
Yes. this requires the slugrunner PR

smothiki:
Also dont merge this untill slugrunner PR gets merged

rootfs/scheduler/__init__.py
```diff
@@ -41,14 +41,14 @@
             "value":"$image"
         },
         {
-            "name": "DOCKERIMAGE",
```
helgi: This information is no longer required?

smothiki: Yes we have moved away from MC binary

helgi:
Does this PR closes https://github.com/deis/controller/issues/458 and https://github.com/deis/controller/issues/521

rootfs/scheduler/__init__.py
```diff
@@ -194,14 +194,22 @@
                 "value":"$appversion"
             },
             {
-                "name": "DOCKERIMAGE",
-                "value":"1"
+                "name": "BUILDER_STORAGE",
+                "value":"$storagetype"
+            },
+            {
+                "name": "DEIS_MINIO_SERVICE_HOST",
```
helgi: Are these two ENV vars not auto exposed to the service? Does that just happen within a given namespace?

smothiki:
>Does this PR closes #458 and #521

yes

rootfs/scheduler/__init__.py
kmala: yes

helgi: Fair enough, not super thrilled having more information in there but it's required so YOLO!

PR #543, opened 2016-03-21
Implements deis/deis#4792 : Add an auth:whoami endpoint

MaxenceAdnot:
I implemented an auth:whoami endpoint for the controller in response to https://github.com/deis/deis/issues/4792

`make test` : OK - coverage : 81.88%
`make deploy` : OK - `maxenceadnot/controller:git-7747b7c` or :canary (on docker.io)

**GET auth/whoami with an admin**
```
{
  "email": "admin@mailemail.me",
  "username": "admin",
  "first_name": "",
  "last_name": "",
  "is_superuser": true,
  "is_staff": true,
  "groups": [],
  "user_permissions": [],
  "last_login": "2016-03-21T00:46:41UTC",
  "date_joined": "2016-03-21T00:46:41UTC",
  "is_active": true
}
```

**GET auth/whoami with a standard user**
```
{
  "email": "test@mailemail.me",
  "username": "test",
  "first_name": "",
  "last_name": "",
  "is_superuser": false,
  "is_staff": false,
  "groups": [],
  "user_permissions": [],
  "last_login": "2016-03-21T00:46:50UTC",
  "date_joined": "2016-03-21T00:46:50UTC",
  "is_active": true
}
```

_PS: Please be indulgent, this is me very first PR ;)_


helgi:
That looks useful! Looking at the output, it may be returning too much information, especially since right now we do not use first / last name, and potentially returning some of the user information in a `whoami` endpoint is a little too much information.

I'd have to look at the permission endpoints but I think that information is covered in those places (permission side of it) tho I understand what the original ticket is going for. That level of detail seems more suited for a users and/or permissions endpoint that returns the relevant information.

The code is self is good and the rest is awesome as well :)

KevinBacas:
I'm currently building a Web UI for Deis and this is exaclty the kind of features that I'm missing.. I need to manually store get e dfrom the APIhaving any possibility to get it from the API. This is kind of annoying :)

Love this PR :smile: :+1:

helgi:
@KevinBacas What kind of information do you need that you are currently caching or getting from other endpoints?

KevinBacas:
Actually:
* Current user informations (Present in this PR)
* Is the token still valid ?
  * I'm using local storage to keep the user logged in but I can't even know if the token is still valid when the user reload the page
* Some API calls does not return any responses (e.g. Destroy application)
  * I have to check if the statusText equals 'No Content' and prevent the JSON parsing for few API calls
* In v2, the endpoint is missing for apps
  * If I deploy an app named 'test', the real URL is 'test.deis.com' for instance but the url returned by deis API is 'test'

e.g.:
```
deis domains --app=nginx
=== nginx Domains
nginx
```

helgi:
That's a lot of different things, it would be best if you opened up tickets for those with the infos you need in each item if you want us to have more back and forth.

For this PR, right now I am mostly worried about the amount of information returned. Such as the first / last name (as I mentioned) since we do not have any place that updates that, or the whole `is_superuser` as we are planning on soon revamping the whole permission system

As far as the API endpoints that do not return content, that's intended, those endpoints return a `HTTP 204` which means `No Content` so you could check the status code. https://www.koozai.com/blog/analytics/server-response-codes-and-what-they-mean/#204 - We may consider changing it from `204` to `200` if you give us good reasons :)

helgi:
I'm okay with this PR going in with the knowledge that this API behaviour may change in not too distant future, with a API revision following semantic versioning.

I'd want @bacongobbler or @mboersma to throw some eyes on this as well.

@KevinBacas any of your changes, including the `204` would be valuable as separate tickets to keep track of those

bacongobbler:
@KevinBacas in the future please feel free to open issues on all your concerns. To answer your questions and to expand on @helgi's response:

> Is the token still valid? I'm using local storage to keep the user logged in but I can't even know if the token is still valid when the user reload the page

An API token is valid forever (read: no expiry date) until a user runs `auth:regenerate`. We rely on Django REST Framework which does not expire tokens by default. We haven't changed that behavior. Pinging an endpoint like this PR's `/v2/auth/whoami` with the token should give you a yes or no answer if the token has been revoked by the server.

> Some API calls does not return any responses (e.g. Destroy application) 

As @helgi said, destroying an application will return a 204 NO CONTENT as described in [the docs](http://docs.deis.io/en/latest/reference/api-v1.7/#destroy-an-application). That is the expected behaviour. Check the response's status code in your code. :)

> In v2, the url is wrong in the domains API call

That is also expected. Since we cannot determine the platform URL ahead of time (we did before in v1, now we don't), any "bare" domains belong to the platform and any specific domains like `foo.example.com` will be explicitly routed. They're treated differently in the router.

See https://github.com/deis/router for more information, specifically CTRL + F for "router.deis.io/domains". :)

Hope this helps!

bacongobbler:
@helgi I haven't reviewed this yet but I feel like we should wait until after v2-beta1 has been shipped to add this new feature.

deis-admin:

Your Pull Request can earn you credit at DigitalOcean! See http://docs.deis.io/en/latest/contributing/hacking/#digitalocean-credit for details.

mboersma:
@MaxenceAdnot could you rebase this and address the existing comments? Thanks for helping with Deis Workflow!

MaxenceAdnot:
I've updated the PR to merge the latest changes from master. What do mean by "rebase this and address the existing comments" ?

bacongobbler:
@MaxenceAdnot @mboersma means by rebasing this branch against master (notice that this branch is out of date with master) using `git rebase master` rather than `git merge master`. It messes with the commit history because you'll get a whole bunch of "Merge branch 'master' into master" commits, and "merging" and "rebasing" are used in two different scenarios. See https://help.github.com/articles/about-git-rebase/

Looks like all the comments have been addressed in this PR. We will still need to implement deis/workflow#218 and deis/workflow-cli#54 before we merge this, however.

MaxenceAdnot:
Ok ok, i tried but the merges didn't cleaned up... I'm familiar with the rebase command. 
I'll give another try later.

MaxenceAdnot:
I rebased the branch (only 1 commit now) but it includes all the changes from master...

jchauncey:
moved to rc1

bacongobbler:
Would you mind re-PRing this change? It looks like the commit got messed up. If you have the commit sha of the original PR I would be happy to rebase it and create a new PR for you.

MaxenceAdnot:
I'll do that as soon as possible. Thank you for your help.

MaxenceAdnot:
I don't have much time to create a new PR so if you can make it, it will be very nice 👍 

Here are the original commits : 

commit 7747b7cbbc00641a0d10f1276c1c561d2457f168
Author: Maxence Adnot <contact@maxenceadnot.me>
Date:   Mon Mar 21 02:39:34 2016 +0100

    Added a unit test for auth/whoami

commit 87fd9c283e7627a507288b3a1720222fc872b079
Author: Maxence Adnot <contact@maxenceadnot.me>
Date:   Mon Mar 21 02:17:56 2016 +0100

    Added auth/whoami

bacongobbler:
I created #737. Closing!

PR #555, opened 2016-03-23
feat(api): disable adding wildcard certificates

bacongobbler:
port of deis/deis#4986


bacongobbler:
nvm, wildcard certificates are supported in v2.

PR #557, opened 2016-03-24
feat(apps): validate routeability through deis-router upon application creation

### LINKED ISSUE #551: Validate routeability through deis-router upon application creation.
Controller should only return from `create` operations until the application has been "discovered" by the router.

Controller should use the router svc address with app hostname to verify non-404 response.

When no `HEALTCHECK_URL` has been provided, controller should accept any non-404 response (including 502) as routeable.

Refs deis/workflow-e2e#119

helgi:
Hits the router IP via python-requests with a Host header (uses nip.io) to emulate a request to the application. Sends a Deis Controller UA to make debugging easier
HTTP mocking is required for this to work in tests and all tests now get peppered with mock_requests object, which they can use for other purposes as well if required

Depends on a change to the requests_mock python package for testing only, pulling that from git currently

Tests take 2x the time because of added timing jitter to emulate router being slow to respond at times

Fixes #551

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -3,15 +3,19 @@ import logging
 import random
 import re
 import requests
+from requests_toolbelt import user_agent
 import string
 import time
+from urllib.parse import urljoin
 
 from django.conf import settings
 from django.db import models
 from django.core.exceptions import ValidationError
 from jsonfield import JSONField
 
+from deis import __version__ as deis_version
 from api.models import UuidAuditedModel, log_event, AlreadyExists
+
 from api.utils import generate_app_name, app_build_type
 from api.models.release import Release
 from api.models.config import Config
@@ -304,7 +308,9 @@ class App(UuidAuditedModel):
                 'replicas': scale_types[scale_type],
                 'app_type': scale_type,
                 'build_type': build_type,
-                'healthcheck': release.config.healthcheck()
+                'healthcheck': release.config.healthcheck(),
+                # http://docs.deis.io/en/latest/using_deis/process-types/#web-vs-cmd-process-types
+                'routable': True if scale_type in ['web', 'cmd'] else False
             }
 
             command = self._get_command(scale_type)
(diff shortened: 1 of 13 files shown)


vdice:
Running tests w/ these changes it appears we can now re-enable a few more tests (specifically actions immediately post app deploy).  I'll create the correlating PR in workflow-e2e

rootfs/scheduler/__init__.py
```diff
@@ -312,6 +312,7 @@ def __init__(self):
             'User-Agent': user_agent('Deis Controller', deis_version)
         }
         # TODO: accessing the k8s api server by IP address rather than hostname avoids
+        # TODO look at https://toolbelt.readthedocs.org/en/latest/adapters.html#fingerprintadapter
```
arschles: nit: looks like this TODO interrupts the sentence started above

rootfs/api/models/app.py
```diff
@@ -385,6 +400,79 @@ def _default_structure(self, release):
...
+            return
+
+        app_type = kwargs.get('app_type')
+        self.log(
+            'Waiting for router to be ready to serve traffic to process type {}'.format(app_type),
+            level=logging.DEBUG
+        )
+
+        # Get the router host and append healthcheck path
+        url = 'http://{}:{}'.format(settings.ROUTER_HOST, settings.ROUTER_PORT)
```
arschles: I thought we were aiming to not have the controller talk directly to the router, and prefer to use the resources in k8s as a communication medium. Why the exception?

helgi: There is no way for us to verify the router has put the files into rotation and is able to respond on it without pinging the router like this.

We could possibly try to have the router try to self verify itself and update the SVC to say it's ready but @krancour would have to weigh in

krancour: > I thought we were aiming to not have the controller talk directly to the router

I think it's with respect to implementation-specifics that we don't want these two components communicating.  Since the controller is only making a request to the application, passing through the router in the process, much as any actual end-client would, I think this is ok.

Having said that, @arschles and I have traded a lot of ideas back and forth about how to make all the nodes of the router mesh more immediately consistent with one another, with the ability to report back to any interested party about the current state of its configuration... however, _that_ would involve some implementation-specifics, and I think @arschles and I agreed we'd hold off on trying to solve this difficult problem until post-stable.

For the time being, I think what @helgi has done the right thing.

PR #558, opened 2016-03-24
fix(Makefile): fix commit-msg task

Joshua-Anderson:
When contrib scripts were moved to _scripts this didn't get updated.

Now that we have many deis repositories however, it may more sense to remove this from here and instead move it to the documentation.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ clean: check-docker
 	docker rmi $(IMAGE)
 
 commit-hook:
-	cp contrib/util/commit-msg .git/hooks/commit-msg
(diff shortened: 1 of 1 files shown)


mboersma:
> it may more sense to remove this from here and instead move it to the documentation.

Yes, and/or add it to the [prototype-repo](https://github.com/deis/prototype-repo) and try to ensure each component has it.

Joshua-Anderson:
I think the prototype repo would be ideal, but whenever the script is updated it would require ~10 PRs and everybody would need to run `make commit-hook` for every repo, which might be unmaintainable.

I'd be happy to go whatever way people think makes the most sense.

PR #560, opened 2016-03-24
chore(confd): remove confd in favor env vars that can configure things

### LINKED ISSUE #358: Remove confd
Right now confd is used for a handful of things, mostly to write out a production configuration that is not done during a test run.

We need to figure out how to source the very few values that are left in confd and how we are going to load different configurations for production vs testing

helgi:
New Env Vars are:
* `REGISTRATION_MODE` which defaults to `enabled`
* `RESERVED_NAME` which defaults to `deis`
* `GUNICORN_WORKERS` which defaults to `not set`

Gunicorn reload script was left behind just to have an easy script to reload the process if that is desired instead of cycling the pod

Fixes #358


rootfs/deis/production.py
```diff
@@ -1,6 +1,4 @@
-import os
```
bacongobbler: I think `production_settings.py` is more fitting, or does that clash with `--settings deis.production`?

helgi: The calls would have to be changed to `--settings deis.production_settings` or we can do `deis.settings.production` if we make it a module

If you look at things in `production.py` tho then you see they are all secrets at this point and in k8s 1.2 we can mount them as ENV vars and it can all move into `settings.py` proper, killing off `production.py`

bacongobbler: yes, that would be a lovely world to live in if we didn't have to support k8s v1.1 clusters as well, but I agree that we can eventually turn this into a separate module. It's good for now, just curious about the name change. :)

arschles:
Removing my LGTM since this has a Kubernetes v1.2 requirement.

arschles:
Ignore my previous comment. I misunderstood the context of the aforementioned comment thread. @helgi helped me understand in an offline discussion that `production.py` still reads secrets files. In v1.2, we'll be able to remove it in favor of reading & reloading secrets in the same way as other configs (with env vars). Since that file still exists, and reads secrets from files, I'm re-adding my LGTM.

PR #564, opened 2016-03-24
fix(scheduler): apply healthcheck to the right application container instead of assuming

### LINKED ISSUE #518: Health Checks need to be added on the right container if there are more than one within the Pod
Currently in the code there is an assumption that the first container in a Pod is the application container when applying Health Checks.

That is a sound like for applications Deis has created but for the sake of sanity it would be easier to do a simple loop and name matching to grab the correct container.

helgi:
Right now there is an assumption that the first container in a Pod is the Deis application container. While not wrong it is better to find the right container and only apply to it

Fixes #518

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -917,7 +917,8 @@ class KubeHTTPClient(object):
 
     def _healthcheck(self, controller, path='/', port=8080, delay=30, timeout=1):
         # FIXME this logic ideally should live higher up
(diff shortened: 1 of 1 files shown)


krancour:
Aren't the healthchecks done over HTTP?  All containers in a pod share a common IP as if they are one logical machine, so this shouldn't matter... unless you are also exposing a different healthcheck option, like running some certain command.

helgi:
Right now it is HTTP but later it may not. It was an easy change to tack this on to the right container

arschles:
I like the defensiveness of this change :)

PR #566, opened 2016-03-25
fix(app): check if app name already exists in kubernetes during app creation

### LINKED ISSUE #316: Destroying a Deis cluster with a live app seems to cause deployments to the same app name to fail on a new cluster
Repro steps:

1. `helm install deis-dev`
2. `git clone example-go`
3. `cd example-go`
4. `deis register deis.$INGRESS_IP.xip.io`
5. `deis keys:add ~/.ssh/id_rsa.pub`
6. `deis create gotest`
7. `git push deis master`
8. `git commit -m “empty” —allow-empty && git push deis master`
9. Verify your app is up: `curl gotest.$INGRESS_IP.xip.io`
10. `helm uninstall deis-dev`
11. Notice the gotest namespace is still there
12. `helm install deis-dev`
13. `git remote remove deis`
14. Repeat steps 4-7
15. Verify that the app is still on the old v2 version: `curl gotest.$INGRESS_IP.xip.io`

Also, if you kubectl get pod —namespace=gotest, you’ll see the old v2 pod is still running

helgi:
Fixes #316

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -11,7 +11,7 @@ from django.db import models
 from django.core.exceptions import ValidationError
 from jsonfield import JSONField
 
-from api.models import UuidAuditedModel, log_event
+from api.models import UuidAuditedModel, log_event, AlreadyExists
 from api.utils import generate_app_name, app_build_type
 from api.models.release import Release
 from api.models.config import Config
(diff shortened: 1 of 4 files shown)


helgi:
Test plan for this one is to follow the steps @arschles outlined in #316 or simply create an app with the same name as an existing `k8s` namespace, that is not an application in your deis setup

helgi:
```
k get ns
NAME              LABELS          STATUS    AGE
deis              heritage=deis   Active    1d
kube-system       <none>          Active    11d
velcro-instinct   <none>          Active    11d
```

And then

```
deis create --no-remote velcro-instinct
Creating Application... Error:
409 Conflict
detail: velcro-instinct already exists as a namespace in this kuberenetes setup
```

PR #567, opened 2016-03-26
fix(boot): make wsgi aware of what settings file to use (production)

helgi:
boot script and friends worked just fine because the settings were passed to it but wsgi.py auto uses settings.py if nothing else is passed in

### DIFF
--- a/rootfs/api/views.py
+++ b/rootfs/api/views.py
@@ -38,6 +38,7 @@ class HealthCheckView(View):
                 c.execute("SELECT 0")
         except django.db.Error as e:
             logger.critical("Database health check failed")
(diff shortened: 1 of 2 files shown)


rootfs/api/views.py
```diff
@@ -38,6 +38,7 @@ def get(self, request):
                 c.execute("SELECT 0")
         except django.db.Error as e:
             logger.critical("Database health check failed")
+            # FIXME why is turning on DEBUG=true in env not outputting this error?
```
helgi: I made https://github.com/deis/controller/issues/568 for this. Had to change it to `critical` in my local tests

PR #573, opened 2016-03-26
fix(scheduler): only update application service definition once

helgi:
The problem here is that multiple process types will let the *last* process type win out. Instead let the first one win


aledbf:
Code LGTM

PR #575, opened 2016-03-28
feat(settings): use k8s 1.2 secrets as env vars to configure production settings

### LINKED ISSUE #574: Expose current secrets as ENV Vars to get rid of production.py
Now that 1.2 is the minimum k8s requirement it is possible to map the current secrets that the Controller needs for operations, as ENV Var instead of reading the file from the FS

https://github.com/kubernetes/kubernetes/issues/4710

helgi:
http://kubernetes.io/docs/user-guide/secrets/#using-secrets-as-environment-variables

Closes #574


bacongobbler:
my only concern is that we no longer work on k8s 1.1 but I guess it's already too late for that.

helgi:
1.1 isn't the goal anymore :)

PR #577, opened 2016-03-29
feat(k8s): match cert data structure to the TLS one for Ingress

### LINKED ISSUE #576: Change the certificate secret to match the Ingress TLS secret format
Supporting the [Ingress TLS secret format](http://kubernetes.io/docs/user-guide/ingress/#tls) makes us forward compatible

This is only about changing the actual keys inside the `data` portion. Router could support the current v2 format and the Ingress format at the same time with a simple `if not set` logic

/cc @aledbf @krancour

helgi:
http://kubernetes.io/docs/user-guide/ingress/#tls

Fixes #576 

/cc @aledbf @krancour 

Requires a router changes as per my description in #576


aledbf:
Code LGTM

PR #578, opened 2016-03-29
fix(scheduler): only update port on routable services

### LINKED ISSUE #571: bug: Incorrect type in service
In an app with multiple types (web and cron), the created service used cron as `type`.
(if the app defines multiple type, always uses the last one as `type` in the service)

helgi:
Depends on #557 
Fixes #571

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -1,3 +1,4 @@
+from collections import OrderedDict
 from datetime import datetime
 import logging
 import random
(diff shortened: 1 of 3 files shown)


smothiki:
why cant we just create service during first deploy

helgi:
We used to do that. Problem was that we kept re-creating the default application domain even if people deleted it, and domains require the service to be available at the time. 

It may be more doable to have it only happen once now that I have cleared out bunch of code, but I'd prefer to do that as a follow up to this PR if people feel strongly about it

PR #582, opened 2016-03-30
feat(permissions): add option to disable app creation for normal users.

Joshua-Anderson:
This adds a environmental variable toggle (`$DISABLE_APP_CREATION`) to restrict app creation to administrators only.

This my start in re-implementing some of the features in deis/deis#4265 and the proposal deis/deis#4150

Fixes deis/deis#4052

If this is the right approach, I'll start implementing some of the other feature flags in deis/deis#4150.

### DIFF
--- a/rootfs/api/permissions.py
+++ b/rootfs/api/permissions.py
@@ -19,6 +19,16 @@ def is_app_user(request, obj):
         return False
 
 
+def can_create_app(request):
+    """
+    Return `True` if user can create app, `False` otherwise.
+    """
+    if request.user.is_superuser:
+        return True
+    else:
+        return not settings.DISABLE_APP_CREATION
+
+
 class IsAnonymous(permissions.BasePermission):
     """
     View permission to allow anonymous users.
--- a/rootfs/api/tests/test_app.py
+++ b/rootfs/api/tests/test_app.py
@@ -10,6 +10,7 @@ import requests
 from django.contrib.auth.models import User
 from django.core.cache import cache
 from rest_framework.test import APITestCase
+from django.test.utils import override_settings
 from rest_framework.authtoken.models import Token
 
 from api.models import App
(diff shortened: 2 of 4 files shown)


Joshua-Anderson:
Is this the desired approach to accomplish this? Does anybody have any feedback on this PR?

mboersma:
@Joshua-Anderson this looks like a clean implementation of deis/deis#4052, and your timing is impeccable as (you know) v2 feature development is very much preferred. I appreciate the incremental approach toward the larger PR.

Apologies for the CI failure--we are working out some issues and I know you're not unfamiliar with that situation. :smiling_imp: But it will be so much better this time. Thanks!

Joshua-Anderson:
Rebased!

krancour:
I think this is a nice feature, but I think we need a holistic approach to re-implementing permissions.  This implementation gets the job done for the narrow use case of allowing _only_ admins to create apps, but it doesn't move us any closer to role-based access control.  What if I want "team leads" to be able to create apps?

Joshua-Anderson:
@krancour I talked about this in my proposal for controller permissions in deis/deis#4150.

I was thinking that deis would support customizing both default permissions and user specific permissions.

Default permissions would allow cluster administrators control how much access normal users get on the platform by default. This could be updated at any time and effect every user.

If you want to override the default platform permissions, such as for a team lead, you could explicitly grant them permission with `$ deis perms:create team-lead --apps`.

If you invite an untrusted user into the platform, you could also explicitly limit their permissions. `$ deis perms:delete foo --apps`

Implementing a PR with a complete permissions rewrite would be huge and very difficult to review. So, I thought it would make sense to make many small PRs like this one (which only implements default platform permissions for apps) to slowly add more and more permissions over time.

This PR at the moment does have a narrow use case, but if I couple it with a subsequent PR to allow user specific app permissions, it then becomes much more useful. Administrators could then restrict app creation and grant it only to specific members. Or, they could permit app creation and restrict certain members from creating apps.

krancour:
I understand what you're saying, and I think others should weigh in.  In my mind, the idea of "default" permissions makes sense, but the managemebt of those permissions should not differ from how permissions for any other role are managed.  i.e. "default" should be a role that every new user gets up front.  Admins should administer it as they would other roles.

Small PRs are nice, but there are times when a hollistic approach is called for, otherwise the UX that emerges can ends up feeling inconsistent, which is how I think this would make me feel.

Joshua-Anderson:
>  management of those permissions should not differ from how permissions for any other role are managed. i.e. "default" should be a role that every new user gets up front. Admins should administer it as they would other roles.

Maybe we should have a 'default' user in the database which could store the default permissions, so that you could do something like `$ deis perms:delete default --apps`? We'd need to figure out some way to prevent people from assigning apps, deleting, setting a password for, and so on to that default user though, which might get pretty hacky. The reason I went with an environmental variable is I couldn't think of a good way to save this in the database. Is there a better approach than creating a mock user?

> i.e. "default" should be a role that every new user gets up front.

I was thinking that administrators would expect that changing the default permissions would affect every user that has not had the permission explicitly set. That way changing for example, app creation permissions, wouldn't require manually updating every user that already exists.

krancour:
No... not a default user... that complicates things, because as you said, you'd have to figure out how to treat that user differently than the others.  What we need are _roles_... full fledge role-based access control.

Joshua-Anderson:
@krancour Do you have any suggestions for packages I should look into? We're currently using django-guardian for app permissions. It's pretty lightweight and allows you to assign permissions to an object. The problem is it doesn't seem to really support storing and updating default permission levels.

That's why I designed this with ENV variables for default permissions and planned to use django-guardian for per-user permissions.

krancour:
I cannot profess to know much about django.  The actual implementation of any role-based access control requires some careful design.  I think @helgi should be a part of any such conversations.

Joshua-Anderson:
I'm going to close this for now and possibly revisit it later.

PR #587, opened 2016-03-31
feat(scheduler): store env vars as secrets in k8s and map them in the RC

### LINKED ISSUE #299: [question]: is there a reason why the rc uses env and not secrets?
Something like one secret per version or the one secret per app that is updated in each version.

helgi:
This provides more security around sensitive information and lets k8s admins put more controls in place on who can update what

Went with a secret per release instead of a secret per app that we keep updating. Thoughts on that?

Closes #299

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -377,9 +377,15 @@ class KubeHTTPClient(object):
                 new_rc["metadata"]["name"], desired)
             )
 
-            # Remove old release
+            # Remove new release
             self._scale_rc(namespace, new_rc["metadata"]["name"], 0)
             self._delete_rc(namespace, new_rc["metadata"]["name"])
+            # remove secret that contains env vars for the release
+            try:
+                secret_name = "{}-{}-env".format(namespace, kwargs.get('version'))
+                self._delete_secret(namespace, secret_name)
+            except KubeHTTPException:
+                pass
 
             # Bring back old release if available
             if old_rc:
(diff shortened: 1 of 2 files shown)


aledbf:
Code LGTM

aledbf:
@helgi it's possible to add a warning or a reference indicating that the secret keys are overwritten in each  release? (so if an user change the value of the secret is aware of this behavior)

rootfs/scheduler/__init__.py
```diff
@@ -390,6 +397,13 @@ def deploy(self, namespace, name, image, command, **kwargs):
         # New release is live and kicking. Clean up old release
         if old_rc:
             self._delete_rc(namespace, old_rc["metadata"]["name"])
+            # remove secret that contains env vars for the release
+            secret_name = "{}-{}-env".format(namespace, old_rc['metadata']['labels']['version'])
+            try:
+                self._get_secret(namespace, secret_name)
+                self._delete_secret(namespace, secret_name)
```
slack: Do we need to get, then delete? Or can we just fire the delete straight up?

rootfs/scheduler/__init__.py
```diff
@@ -528,7 +542,7 @@ def run(self, namespace, name, image, entrypoint, command, **kwargs):
             error(response, 'create Pod in Namespace "{}"', namespace)
 
         data = ''
-        duration = 30
+        duration = 120
```
slack: Unrelated?

helgi: Unrelated

rootfs/scheduler/__init__.py
helgi: I can fire off the delete straight up but felt it was more appropriate to hit the k8s API cache before trying to delete from it, probably an overkill but doesn't hurt either

slack:
LGTM. Also seems that folks who upgrade controller would have their app configs "upgraded" to k8s secrets on the next deploy?

We need to get used to thinking through those compatibility scenarios!

helgi:
Next deploy (be it via pull, config:set or any operation that causes a deploy) you will get auto upgraded

slack:
Yep, and looks like an exception during secret delete is `caught -> pass`

helgi:
That's intentional :)

slack:
Good jerb.

PR #591, opened 2016-04-01
feat(config): setting one healthcheck value now results in all defaults for health to be saved

### LINKED ISSUE #549: Health Checks and how they are displayed to the end user
Ever since https://github.com/deis/controller/commit/81a5c37fb4d3b35a022fe8eac4bbd43113f75ae7#diff-ffb566d124a367aa8d7bd107a49d8313 health checks have not been enabled unless `HEALTHCHECK_URL` is set but even then we are not auto backfilling other other default values back into the database.

This means if I set a healh check url and then list out the configs for my app I will not be able to see all the other default values, so as a developer I have to know what defaults Deis Workflow set. That's not very useful

One option is to start auto saving the default values into the DB if the developer is not using them. That way a `config:list` will show all the required values, as I would expect to see from a Web UI. The problem with that approach is that if we ever decide to change the defaults that means anything we have auto saved to the database is "dirty" and a migration for that is not the simplest.

The other option is to return all the defaults via `config:list` or similar that are not set via `config:set` but not save them to the database, only if the URL is set minimum.

Thoughts?

/cc @slack

helgi:
This both helps the end user to see what the other healthcheck values are and saves the state to the DB, that way platform upgrades will not suddenly change defaults from undernearth users

```
controller  (health>) $ deis config:set HEALTHCHECK_URL=/health -a excess-ziggurat
Creating config... done

=== excess-ziggurat Config
HEALTHCHECK_INITIAL_DELAY      50
HEALTHCHECK_PORT               5000
HEALTHCHECK_TIMEOUT            50
HEALTHCHECK_URL                /health
Helgi                          Foo
Helgi_Hi                       Foo

controller  (health>) $ deis config:set HEALTHCHECK_PORT=5001 -a excess-ziggurat
Creating config... done

=== excess-ziggurat Config
HEALTHCHECK_INITIAL_DELAY      50
HEALTHCHECK_PORT               5001
HEALTHCHECK_TIMEOUT            50
HEALTHCHECK_URL                /health
Helgi                          Foo
Helgi_Hi                       Foo
```
Closes #549

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -942,7 +942,7 @@ class KubeHTTPClient(object):
 
         # add in healtchecks
         if kwargs.get('healthcheck'):
-            template = self._healthcheck(template, **kwargs['healthcheck'])
+            template = self._healthcheck(template, kwargs['routable'], **kwargs['healthcheck'])
 
         url = self._api("/namespaces/{}/replicationcontrollers", namespace)
         resp = self.session.post(url, json=template)
@@ -986,10 +986,8 @@ class KubeHTTPClient(object):
 
         return response
 
-    def _healthcheck(self, controller, path='/', port=8080, delay=30, timeout=1):
-        # FIXME this logic ideally should live higher up
-        app_type = controller['spec']['selector']['type']
-        if app_type not in ['web', 'cmd']:
+    def _healthcheck(self, controller, routable=False, path='/', port=5000, delay=30, timeout=5):  # noqa
+        if not routable:
             return controller
 
         namespace = controller['spec']['selector']['app']
@@ -1029,6 +1027,7 @@ class KubeHTTPClient(object):
         }
 
         # Update only the application container with the health check
+        app_type = controller['spec']['selector']['type']
         container_name = '{}-{}'.format(namespace, app_type)
         containers = controller['spec']['template']['spec']['containers']
         for container in containers:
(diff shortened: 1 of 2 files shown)


rootfs/scheduler/__init__.py
```diff
@@ -954,10 +954,8 @@ def _delete_rc(self, namespace, name):
 
         return response
 
-    def _healthcheck(self, controller, path='/', port=8080, delay=30, timeout=1):
-        # FIXME this logic ideally should live higher up
-        app_type = controller['spec']['selector']['type']
-        if app_type not in ['web', 'cmd']:
+    def _healthcheck(self, controller, routable = False, path = '/', port = 5000, delay = 30, timeout = 5):  # noqa
```
mboersma: PEP8 says
> Don't use spaces around the = sign when used to indicate a keyword argument or a default parameter value.

mboersma:
Does `targetPort: 8080` in the `SERVICE_TEMPLATE` also need to change to 5000?

helgi:
Not really related. Services just need a default value that we anyway overwrite. Think I'd rather do it in a separate pull if I want to unify ports

rootfs/scheduler/__init__.py
helgi: Yeah but for some reason my local flake didn't like it. Will have a look again why that happened

helgi: Fixed

jchauncey:
do we have any e2e tests for this?

helgi:
Thought we did and then I looked at the health check test and it looks like that does nothing at all

Issue #132, opened 2015-12-18
`deis delete` prompts me to delete kubernetes (hahahahaha)


technosophos:
When I ran `deis delete` on an app named `humble-traverse`, I got this prompt:

```
⇒  deis destroy
 !    WARNING: Potentially Destructive Action
 !    This command will destroy the application: kubernetes
 !    To proceed, type "kubernetes" or re-run this command with --confirm=kubernetes

> kubernetes
Destroying kubernetes...
Error:
404 NOT FOUND
<h1>Not Found</h1><p>The requested URL /v1/apps/kubernetes/ was not found on this server.</p>
```

technosophos:
Or, I'm an idiot and ran this in the wrong terminal window. :facepunch:

PR #592, opened 2016-04-01
fix(controller): use django HttpResponse for logs

bacongobbler:
Django Rest Framework will wrap the string with quotes because
of the JSONRenderer. Django's HttpResponse class handles the
application logs as we expect it to.

this is a port of deis/deis#5004

related PRs: deis/workflow-cli#33 and deis/workflow#132

### DIFF
--- a/rootfs/api/tests/test_app.py
+++ b/rootfs/api/tests/test_app.py
@@ -108,7 +108,10 @@ class AppTest(APITestCase):
         # test logs - unanticipated status code from deis-logger
         mock_response.status_code = 400
         response = self.client.get(url)
-        self.assertContains(response, "Error accessing logs for {}".format(app_id), status_code=500)
+        self.assertContains(
+            response,
+            "Error accessing logs for {}".format(app_id),
+            status_code=500)
 
         # test logs - success accessing deis-logger
         mock_response.status_code = 200
@@ -119,7 +122,10 @@ class AppTest(APITestCase):
         # test logs - HTTP request error while accessing deis-logger
         mock_get.side_effect = requests.exceptions.RequestException('Boom!')
         response = self.client.get(url)
-        self.assertContains(response, "Error accessing logs for {}".format(app_id), status_code=500)
+        self.assertContains(
+            response,
+            "Error accessing logs for {}".format(app_id),
+            status_code=500)
 
         # TODO: test run needs an initial build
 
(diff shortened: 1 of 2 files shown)


bacongobbler:
throwing on a "needs manual testing" because I've tested this fix on v1 but not on v2.

helgi:
There is no way to make DRF do what we want?

bacongobbler:
Not optimally. You'd have to create a third party PlainTextRenderer class, but then you have to figure out a way to get it to only apply to the `logs` function in the AppView.

bacongobbler:
AFAIK DRF only allows renderers to apply to the entire class or to a module function but not to a specific function in a class.

helgi:
Ah yeah I see now, http://www.django-rest-framework.org/api-guide/renderers/#example why on earth wouldn't they just include that

However applying a single render is easy `@renderer_classes((PlainTextRenderer))` decorator on the given functions

bacongobbler:
@helgi any clue on how to compare a byte string to a regular string? Looks like more python 3 snarf again: https://travis-ci.org/deis/controller/builds/122652731#L346

helgi:
Then you are converting it wrong, a `str()` is not enough. You can also try `assertIn` or `assertContains`

helgi:
`response.content.decode(encoding='UTF-8')` would be how you convert it to a string

helgi:
`str(bytesThing, encoding='UTF-8')` might... Can't recall

bacongobbler:
From https://docs.djangoproject.com/es/1.9/topics/python3/#httprequest-and-httpresponse-objects:

> According to PEP 3333:

> - headers are always str objects,
> - input and output streams are always bytes objects.

> Specifically, `HttpResponse.content` contains `bytes`, which may become an issue if you compare it with a `str` in your tests. The preferred solution is to rely on `assertContains()` and `assertNotContains()`. These methods accept a response and a unicode string as arguments.

helgi:
I've def used `assertContains` a few times in our tests to get past a `b'foo'` problem :) Lets use that for consistency but I'd recommend adding the `status_code` to the `contains` call so it is a "single" failure instead of 2 assertions like they are now

bacongobbler:
Jenkins really doesn't like me today. Those test failures are `ps` test failures as well as router issues, so it's unrelated to these changes.

helgi:
I've been seeing those as well on other PRs...

PR #594, opened 2016-04-03
fix(views): prevent overlapping config:set operations

mboersma:
Uses a file lock to ensure that multiple gunicorn processes can be coordinated in the container. If a `deis config:set` is in progress, another client attempting `config:set` will try to obtain the lock for 5 seconds, then return an error:
```
$ deis config:set POWERED_BY=Second
Creating config... Error: 
409 CONFLICT
error: Config changes already in progress.
[Errno 11] Resource temporarily unavailable
```

See also deis/deis#5008.

### DIFF
--- a/rootfs/api/tests/test_config.py
+++ b/rootfs/api/tests/test_config.py
@@ -5,6 +5,8 @@ Unit tests for the Deis api app.
 Run the tests with "./manage.py test api"
 """
 import json
+import os.path
+import tempfile
 
 from django.contrib.auth.models import User
 from django.core.cache import cache
@@ -15,6 +17,7 @@ from api.models import App, Config
 
 from api.tests import adapter, mock_port, DeisTransactionTestCase
 import requests_mock
+from simpleflock import SimpleFlock
 
 
 @requests_mock.Mocker(real_http=True, adapter=adapter)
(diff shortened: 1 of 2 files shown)


helgi:
Must admit I'm not a big fan of pulling in SimpleLock into more places but I suppose that's hard to prevent without a better operations system going on ;)

My main concern is we are adding logic into the views, logic that should live in the model probably? lock in the model and have the view understand the error so it can throw the 409

mboersma:
>  logic that should live in the model probably

That's a good point. This began as a naive port of deis/deis#4758, but could be improved. I'll take a look at re-homing that code.

helgi:
@mboersma awesome :) just wanted to make use try to use views less for "real" logic because it makes testing easier and reduces the cognitive load when trying to grok the code - We def have places where we kinda ... cheated :P Not that Django is super strict on the separation ;)

vdice:
@mboersma is this PR still intended for the `v2.0-rc1` milestone?

bacongobbler:
LGTM, though we should probably punt this to v2.2 and add docs in the API reference guide, eh?

bacongobbler:
Actually docs shouldn't be necessarily required for this, since it's just a new error code and we don't document the errors returned by the server. Ping @helgi for reviews

helgi:
The deployment pr prevents parallel updates, but only for deployments. Does supersed this PR somewhat

bacongobbler:
Personally I'd still want to merge this even if #854 supercedes this for the unit test alone so we can verify that.

helgi:
Then i would rather see those tests in their own PR. Seems wrong to me to merge flock things we will rip out within days

mboersma:
This isn't relevant now for default deployment-based apps, and we aren't excited about using more `flock` hacks, so I'm closing it.

PR #134, opened 2015-12-18
fix(k8s): pull image from private registry and inspect image

bacongobbler:
Instead of assuming the container is running on the same host as the controller is running on, we pull the image down and inspect its ports. Surpisingly the API endpoints are the exact same from image and container config, so it wasn't that hard to fix.

closes #107

PR #600, opened 2016-04-04
feat(health checks): implement new kubernetes 1.2 health check features

### LINKED ISSUE #251: Expose new healthcheck probe params as soon as they are available in k8s

helgi:
Added the following new functionalities

* `HEALTHCHECK_PERIOD_SECONDS` - How often (in seconds) to perform the probe
* `HEALTHCHECK_SUCCESS_THRESHOLD` - How many probe runs need to be done after failure to consider it a success again
* `HEALTHCHECK_FAILURE_THRESHOLD` - How many times a probe should fail before considered being down

These defaults are all copied from kubernetes but may have to be adjusted due to our high default timeout, or we bring the default timeout down

Closes #251
Depends on #591 being merged first

### DIFF
--- a/rootfs/api/models/config.py
+++ b/rootfs/api/models/config.py
@@ -38,8 +38,19 @@ class Config(UuidAuditedModel):
         timeout = int(self.values.get('HEALTHCHECK_TIMEOUT', 50))
         delay = int(self.values.get('HEALTHCHECK_INITIAL_DELAY', 50))
         port = int(self.values.get('HEALTHCHECK_PORT', 5000))
+        period_seconds = int(self.values.get('HEALTHCHECK_PERIOD_SECONDS', 10))
+        success_threshold = int(self.values.get('HEALTHCHECK_SUCCESS_THRESHOLD', 1))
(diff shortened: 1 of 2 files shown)


slack:
Can you drop a `deis/workflow` issue to track adding these to the docs?

helgi:
Yeah, I have https://github.com/deis/controller/issues/570 but realised just now these are not ENV Vars :P  I need to turn Health Checks into a section of the config instead of general configuration

helgi:
Created https://github.com/deis/workflow/issues/134 to track documentation needs

helgi:
Wrote https://github.com/deis/workflow/pull/135 for the documentation

arschles:
This lgtm.

In case it helps anyone else, I asked @helgi for some additional information on what's going on here, so adding that info here in case it helps anyone else:

this PR adds the above env variables that are settable via `config:set`, which tell the controller how to configure app pods w.r.t. health check information. There is also a plan in #603 to move these config values to a top level config element, similar to what we do for tags.

Issue #410, opened 2016-02-23
A failed release that doesn't end up k8s pods means no further releases can be done on app


helgi:
If for any reason a release fails in such a way a pod for the release is not kicked off it means that subsequent releases can not be done.

In my example I had a v3, did some workflow changes, attempted a release for v4, it bombed due to my changes but then on the next try we used v5 instead of v4 and that confuses all business logic we have

helgi:
I lied ... but you do get a release that is non-functional, db things all is good, k8s is oblivious

helgi:
The error handling logic is much improved around all of this. Going to close this until we start seeing this crop up again

Issue #320, opened 2016-01-29
Creating tags and failing still creates a release record


helgi:
That's a problem for future releases, applying tags, configs or anything else. Relates to #319 in many ways

helgi:
Not been able to reproduce this, closing for now

Issue #319, opened 2016-01-29
Deploy fails due to other reasons than the RC/k8s, RC gets created. DB kills its record


helgi:
This is more apparent during development due to bugs in code

```
detail: failed to get ReplicationController "aerial-icehouse-v2-cmd" in Namespace "aerial-icehouse": 404 Not Found
{'reason': 'NotFound', 'message': 'replicationControllers "aerial-icehouse-v2-cmd" not found', 'status': 'Failure', 'metadata': {}, 'kind': 'Status', 'apiVersion': 'v1', 'details': {'kind': 'replicationControllers', 'name': 'aerial-icehouse-v2-cmd'}, 'code': 404}
```

What now???

helgi:
The error handling logic is much improved around all of this. Going to close this until we start seeing this crop up again

Issue #187, opened 2015-12-22
rolling back sometimes throws a 500, but still creates a new version


arschles:
given the following releases:

```console
ENG000656:workflow aaronschlesinger$ deis releases:list -a app2
=== app2 Releases
v12	2015-12-22T21:26:57UTC	arschles rolled back to v10
v11	2015-12-22T21:25:53UTC	arschles rolled back to v3
v10	2015-12-22T21:25:46UTC	arschles rolled back to v4
v9	2015-12-22T21:25:32UTC	arschles rolled back to v5
v8	2015-12-22T21:25:25UTC	arschles rolled back to v4
v7	2015-12-22T21:25:14UTC	arschles rolled back to v5
v6	2015-12-22T21:24:50UTC	arschles rolled back to v1
v5	2015-12-22T21:24:17UTC	arschles rolled back to v3
v4	2015-12-22T21:24:13UTC	arschles rolled back to v1
v3	2015-12-22T21:23:48UTC	arschles rolled back to v2
v2	2015-12-22T21:22:02UTC	arschles deployed deis/example-go
v1	2015-12-22T21:20:31UTC	arschles created initial release
```

I haven't been able to reliably reproduce which rollbacks return 500, but neither the latest nor 2nd to latest version have thrown errors. When I do get an error, however, a new release is created:

```console
ENG000656:workflow aaronschlesinger$ deis releases:rollback v11 -a app2
Rolling back to v11... done, v13
ENG000656:workflow aaronschlesinger$ deis releases:rollback v10 -a app2
Rolling back to v10... Error: 
500 INTERNAL SERVER ERROR
<h1>Server Error (500)</h1>
ENG000656:workflow aaronschlesinger$ deis releases:list -a app2
=== app2 Releases
v14	2015-12-22T21:28:49UTC	arschles rolled back to v10
v13	2015-12-22T21:28:31UTC	arschles rolled back to v11
v12	2015-12-22T21:26:57UTC	arschles rolled back to v10
v11	2015-12-22T21:25:53UTC	arschles rolled back to v3
v10	2015-12-22T21:25:46UTC	arschles rolled back to v4
v9	2015-12-22T21:25:32UTC	arschles rolled back to v5
v8	2015-12-22T21:25:25UTC	arschles rolled back to v4
v7	2015-12-22T21:25:14UTC	arschles rolled back to v5
v6	2015-12-22T21:24:50UTC	arschles rolled back to v1
[...]

PR #604, opened 2016-04-05
fix(deploy): catch Exception instead of RuntimeError

helgi:
In the deploy function Exception is caught, problem logged and then re-raised. Best to catch Exception as well to clean up the release object around it

Potentially affects the following tickets:
#410, #319, #187, #320

### DIFF
--- a/rootfs/api/models/build.py
+++ b/rootfs/api/models/build.py
@@ -43,7 +43,7 @@ class Build(UuidAuditedModel):
         try:
             self.app.deploy(user, new_release)
             return new_release
-        except RuntimeError:
+        except Exception:
             if 'new_release' in locals():
                 new_release.delete()
             raise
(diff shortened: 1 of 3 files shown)


slack:
Not familiar with the python exception hierarchy... Is this too high up the stack?

helgi:
IMO no, we are already catching pure Exception in `deploy` itself and logging. This just kinda keeps with that.

I can make it more narrow if we harmonized the exceptions across Deis to extend from DeisException and k8s and our own error handling comes from that, but then we never account for Django / Python problems. However in `deploy` and co, instead of just doing `raise` I could re-raise as a diff kind

https://docs.python.org/3/library/exceptions.html#exception-hierarchy is the doc where RuntimeError lives vs all the other ones, https://docs.python.org/3/library/exceptions.html#RuntimeError explains what is was about

If you look at v1 then fleet and the internal Container Model used to use that exception, which we do not use anywhere anymore

PR #608, opened 2016-04-07
fix(deploy): use the passed in scale of a proc type for deploy

### LINKED ISSUE #607: A new release of an app that has no older RC behind will be treated like a new application and scaled to 1
The scenario is that a user has an application with 50+ releases behind it, with a proc scaled (one of many, such as `cron`) to `0` or `10` yet there are no RCs available for any of the previous releases then actions like `config:set` on the application will be treated as if it was a new application deploy (the first real deploy) and `controller` will scale the RC to `1` instead of the `scale` value saved in the App model

The logic behind this is difficult to grok no doubt but this is a scenario that can happen when the user has deleted the old RC by hand *or* has recovered a deis installation via a DB restore onto a new k8s cluster

helgi:
Before a scale of 1 was hard coded if an old RC could not be found, instead of that use information from the structure in the App model which already defaults to 1 on first deploy

Fixes #607

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -346,9 +346,11 @@ class KubeHTTPClient(object):
         new_rc = self._create_rc(namespace, name, image, command, **kwargs)
 
         # Get the desired number to scale to
-        desired = 1
         if old_rc:
             desired = int(old_rc["spec"]["replicas"])
+        else:
+            desired = kwargs['replicas']
+            logger.debug('No prior RC could be found for {}-{}'.format(namespace, app_type))
 
         try:
             count = 1
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -297,7 +297,7 @@ class App(UuidAuditedModel):
     def _scale_pods(self, scale_types):
         release = self.release_set.latest()
         build_type = app_build_type(release)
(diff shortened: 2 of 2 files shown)


aledbf:
Code LGTM

rootfs/scheduler/__init__.py
```diff
@@ -346,9 +346,11 @@ def deploy(self, namespace, name, image, command, **kwargs):
         new_rc = self._create_rc(namespace, name, image, command, **kwargs)
 
         # Get the desired number to scale to
-        desired = 1
         if old_rc:
             desired = int(old_rc["spec"]["replicas"])
+        else:
+            desired = kwargs['replicas']
+            logger.debug('No prior RC could be found for {}-{}'.format(namespace, app_type))
```
arschles: nit: maybe add to this log statement to indicate that it's gonna use `desired` for the new RC's replica value?

arschles:
@helgi code LGTM. added one nit, but not blocking for this PR IMO

rootfs/scheduler/__init__.py
helgi: There is ton of logging around `desired` and `current` a few lines down so it feels redundant

arschles: sgtm then

PR #613, opened 2016-04-07
chore(tests): increase coverage of certs, test bad keys

helgi:


rootfs/api/tests/test_certificate.py
```diff
@@ -120,3 +121,28 @@ def test_delete_certificate(self):
...
+        response = self.client.post(
+            self.url,
+            {
+                'name': 'random-test-cert',
+                'certificate': 'i am bad data',
+                'key': 'i am bad data as well'
+            }
+        )
+        self.assertEqual(response.status_code, 400)
+        # match partial since right now the rest is pyopenssl errors
```
bacongobbler: :+1:

bacongobbler: we could also run a test like so (this is off my brain):

```
with self.assertRaises(pyopenssl.Exception):
    cert = Certificate(...)
    cert.save()
```

helgi: We wrap the exception in `ValidationError` and SupsciousOperation` depending on which part of the system it happens in so we couldn't catch `pyoenssl.Exception`, right now anyway :|

PR #615, opened 2016-04-08
feat(release): re-use image source if Build is called with an image that is already in the deis registry

### LINKED ISSUE #614: Builds with full path image reference
In certain cases a user may have a test/qa setup of Deis, using the Deis Registry as a backing where they `git push` or `deis pull` to as appropriate, creating a copy in the internal Registry.

When they have verified that that everything is okay they will do `deis pull deis.registry.internal/application-name:v1990339` or similar, specifically referencing the internal registry ** and** a tag.

This means that the **source** and **target** registry destination are the same.

This operation right now will make our docker client `pull / tag / push` internally. That creates a lot of traffic and disk utilisation. Also there is a lock to prevent concurrent operations on the registry (due to a docker bug)

Instead of tagging the image then instead Deis could detect this scenario and return the image passed into the build object (via `Release::image`) + prevent `publish_release` in our docker client code base.

It may seem confusing to not have a image tagged per release in the registry but since users rarely are looking at the registry *and* a `Build` object exists per `Release` object then the confusion should be minimal. And we do not embed last mile configuration on the images anymore.

helgi:
Closes #614

### DIFF
--- a/rootfs/api/models/release.py
+++ b/rootfs/api/models/release.py
@@ -1,3 +1,5 @@
+import logging
+
 from django.conf import settings
 from django.db import models
 
@@ -5,6 +7,8 @@ from registry import publish_release
 from api.utils import dict_diff
 
 from api.models import UuidAuditedModel, log_event
+
+logger = logging.getLogger(__name__)
 
 
 class Release(UuidAuditedModel):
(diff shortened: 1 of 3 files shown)


aledbf:
Code LGTM

aledbf:
This improves dramatically the speed of deployments.
My workflow is:
- One app is used by one or more devs (multiple `git push`) as common ground (integration)
- A different app is used for QA/CD tests
- N apps (one per customers)

The ratio is N deploys in dev (lets say 100), less than half for QA but N*customer.
Only the first one uses `git push` the others use `deis pull`.

Until this PR any attempt to deploy more than one app using `deis pull` was serial (deis controller uses a lock file for docker pulls) just to pull, tag and push the same image just with a different name.
A side effect of this was the CPU used by docker (more than 80% during simultaneous `deis pull`).
Average time for this was ~3 minutes per pull.

After this a deploy of 5 simultaneous apps using `deis pull` take less than 20 seconds.

*Note:* this PR helps only if the docker registry used in `deis pull`is the same one used by the controller.

Thanks @helgi 👍

mboersma:
I tested this by using `deis pull` on an image that controller logs showed was already in the local repository:
```console
$ deis pull 10.3.240.116:80/hungry-sandwich:v2 -a hungry-sandwich
Creating build... done
```
Works as advertised and is much faster.

PR #623, opened 2016-04-10
chore(.travis.yml): update test PostgreSQL to 9.4

mboersma:
This is closer to parity with v9.4.6 currently in use by the [deis/postgres](https://github.com/deis/postgres) component.


mboersma:
This PR passed [Travis CI](https://travis-ci.org/deis/controller/builds/122144934), the stage relevant to this change.

PR #626, opened 2016-04-11
fix(views): use Exception instead of RuntimeError in view around rollbacks

helgi:
Same as the change around `deploy`

### DIFF
--- a/rootfs/api/views.py
+++ b/rootfs/api/views.py
@@ -18,6 +18,7 @@ from rest_framework.authtoken.models import Token
 
 from api import authentication, models, permissions, serializers, viewsets
 from api.models import AlreadyExists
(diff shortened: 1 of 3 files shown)


rootfs/api/views.py
```diff
@@ -152,7 +152,7 @@ def create(self, request, *args, **kwargs):
         try:
             return super(BaseDeisViewSet, self).create(request, *args, **kwargs)
         # If the scheduler oopsie'd
-        except RuntimeError as e:
+        except Exception as e:
```
helgi: Debating if I should rather import `KubeException` here and use that if we are specifically wanting `scheduler` problems but I have a feeling this is more of a catch all scenario, which can include Django and other ones?

PR #627, opened 2016-04-12
fix(release): attempt to delete stray RCs when Release DB model does a delete

helgi:
# Summary of Changes

This will reduce the occurance of stray RCs when the Release record is nuked, due to whatever problems may arise

I believe this will help with #319 and possibly others

# Testing Instructions

Somewhat hard to test this one since it is not possible to trigger a Release model delete without causing havoc in other places, which we also try to prevent

### DIFF
--- a/rootfs/api/models/release.py
+++ b/rootfs/api/models/release.py
@@ -6,8 +6,8 @@ from rest_framework.serializers import ValidationError
 
 from registry import publish_release, RegistryException
 from api.utils import dict_diff
-
 from api.models import UuidAuditedModel, log_event
+from scheduler import KubeHTTPException
 
 logger = logging.getLogger(__name__)
 
@@ -95,7 +95,7 @@ class Release(UuidAuditedModel):
             source_image.startswith(settings.REGISTRY_HOST) or
             source_image.startswith(settings.REGISTRY_URL)
         ):
-            logger.debug('{} already exists in the target registry. Using this image for release {} of app {}'.format(source_image, self.version, self.app))  # noqa
+            log_event(self.app, '{} already exists in the target registry. Using this image for release {} of app {}'.format(source_image, self.version, self.app))  # noqa
             return
 
         # add tag if it was not provided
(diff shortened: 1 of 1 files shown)


rootfs/api/models/release.py
```diff
@@ -95,7 +95,7 @@ def publish(self, source_version='latest'):
             source_image.startswith(settings.REGISTRY_HOST) or
             source_image.startswith(settings.REGISTRY_URL)
         ):
-            logger.debug('{} already exists in the target registry. Using this image for release {} of app {}'.format(source_image, self.version, self.app))  # noqa
+            log_event(self.app, '{} already exists in the target registry. Using this image for release {} of app {}'.format(source_image, self.version, self.app))  # noqa
```
arschles: is this related to the intent of the PR? ok if it is, just wondering for context

rootfs/api/models/release.py
```diff
@@ -148,6 +148,27 @@ def rollback(self, user, version):
...
+                self._scheduler._delete_rc(self.app.id, controller['metadata']['name'])
+        except KubeHTTPException as e:
+            # 404 means they were already cleaned up
+            if e.status_code is not 404:
+                # Another problem came up
+                message = 'Could not to cleanup RCs for release {}'.format(self.version)
+                log_event(self.app, message)
+                logger.warning(message + ' - ' + str(e))
+
+        super(Release, self).delete(*args, **kwargs)
```
arschles: would it make sense to call this in a `finally` block, so it's called no matter what?

rootfs/api/models/release.py
helgi: Not directly related beyond I realised I was relaying information in the wrong context and made this match what I did in the delete exception

rootfs/api/models/release.py
helgi: The codebase does not have many (any?) `finally` but it feels like a good practice to do since we want to delete no matter what, added it in

rootfs/api/models/release.py
arschles: sgtm

PR #629, opened 2016-04-12
ref(scheduler): simplify _get_old_rc in scheduler. Use existing functions and filtering

helgi:


rootfs/scheduler/__init__.py
```diff
@@ -768,27 +768,15 @@ def _delete_namespace(self, namespace):
...
-                break
-        if exists:
-            return prev_rc
+        labels = {
+            'app': namespace,
+            'type': app_type
+        }
+        controllers = self._get_rcs(namespace, labels=labels).json()
+        if len(controllers['items']) == 0:
+            return False
```
bacongobbler: Wouldn't we want to return `None` instead?

helgi: Before we returned `0`, we do a `if old_rc` in `deploy` so `False` seems more correct than `None` but really amounts to the same given how we use it

bacongobbler: pretty much. Unless we're specifically checking for `None` I don't think there's much of a difference.

PR #630, opened 2016-04-12
fix(app): deploy applications on Controller start to create RCs and Pods

### LINKED ISSUE #541: controller not recovering dead processes on restart
```
><> k --namespace=rb get po
NAME              READY     STATUS    RESTARTS   AGE
rb-v2-web-85cnj   1/1       Running   0          7m
rb-v4-web-7m9yf   1/1       Running   0          1m
rb-v4-web-fo8g5   1/1       Running   0          1m
><> k delete namespace rb
namespace "rb" deleted
><> kd get po
NAME                          READY     STATUS    RESTARTS   AGE
deis-builder-jvfvk            1/1       Running   3          33m
deis-controller-vf9uf         1/1       Running   5          33m
deis-database-8zzpy           1/1       Running   0          33m
deis-minio-mvnx9              1/1       Running   0          33m
deis-registry-ghba9           1/1       Running   6          33m
deis-router-svv9y             1/1       Running   1          33m
deis-workflow-manager-cqv06   1/1       Running   0          33m
><> kd delete po deis-controller-vf9uf
pod "deis-controller-vf9uf" deleted
><> k get namespaces
NAME          LABELS          STATUS    AGE
default       <none>          Active    40m
deis          heritage=deis   Active    34m
kube-system   <none>          Active    40m
rb            <none>          Active    8s
><> k --namespace=rb get po
NAME      READY     STATUS    RESTARTS   AGE
><> k --namespace=rb get rc
CONTROLLER   CONTAINER(S)   IMAGE(S)   SELECTOR   REPLICAS   AGE
```

So it looks like the namespace was re-created, but not the replication controller.

helgi:
# Summary of Changes

During Controller start all available applications are deployed with their latest release information. Facilitates proper recovery of an application when pulling information from DB only, on a fresh k8s cluster

Fixes #541

# Testing Instructions

Following test steps can be taken:

```
deis create --no-remote rb
Creating Application... done, created rb
remote available at ssh://git@deis.micro-kube.io:2222/rb.git

deis pull deis/example-go -a rb
Creating build... done
micro-kube  (1.2.2) $ k --namespace=rb get po
NAME              READY     STATUS    RESTARTS   AGE
rb-v2-cmd-2iwb2   1/1       Running   0          11s

k delete ns rb
namespace "rb" deleted

k get po
NAME                          READY     STATUS    RESTARTS   AGE
deis-builder-lp5gf            1/1       Running   1          8m
deis-controller-yq1h3         1/1       Running   0          4m
deis-database-se0kn           1/1       Running   0          8m
deis-logger-64vof             1/1       Running   0          8m
deis-logger-fluentd-5jprr     1/1       Running   0          8m
deis-minio-a64yg              1/1       Running   0          8m
deis-registry-j0axt           1/1       Running   1          8m
deis-router-p5sx1             1/1       Running   0          8m
deis-workflow-manager-trd2i   1/1       Running   0          8m

k delete po deis-controller-yq1h3
pod "deis-controller-yq1h3" deleted
micro-kube  (1.2.2) $ k get ns
NAME          STATUS    AGE
default       Active    10m
deis          Active    9m
kube-system   Active    10m
rb            Active    12s

 k --namespace=rb get po
NAME              READY     STATUS    RESTARTS   AGE
rb-v2-cmd-0yzlm   1/1       Running   0          20s
```

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -329,7 +329,7 @@ class App(UuidAuditedModel):
                 log_event(self, err, logging.ERROR)
                 raise
 
-    def deploy(self, user, release):
+    def deploy(self, release):
         """Deploy a new release to this application"""
         if release.build is None:
             raise EnvironmentError('No build associated with this release')
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -336,14 +336,21 @@ class KubeHTTPClient(object):
         session.verify = False
         self.session = session
 
-    def deploy(self, namespace, name, image, command, **kwargs):
+    def deploy(self, namespace, name, image, command, **kwargs):  # noqa
         logger.debug('deploy {}, img {}, params {}, cmd "{}"'.format(name, image, kwargs, command))
         app_type = kwargs.get('app_type')
         routable = kwargs.get('routable', False)
 
         # Fetch old RC and create the new one for a release
         old_rc = self._get_old_rc(namespace, app_type)
-        new_rc = self._create_rc(namespace, name, image, command, **kwargs)
+
+        # If an RC already exists then stop processing of the deploy
+        try:
+            self._get_rc(namespace, name)
+            logger.debug('RC {} already exists under Namespace {}. Stopping deploy'.format(name, namespace))  # noqa
+            return
+        except KubeHTTPException:
+            new_rc = self._create_rc(namespace, name, image, command, **kwargs)
 
         # Get the desired number to scale to
         if old_rc:
--- a/rootfs/api/management/commands/load_db_state_to_k8s.py
+++ b/rootfs/api/management/commands/load_db_state_to_k8s.py
@@ -9,8 +9,8 @@ class Command(BaseCommand):
     to k8s.
     """
     def handle(self, *args, **options):
-        """Publishes Deis platform state from the database to etcd."""
-        print("Publishing DB state to k8s...")
+        """Publishes Deis platform state from the database to kubernetes."""
+        print("Publishing DB state to kubernetes...")
         for model in (Key, App, Domain, Certificate, Config):
             for obj in model.objects.all():
                 obj.save()
@@ -21,4 +21,10 @@ class Command(BaseCommand):
                 domain = get_object_or_404(Domain, domain=domain)
                 cert.attach_in_kubernetes(domain)
 
-        print("Done Publishing DB state to k8s.")
+        # deploy applications
+        print("Deploying available applications")
+        for application in App.objects.all():
+            rel = application.release_set.latest()
+            application.deploy(rel)
+
+        print("Done Publishing DB state to kubernetes.")
--- a/rootfs/api/models/build.py
+++ b/rootfs/api/models/build.py
@@ -41,7 +41,7 @@ class Build(UuidAuditedModel):
         )
 
         try:
-            self.app.deploy(user, new_release)
+            self.app.deploy(new_release)
             return new_release
         except Exception:
             if 'new_release' in locals():
--- a/rootfs/api/models/release.py
+++ b/rootfs/api/models/release.py
@@ -141,7 +141,7 @@ class Release(UuidAuditedModel):
 
         try:
             if self.build is not None:
-                self.app.deploy(user, new_release)
+                self.app.deploy(new_release)
             return new_release
         except Exception:
             if 'new_release' in locals():
--- a/rootfs/api/views.py
+++ b/rootfs/api/views.py
@@ -304,7 +304,7 @@ class ConfigViewSet(ReleasableViewSet):
         try:
             # It's possible to set config values before a build
             if self.release.build is not None:
-                config.app.deploy(self.request.user, self.release)
+                config.app.deploy(self.release)
         except Exception:
             self.release.delete()
             raise


rootfs/api/models/app.py
```diff
@@ -329,7 +329,7 @@ def _scale_pods(self, scale_types):
                 log_event(self, err, logging.ERROR)
                 raise
 
-    def deploy(self, user, release):
+    def deploy(self, release):
```
mboersma: This didn't use the `user` arg? Crufty goodness.

(If this and the below changes are an unrelated cleanup, they should ideally be a separate commit and a separate PR though.)

mboersma:
I rebuilt and redeployed controller with this PR, but it went into `CrashLoopBackoff` hell:
```python
Publishing DB state to kubernetes...
Deploying available applications
  File "/app/manage.py", line 12, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/lib/python3.5/site-packages/django/core/management/__init__.py", line 353, in execute_from_command_line
    utility.execute()
  File "/usr/lib/python3.5/site-packages/django/core/management/__init__.py", line 345, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/lib/python3.5/site-packages/django/core/management/base.py", line 348, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/usr/lib/python3.5/site-packages/django/core/management/base.py", line 399, in execute
    output = self.handle(*args, **options)
  File "/app/api/management/commands/load_db_state_to_k8s.py", line 21, in handle
    application.deploy(rel)
  File "/app/api/models/app.py", line 373, in deploy
    **kwargs
  File "/app/scheduler/__init__.py", line 347, in deploy
    new_rc = self._create_rc(namespace, name, image, command, **kwargs)
  File "/app/scheduler/__init__.py", line 956, in _create_rc
    error(resp, 'create ReplicationController "{}" in Namespace "{}"', name, namespace)
  File "/app/scheduler/__init__.py", line 307, in error
    raise KubeHTTPException(errmsg, response=response)
scheduler.KubeHTTPException: failed to create ReplicationController "rococo-woodshed-v2-web" in Namespace "rococo-woodshed": 409 Conflict
{'apiVersion': 'v1', 'message': 'replicationcontrollers "rococo-woodshed-v2-web" already exists', 'reason': 'AlreadyExists', 'details': {'name': 'rococo-woodshed-v2-web', 'kind': 'replicationcontrollers'}, 'kind': 'Status', 'status': 'Failure', 'code': 409, 'metadata': {}}
```

helgi:
ah damn... I did not try the scenario where the RC was still there... I think RC is the only one that is not "relaxed" as NS / SVC... Hmmm I'll have to think on this a little to decide how I want to resolve this

I guess the big question is, should we bubble up the 409 if you try to deploy an RC that already exists? e.g. if I have v15 deployed and I programmatically do `deploy v15` then should we just `noop` because all the resources are there or error out?

rootfs/api/models/app.py
helgi: They can be a separate PR. It was mostly that I didn't want to add in a fake user or use a random user when calling deploy from the management script

helgi:
I have updated the code so that now if an RC exists for the release then we log that as debug information and stop the deploy operation.

I did not dive into trying to update the controller since `_create_rc` contains a lot of business logic while currently `_update_rc` is a very dumb passthrough. I could take on the work if people want to do that.

Benefit of doing the update is that the latest release would get any new RC attributes and changes we would have been releasing, if the Workflow Controller restart was done due to an upgrade. This could aid with upgrade scenarios but is more work.

Downside of doing the update is that we are touching the RC objects beyond what the user may expect us to.

Of course the user could also get the latest and greatest by just doing a normal build (new builds always get the latest of all).

Thoughts on which approach we want to take?

slack:
IMO, leaving an existing RC in place is fine for now. Controller being able to boot is top priority!

helgi:
Alright, if we are not going to update the RC then the PR is ready for more testing and review :)

rootfs/scheduler/__init__.py
```diff
@@ -961,7 +968,7 @@ def _create_rc(self, namespace, name, image, command, **kwargs):  # noqa
 
             time.sleep(1)
 
-        return resp.json()
+        return response
```
bacongobbler: is there a particular reason why we don't want to return the json object here? Miind making this a separate commit with justification, at least?

helgi: I can move it out into a separate PR - I did this as a part of a previous iteration of this PR

rootfs/scheduler/__init__.py
```diff
@@ -337,14 +337,21 @@ def __init__(self):
...
         # Fetch old RC and create the new one for a release
         old_rc = self._get_old_rc(namespace, app_type)
-        new_rc = self._create_rc(namespace, name, image, command, **kwargs)
+
+        # If an RC already exists then stop processing of the deploy
+        try:
+            self._get_rc(namespace, name).json()
+            logger.debug('RC {} already exists under Namespace {} - Stopping deploy'.format(name, namespace))  # noqa
+            return
+        except KubeHTTPException:
```
bacongobbler: Couldn't this also return a KubeHTTPException if the server returns an error? I guess in this case then we'd be making another call to the server which *should* return the same error but I wonder if there's a better way we can handle this.

helgi: Yeah, this is a blanket catch all, at best I can do `404` catching but it seemed excessive since the next call (create) will catch any server issues

mboersma:
This code looks good, but I want to re-test. I'll do so ASAP tomorrow morning.

mboersma:
I repeated the same scenario above and behavior is much improved:
```console
DEBUG create absurd-yardwork
DEBUG deploy absurd-yardwork-v2-web, img home/absurd-yardwork:git-de2a15d6/push/slug.tgz, params {'memory': {}, 'cpu': {}, 'envs': {}, 'routable': True, 'replicas': 1, 'version': 'v2', 'build_type': 'buildpack', 'tags': {}, 'healthcheck': {}, 'app_type': 'web'}, cmd "start web"
DEBUG RC absurd-yardwork-v2-web already exists under Namespace absurd-yardwork - Stopping deploy
Publishing DB state to kubernetes...
Deploying available applications
Done Publishing DB state to kubernetes.
deis-controller running...
```

bacongobbler:
let's hold off on merging this until beta2 has been released

arschles:
I'm punting this to beta3

PR #635, opened 2016-04-14
fix(scheduler): copy object-storage creds during RC creation instead …

helgi:


smothiki:
what difference does it make ? creation during RC phase

smothiki:
got it no worries

helgi:
Prior to that I had added a check where we only added the object keys when it was the right build type, however `create` actually gets no information passed to it and actually does not have the build information at the time anyway so it was a dud to begin with

mboersma:
I tested this and it does avoid the `Error syncing pod, skipping: secrets "objectstorage-keyfile" not found` problem.

PR #636, opened 2016-04-14
fix(secrets): supply python-requests with the changed response differnetly in get_secret

helgi:


mboersma:
I tested this locally and it resolves the objectstorage secret issue that was stopping buildpack apps when testing beta2.

PR #637, opened 2016-04-15
fix(models): make dates consistent, output Z instead of UTC for ISO8601 compatibility

### LINKED ISSUE #625: GET /v2/apps/go/pods output timedate is inconsistent
the pods output returns a Z timestamp:

```
><> curl -s http://deis.10.245.1.3.xip.io/v2/apps/go/pods -H 'Authorization: token 61dec82c88cc373353c5cf7dda1888cd163e18dd' | jpretty
{
    "count": 1,
    "results": [
        {
            "name": "go-v2-cmd-e7dej",
            "release": "v2",
            "started": "2016-04-11T20:52:37Z",
            "state": "up",
            "type": "cmd"
        }
    ]
}
```

Whereas the rest of the API uses `UTC` as the timezone (which is [wrong according to ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#UTC), but consistent):

```
><> curl -s http://deis.10.245.1.3.xip.io/v2/apps/go -H 'Authorization: token 61dec82c88cc373353c5cf7dda1888cd163e18dd' | jpretty
{
    "created": "2016-04-11T20:52:19UTC",
    "id": "go",
    "owner": "bacongobbler",
    "structure": {
        "cmd": 1
    },
    "updated": "2016-04-11T20:52:37UTC",
    "uuid": "1c2b1fdd-7f18-4ba1-b291-d9dc8d809229"
}
```

helgi:
# Summary of Changes

Changes all output format from using UTC to using Z, which is ISO8601 compatible

Fixes #625 

# Testing Instructions

See #625 for testing steps - Should see Z instead of UTC on started / updated

### DIFF
--- a/rootfs/api/tests/test_certificate.py
+++ b/rootfs/api/tests/test_certificate.py
@@ -95,7 +95,7 @@ class CertificateTest(APITestCase):
 
         expected = {
             'common_name': 'autotest.example.com',
-            'expires': '2016-03-05T17:14:27UTC',
+            'expires': '2016-03-05T17:14:27Z',
             'fingerprint': '37:24:D8:EB:DC:A4:2C:DA:88:55:C5:19:71:D3:9B:43:BA:AC:3A:CE:33:8E:07:52:1C:51:01:A0:97:43:C9:4D',  # noqa
             'san': [],
             'domains': [],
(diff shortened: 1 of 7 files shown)


helgi:
https://github.com/deis/workflow-e2e/pull/173 is the e2e change, this won't go green, and vice versa, until both are merged :P

Issue #168, opened 2015-12-21
`deis tag` should add nodeSelector to pod template


technosophos:
I guess that our plan for `deis tag` is to have it create node selectors as documented here: http://kubernetes.io/v1.1/docs/user-guide/node-selection/README.html

rimusz:
GKE does not really support proper nodes labelling yet, labels are gone then the node gets upgraded.
So we have to take this in to mind for `deis tag`

krancour:
> labels are gone then the node gets upgraded

Didn't understand that part.  Does this mean it's nit a graceful degradation?

rimusz:
automatic node upgrade on GKE is deleting the node and creating the new one

krancour:
I think we should still do this.  The fact that one particular managed k8s offering (even if it's the premiere one) has trouble with node labeling probably shouldn't stop us.  We could note in the documentation that mileage from this feature may vary with k8s distribution.

I've experimented with labeling nodes on my kube-aws cluster and it works great.

PR #640, opened 2016-04-15
feat(scheduler): allow setting gracefulTerminationPeriod and account for in terminations

### LINKED ISSUE #620: Proposal feat(k8s): allow custom terminationGracePeriodSeconds

### LINKED ISSUE #621: Proposal ref(controller): do not wait 120s for termination
The default `terminationGracePeriodSeconds` is 30 seconds.This means that after this time the pod is killed. During a deploy the controller waits for the removal of the pod. 
This makes no sense because some pods can stay in the state of termination (for several reasons) and the controller spent up to 120s per pod in an operation out of the scope of deis.

### LINKED ISSUE #631: deis ps:restart specific pod fails in e2e
Restarting seems to work, but running a `ps:list` afterwards is missing a pod.

See the workaround in https://github.com/deis/workflow-e2e/pull/164, which should get pulled out once a fix has been identified/merged.

helgi:
# Summary of Changes

Allows users to set how long k8s allows pods to stay alive after a SIGTERM before they send SIGKILL - Defaults to 30 (like k8s) but lets users set a lower value for faster deploys if the apps can handle it.

Reduces the pod termination wait to the new setting and uses the existance of "deletionTimestamp" as an indicator if a pod is Terminating, also checks if the pod is past the graceful period and hides it from the pod list via ps:list

Fixes #620
Fixes #621
Fixes #631

This change fixes up items marked as pending due to `ps:restart` on a single pod. See #631

# Testing Instructions

This is one slightly hard to test beyond marking the `ps:restart` as not Pending anymore. Most of this code is dealing with Pods that get stuck in Terminating state, see https://github.com/kubernetes/kubernetes/search?q=terminating&type=Issues for the never ending problem that seems to be around in k8s around that

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -582,9 +582,28 @@ class App(UuidAuditedModel):
                 if p['metadata']['labels']['type'] == 'run':
                     continue
 
+                state = self._scheduler.resolve_state(p).name
+
+                # follows kubelete convention - these are hidden unless show-all is set
+                if state in ['down', 'crashed']:
+                    continue
+
+                # hide pods that are past their graceful termination
+                # https://github.com/kubernetes/kubernetes/blob/release-1.2/docs/devel/api-conventions.md#metadata
+                # http://kubernetes.io/docs/user-guide/pods/#termination-of-pods
+                if 'deletionTimestamp' in p['metadata']:
+                    deletion = datetime.strptime(
+                        p['metadata']['deletionTimestamp'],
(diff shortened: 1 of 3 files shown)


bacongobbler:
Would you mind filing a ticket to add documentation for this? I imagine users would want to know how to use this.

helgi:
I will - originally I did not want this to be the final UX

helgi:
Created https://github.com/deis/workflow/issues/168

PR #644, opened 2016-04-18
feat(slugrunner): add readiness check for the slugrunner pod

kmala:
# Summary of Changes
Add readiness probe which check whether the downloded slug has started running

# Testing Instructions

1. Create a Deis Cluster
2. deploy a build pack app and simultaneously do watch on kubectl get pods in another window
3. Once the app is deployed you should see a state where the pod is in running state but doesn't show it as ready.

# Pull Request Hygiene TODOs

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -935,9 +935,11 @@ class KubeHTTPClient(object):
 
         self._set_environment(container, namespace, **kwargs)
 
-        # add in healtchecks
+        # add in healthchecks
         if kwargs.get('healthcheck'):
             template = self._healthcheck(template, kwargs['routable'], **kwargs['healthcheck'])
+        elif kwargs.get('build_type') == "buildpack":
+            template = self._buildpack_readiness_probe(template)
 
         url = self._api("/namespaces/{}/replicationcontrollers", namespace)
         resp = self.session.post(url, json=template)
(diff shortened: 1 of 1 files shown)


rootfs/scheduler/__init__.py
```diff
@@ -1049,6 +1051,44 @@ def _healthcheck(self, controller, routable=False, path='/', port=5000, delay=30
 
         return controller
 
+    def _readiness_probe(self, controller, delay=30, timeout=5, period_seconds=5,
```
helgi: Maybe we should name it differently and add in more description about it being there *only* when a healthcheck is not provided and only when it is a build pack application

rootfs/scheduler/__init__.py
```diff
@@ -1049,6 +1051,44 @@ def _healthcheck(self, controller, routable=False, path='/', port=5000, delay=30
...
+                # an exec probe
+                'exec': {
+                    "command": [
+                        "bash",
+                        "-c",
+                        "[[ '$(ps -p 1 -o args)' != *'bash /runner/init'* ]]"
+                    ]
+                },
+                # length of time to wait for a pod to initialize
+                # after pod startup, before applying health checking
```
helgi: You can probably just hardcode these values instead of having it configurable via the method signature - Do you think we will / should ever made it usable anywhere else?

rootfs/scheduler/__init__.py
```diff
@@ -942,6 +942,8 @@ def _create_rc(self, namespace, name, image, command, **kwargs):  # noqa
         # add in healtchecks
         if kwargs.get('healthcheck'):
             template = self._healthcheck(template, kwargs['routable'], **kwargs['healthcheck'])
+        elif kwargs.get('build_type') == "buildpack":
+            template = self._builpack_readiness_probe(template)
```
mboersma: Could you fix the func name to be `_buildpack_readiness_probe` here and below?

rootfs/scheduler/__init__.py
```diff
@@ -1049,6 +1051,53 @@ def _healthcheck(self, controller, routable=False, path='/', port=5000, delay=30
 
         return controller
 
+    '''
+    Applies exec readiness probe to the slugrunner container.
+    http://kubernetes.io/docs/user-guide/pod-states/#container-probes
+
+    /runner/init is the entry point of the slugrunner.
+    https://github.com/deis/slugrunner/blob/master/rootfs/Dockerfile#L24
```
mboersma: It might be better to leave such specific GitHub links out of the actual commit. They work now, but may point to entirely different lines of code soon.

rootfs/scheduler/__init__.py
```diff
@@ -942,6 +942,8 @@ def _create_rc(self, namespace, name, image, command, **kwargs):  # noqa
         # add in healtchecks
```
krancour: I know this wasn't introduced in this PR, but let's just fix this spelling mistake. Since we _are_ working on healthchecks here, I think it's allowable.

kmala: done.

rootfs/scheduler/__init__.py
kmala: updated

krancour:
Btw... where did all the junk in the OP come from?  It looks like some of our contributing docs.  How did it get in here?

PR #646, opened 2016-04-20
fix(deploy): should deploy successfullly without web type in the procfile

kmala:
# Summary of Changes

Deploy should be successful even when web doesn't exist as proc type in the procfile for a buildpack app.

# Testing Instructions

1. Create a Deis Cluster
2. deploy a buildpack app with proc type as test or anything other than web
3. app should get successfully deployed

# Pull Request Hygiene TODOs

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -393,13 +393,18 @@ class App(UuidAuditedModel):
         if not release.build.sha:
             structure = {'cmd': 1}
 
-        # if a dockerfile exists without a procfile, assume docker workflow
-        elif release.build.dockerfile and not release.build.procfile:
+        elif release.build.procfile and 'web' in release.build.procfile:
+            structure = {'web': 1}
+
+        # if a dockerfile, assume docker workflow
+        elif release.build.dockerfile:
             structure = {'cmd': 1}
 
-        # if a procfile exists without a web entry, assume docker workflow
+        # if a procfile exists without a web entry and dockerfile, assume heroku workflow
+        # and return empty structure as only web type needs to be created by default and
+        # other types have to be manually scaled
         elif release.build.procfile and 'web' not in release.build.procfile:
-            structure = {'cmd': 1}
+            structure = {}
 
         # default to heroku workflow
         else:


rootfs/api/models/app.py
```diff
@@ -393,13 +393,14 @@ def _default_structure(self, release):
...
+        # if a dockerfile, assume docker workflow
+        elif release.build.dockerfile:
             structure = {'cmd': 1}
 
-        # if a procfile exists without a web entry, assume docker workflow
+        # if a procfile exists without a web entry,
+        # assume heroku workflow with nothing to scale by default
         elif release.build.procfile and 'web' not in release.build.procfile:
-            structure = {'cmd': 1}
+            structure = {}
```
helgi: Are you sure we want to have *no* proc types in here?

kmala: yes... @slack suggested it that way

helgi: This is the only place that now defines the default, I removed all other hard coded places. As it stands it should give you an app with no RC, I would recommend trying this and verifying by hand

kmala: yes i tested it using procfiles with having 'web' and no 'web'

helgi: When you have a `Procfile` but it does not have `web` in it, then what RCs get created? Can you add the tests for it?

kmala: doesn't scale by default so no RC's get created during the deploy.we need to do a manual scale then the rc with the proctype name present in the procfile gets created.

helgi: Ah, I see what you are going for. Why wouldn't we grab the type in `Procfile` and scale that to 1? In case we have more than 1 type? I'm going on a limb that not auto picking a type is how Heroku behaves?

kmala: yes..only the web type gets scaled by default to 1 other we have to manually scale them.

helgi: Sounds good, can you add a comment to that line as to why it is empty and what the intent behaviour is

kmala: i had already changed the comment 'assume heroku workflow with nothing to scale by default' isn't that clear enough?

helgi: In retrospect it is but given I asked all these questions I would have to say no :D

kmala: updated.

bacongobbler:
With this new logic, an app with a Dockerfile and a Procfile will default to the `cmd` process type, even if the application defines type `web`. I don't think that's the behaviour we want.

kmala:
@bacongobbler yes i know it used to be the case but from the comments(https://github.com/deis/controller/blob/master/rootfs/api/models/app.py#L391-L406) i thought web is used only for the heroku workflow and not  docker workflow.
Adding the logic again

bacongobbler:
web can be used for anything. We just specifically call out `cmd` to supply an empty process to the container (so `docker run registry` can work, for example).

rootfs/api/models/app.py
```diff
@@ -393,13 +393,18 @@ def _default_structure(self, release):
         if not release.build.sha:
             structure = {'cmd': 1}
 
-        # if a dockerfile exists without a procfile, assume docker workflow
-        elif release.build.dockerfile and not release.build.procfile:
+        elif 'web' in release.build.procfile:
```
bacongobbler: this will fail if `release.build.procfile` == `None`. It'd be better to do

```
elif release.build.procfile and 'web' in release.build.procfile:
```

So we ensure that there's a procfile to parse before checking if there's a web process in there

rootfs/api/models/app.py
```diff
@@ -393,13 +393,18 @@ def _default_structure(self, release):
         if not release.build.sha:
             structure = {'cmd': 1}
 
-        # if a dockerfile exists without a procfile, assume docker workflow
-        elif release.build.dockerfile and not release.build.procfile:
+        elif release.build.procfile and 'web' in release.build.procfile:
```
bacongobbler: It's probably better to clump this and line 406 together in one branch:

```
elif release.build.procfile:
  if 'web' in release.build.procfile:
    structure = {'web': 1}
  else:
    # if a procfile exists without a web entry and dockerfile, assume heroku workflow
    # and return empty structure as only web type needs to be created by default and
    # other types have to be manually scaled
    structure = {}
```

helgi:
We clearly need to write some tests for this function given how much confusion is going on :)

rootfs/api/models/app.py
kmala: this would mean we wont scale by default if there is a dockerfile and procfile with a different proctype

bacongobbler:
We did; they just got removed with the Container model removal.

v1 tests as shown [here](https://github.com/deis/deis/blob/fc65359c2ba5541aadf35836940bf312007956de/controller/api/tests/test_container.py#L450-L499)

rootfs/api/models/app.py
bacongobbler: in the current branch logic if there's a dockerfile and a procfile, the process type will automatically be `{'cmd': 1}` because `elif release.build.dockerfile` will be evaluated before we even get to that point.

Good point though. Another pass at it:

```
elif release.build.procfile:
  if 'web' in release.build.procfile:
    structure = {'web': 1}
  else:
    # if a procfile exists without a web entry, assume heroku workflow
    # and return empty structure as only web type needs to be created by default and
    # other types have to be manually scaled
    structure = {}
    # this release has a procfile and a dockerfile, so we should assume the 'cmd'
    # process type in this case
    if release.build.dockerfile:
      structure = {'cmd': 1}
```

helgi:
@bacongobbler https://github.com/deis/controller/blob/master/rootfs/api/tests/test_pods.py#L479-L535 that test still exists, the updated version may not be exercising it correctly?

rootfs/api/models/app.py
kmala: no..i am checking whether procfile has web before checking the docker file so it would create a type of web

bacongobbler:
When I get some free time I'll write out some unit tests for #647 but I'll stop blocking this with "what-ifs".

PR #648, opened 2016-04-20
ref(dockerclient): use docker.auth.split_repo_name instead of our own strip_prefix

helgi:


bacongobbler:
this apparently does not exist: https://travis-ci.org/deis/controller/builds/124387741#L346

mboersma:
`split_repo_name` isn't re-exported in [`docker/auth.py/__init__.py`](https://github.com/docker/docker-py/blob/master/docker/auth/__init__.py), so you can only get to it as `docker.auth.auth.split_repo_name`.

helgi:
Good catch. That's what I get for cherry-picking from other branches :|

Issue #639, opened 2016-04-15
Private registry spike


krancour:
This issue is a place for us to hammer out our approach to resolving https://github.com/deis/controller/issues/253

cc @helgi @kmala 

Discussed offline earlier today:

Will prove the concept first:
* No CLI changes initially.  Configuration can be set via `deis config:set`.  This _will_ be replaced with a well-thought-out UX and necessary CLI changes before the feature is released.
* Controller will special-case its handling of relevant environment variables set via `deis config:set`-- again, this is temporary.
* Controller will construct `imagePullSecret` from relevant app configuration and make app RCs reference the same.
* Initially, this will bypass Workflow's internal registry.  Kubernetes will pull images _directly_ from the location specified by `deis builds:create` / `deis pull`.

With the concept proven:
* Improve UX / update CLI.
* Do we want do / can we institute private, per-app repos Workflow's internal registry, each requiring authentication in order to push or pull images?  This would improve platform security overall.  Ultimately, then, each app/namespace would have its own `imagePullSecret` for the _internal registry_.  This would not be specified by `deis` users, but could be auto-generated (when an app is created?).  Any registry authentication details provided by the user would be used _by the controller_ to pull images from the private registry before pushing said images to the internal registry-- i.e. similar to what we do today, and correcting for the fact that the PoC will likely bypass this step.

@helgi correct me, please, if any of this seems to differ from what we discussed.

helgi:
The way I like to think about this is 3 phases:

1. Use `config:set` per app to set a Registry authentication to prove that Controller can pull from External Registry (GCR / ECR / etc) and shovel into the Deis Internal Registry, requireing *no* change to Kubernetes as Deis Internal Registry has no auth
2. Better UX (as discussed above) *and* pull from External Registry *directly* via k8s imagePullSecret, bypassing Deis Internal Registry
3. Make Deis Internal Registry smarter around auth (but this will come later)

krancour:
Ok.  On point 1, we differ.  I was thinking we skip the "shoveling" into the internal registry, but you're thinking we keep that and the thing we skip is any `imagePullSecret` at all on the app pod(s).  So just the controller would use the provided auth details to pull from the external private registry...

I'm fine with that.  That seems pretty reasonable.

krancour:
Actually... I am confused now about where we're hoping to land eventually.

No. 1 has us skipping the `imagePullSecrets`, but still caching images in the internal registry.  As I indicated, I'd be fine with that.

[...]

PR #653, opened 2016-04-21
feat(registry): move registry information to its own top level Config resource

helgi:
## Summary of Changes

Moves `registry` information per application from normal ENV VAR to it's own top level in the Config object

Adds the ability to do `registry:set` in the CLI (requires https://github.com/deis/workflow-cli/pull/45), currently supported values are `password` and `username`.
Did not try rejecting things even if people added random keys, should be start doing that? Makes it lose flexibility but also makes it more explicit.

ref #639

## Associated [End To End Test](https://github.com/deis/workflow-e2e) PR(s)

https://github.com/deis/workflow-e2e/pull/177

## Associated [Documentation](https://github.com/deis/workflow) PR(s)

In the works

## Associated [Design Document](http://docs.deis.io/en/latest/contributing/design-documents)(s)

See #639 for more information

## Testing Instructions

1. Create a Deis Cluster
2. Register an app
3. Push an app to a private registry
4. `registry:set username=foo` and same with password
5. deploy application

## Pull Request Hygiene TODOs


rootfs/api/models/config.py
```diff
@@ -80,11 +81,18 @@ def set_healthchecks(self):
         # having succeeded.
         self.values['HEALTHCHECK_FAILURE_THRESHOLD'] = health['failure_threshold']
 
+    def set_registry(self):
+        # lower case all registry options for consistency
+        registry = {}
+        for key, value in self.registry.items():
+            registry[key.lower()] = value
```
mboersma: This could be a dict comprehension:
```python
registry = {key.lower(): value for key, value in self.registry.items()}
```

PR #654, opened 2016-04-21
feat(dockerapp): add default readiness check for the docker file app

kmala:
# Summary of Changes
Add readiness check for the docker app container

# Testing Instructions

1. Create a Deis Cluster
2. Register an docker file/deis pull app
3. It should work fine and you should see a readiness check when you describe the pod

# Pull Request Hygiene TODOs

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -427,6 +427,9 @@ class KubeHTTPClient(object):
             # Find if target port exists already, update / create as required
             if routable:
                 port = self._get_port(image)
+                if port is None:
+                    logger.debug("Failed to find port for Docker image {}, defaulting to 5000".format(image))  # noqa
+                    port = 5000
                 for pos, item in enumerate(service['spec']['ports']):
                     if item['port'] == 80 and port != item['targetPort']:
                         # port 80 is the only one we care about right now
@@ -670,8 +673,7 @@ class KubeHTTPClient(object):
             image_info = docker_cli.inspect_image(image)
             port = int(list(image_info['Config']['ExposedPorts'].keys())[0].split("/")[0])
         except Exception:
-            logger.debug("Failed to find port for Docker image {}, defaulting to 5000".format(image))  # noqa
-            port = 5000
+            port = None
 
         return port
 
@@ -938,8 +940,8 @@ class KubeHTTPClient(object):
         # add in healthchecks
         if kwargs.get('healthcheck'):
             template = self._healthcheck(template, kwargs['routable'], **kwargs['healthcheck'])
-        elif kwargs.get('build_type') == "buildpack":
-            template = self._buildpack_readiness_probe(template)
+        else:
+            template = self._default_readiness_probe(template, image, kwargs.get('build_type'))
 
         url = self._api("/namespaces/{}/replicationcontrollers", namespace)
         resp = self.session.post(url, json=template)
@@ -1047,6 +1049,24 @@ class KubeHTTPClient(object):
 
         return controller
 
+    def _default_readiness_probe(self, controller, image, build_type):
+        if build_type == "buildpack":
+            readinessprobe = self._default_buildpack_readiness_probe()
+        else:
+            readinessprobe = self._default_dockerapp_readiness_probe(image)
+        if readinessprobe is None:
+            return controller
+        # Update only the application container with the health check
+        app_type = controller['spec']['selector']['type']
+        namespace = controller['spec']['selector']['app']
+        container_name = '{}-{}'.format(namespace, app_type)
+        containers = controller['spec']['template']['spec']['containers']
+        for container in containers:
+            if container['name'] == container_name:
+                container.update(readinessprobe)
+
+        return controller
+
     '''
     Applies exec readiness probe to the slugrunner container.
     http://kubernetes.io/docs/user-guide/pod-states/#container-probes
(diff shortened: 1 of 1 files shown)


rootfs/scheduler/__init__.py
```diff
@@ -671,7 +673,7 @@ def _get_port(self, image):
             port = int(list(image_info['Config']['ExposedPorts'].keys())[0].split("/")[0])
         except Exception:
             logger.debug("Failed to find port for Docker image {}, defaulting to 5000".format(image))  # noqa
-            port = 5000
+            port = None
```
helgi: Are you sure about this change? If so then the debug message above may have to change

rootfs/scheduler/__init__.py
```diff
@@ -1094,6 +1098,42 @@ def _buildpack_readiness_probe(self, controller, delay=30, timeout=5, period_sec
 
         return controller
 
+    '''
+    Applies tcp socket readiness probe to the docker app container only if some port is exposed
+    by the docker image.
+    '''
+    def _dockerapp_readiness_probe(self, controller, image, delay=5, timeout=5, period_seconds=5,
```
helgi: 5 seconds delay may be rather low, hmm but not too bad either

kmala: since readiness probe gets executed after the initial delay  from the start of the pod i think we should a bit early as it might start serving immediately unlike the slug type apps where we know for sure that it has to download the slug before it can start serving requests.

helgi: That's a fair point. I'm okay with it

rootfs/scheduler/__init__.py
```diff
@@ -1094,6 +1098,42 @@ def _buildpack_readiness_probe(self, controller, delay=30, timeout=5, period_sec
...
+                                   success_threshold=1, failure_threshold=1):
+        port = self._get_port(image)
+        if port is None:
+            return controller
+        readinessprobe = {
```
bacongobbler: Can we move a lot of this logic into a more generic function like `_readiness_probe`? There's a significant amount of duplicate code here with `_buildpack_readiness_probe()`. Normally I'd say it's fine but there is a significant amount of shared code here.

helgi: IMO that can wait until we have a real need for the generic that comes with future health check work

bacongobbler: Shipping with >10 lines of duplicate code (especially one that mimics the other, save for the probe) is never a good idea, refactoring or not. There's no technical reason why we shouldn't refactor this so there's less duplicate code other than time.

helgi: Time is the big factor, shipping Soon and all that. Time I would rather see spent on more verification around deploys than refactor code. Refactor can always happen during free moments here and there

@kmala up to you tho - if you think you can knock together a refactor that includes the 3 different types without taking a long time then go for it

kmala: @bacongobbler refactored...can you check again?

rootfs/scheduler/__init__.py
```diff
@@ -427,6 +427,9 @@ def _update_application_service(self, namespace, name, app_type, image, routable
             # Find if target port exists already, update / create as required
             if routable:
                 port = self._get_port(image)
+                if port is None:
```
bacongobbler: why are we moving this out of `_get_port()`?

kmala: i wan't to add a readiness probe only if the image exposed some port.

rootfs/scheduler/__init__.py
```diff
@@ -1082,17 +1102,33 @@ def _buildpack_readiness_probe(self, controller, delay=30, timeout=5, period_sec
...
-        return controller
+    '''
+    Applies tcp socket readiness probe to the docker app container only if some port is exposed
+    by the docker image.
+    '''
+    def _default_dockerapp_readiness_probe(self, image, delay=5, timeout=5, period_seconds=5,
+                                           success_threshold=1, failure_threshold=1):
+        port = self._get_port(image)
+        if port is None:
+            return None
```
bacongobbler: if the docker app's service is defaulting to port 5000, why aren't we doing that here? Is there a case where we don't want to add a readiness check here? If so let's document this in the code so others can understand the reasoning behind this.

rootfs/scheduler/__init__.py
bacongobbler: so if the app does not expose a port, we still want to expose port 5000? I think I'm missing something here. :confused:

kmala: that's something i planned to fix in another PR as i need some more details like what to do if a docker app doesn't expose any port, do we need to empty create service assuming he might expose a port during another release or don't create at all?

bacongobbler: k can we add a FIXME or something to note that this is a workaround?

rootfs/scheduler/__init__.py
kmala: i was confused with liveness probe where probe fail will restart the container but since it is readiness probe it shouldn't effect anything other than updating the endpoints which is not concern here as it not a routable app.But still i think its better not to add any probe.

rootfs/scheduler/__init__.py
kmala: i created a issue https://github.com/deis/controller/issues/656 to discuss what should be the behaviour

PR #659, opened 2016-04-24
ref(release): return earlier if build pack or dockerfile app is being used

helgi:
Simplifies the control flow in `publish`


bacongobbler:
nice cleanup.

PR #660, opened 2016-04-24
ref(scheduler): passthrough k8s state if scheduler does not understand it

helgi:
Reasoning here is that we should be showing people more information when we have not specifically covered the bases instead of potentially more obscure information piece, like "initialized"

Also added ContainerCreating as a state to cover a state we had previously not been using for k8s. It got added in 1.2

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -648,13 +648,14 @@ class KubeHTTPClient(object):
             return JobState.destroyed
 
         states = {
-            'Pending': JobState.initialized,
-            'Starting': JobState.starting,
-            'Running': JobState.up,
-            'Terminating': JobState.terminating,
-            'Succeeded': JobState.down,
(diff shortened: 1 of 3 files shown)


rootfs/scheduler/__init__.py
```diff
@@ -667,7 +668,11 @@ def resolve_state(self, pod):
                 # is the pod ready to serve requests?
                 return states[container_status]
 
-        return states[pod['status']['phase']]
+        if pod['status']['phase'] in states:
+            return states[pod['status']['phase']]
+
+        # if no match was found for deis mapping then passthrough the real state
+        return pod['status']['phase']
```
bacongobbler: slight enhancement: we can make this all a two-liner with

```
pod_state = pod['status']['phase']
return states.get(pod_state, pod_state)
```

reducing the cyclomatic complexity by 1.

PR #662, opened 2016-04-25
chore(settings): Allow for comma-delimited RESERVED_NAMES

krancour:
Allow for comma-delimited `RESERVED_NAMES`.

Hard-coded defaults for that env var are also eliminated from the controller itself.  Defaults will be added to the controller's container spec in the chart, thus cluster admins can modify if they choose.

Relates to https://github.com/deis/controller/issues/657

### DIFF
--- a/rootfs/deis/settings.py
+++ b/rootfs/deis/settings.py
@@ -249,7 +249,7 @@ LOG_LINES = 100
 TEMPDIR = tempfile.mkdtemp(prefix='deis')
 
 # names which apps cannot reserve for routing
-DEIS_RESERVED_NAMES = [os.environ.get('RESERVED_NAMES', 'deis')]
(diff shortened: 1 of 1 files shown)


rootfs/deis/settings.py
```diff
@@ -249,7 +249,7 @@
 TEMPDIR = tempfile.mkdtemp(prefix='deis')
 
 # names which apps cannot reserve for routing
-DEIS_RESERVED_NAMES = [os.environ.get('RESERVED_NAMES', 'deis')]
+DEIS_RESERVED_NAMES = os.environ.get('RESERVED_NAMES', '').replace(' ', '').split(',')
```
bacongobbler: we still need `deis` as a reserved name.

bacongobbler:
can you please link here the relevant chart change for RESERVED_NAMES once you're done? Thanks!

krancour:
@bacongobbler complementary chart changes are in https://github.com/deis/charts/pull/241

PR #665, opened 2016-04-26
feat(deploy): support doing a larger number than 1 in, 1 out deployment

helgi:
## Summary of Changes

By default the deploy system will take the node count the application is deployed on (using tags), but can also look at `DEPLOY_BATCHES` per application and a global Django django setting with the same name.

Example of the default flow is as follows: As a user I have 5 Nodes my application deploys to, the application has a scale of 17, this will lead to the following deploy cadence:
5, 5, 5, 2 (each item represents a in / out), so 5 pods being brought up 3x and then at the end 2 pods

e2e will use this by default due to the node count it has

A app specific test is also available at https://github.com/deis/workflow-e2e/pull/184

## Associated [Documentation](https://github.com/deis/workflow) PR(s)

Documentation will come later when we have settled on the env name (may change due to Deployments API, may not)

## Testing Instructions

1. Create a Deis Cluster
2. Register an app
3. One of following needs to be true
  a. Make sure to have more than 1 node
  b. Tell the app it should use `DEPLOY_BATCHES=2`
  b. Configure Controller ENV var `DEPLOY_BATCHES=2`
4. scale to 11
5. Deploy app a few times
6. Look at logs and verify it deploys happened at the step interval defined in `DEPLOY_BATCHES` or the amount of nodes you have

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -341,6 +341,9 @@ class App(UuidAuditedModel):
             self.structure = self._default_structure(release)
             self.save()
 
+        # see if the app config has deploy batch preference, otherwise use global
+        batches = release.config.values.get('DEPLOY_BATCHES', settings.DEPLOY_BATCHES)
+
         # deploy application to k8s. Also handles initial scaling
         deploys = {}
         build_type = app_build_type(release)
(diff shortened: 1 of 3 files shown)


arschles:
Code LGTM. I did not test this manually, however

slack:
Testing Image: quay.io/jhansen/controller:git-55f4256

slack:
Looks good, even when the batch size is > the running process count.

25 unit deploy ~ 3m7s

bacongobbler:
Little late to the conversation, but any way we can namespace this envvar as DEIS_DEPLOY_BATCHES? DEPLOY_BATCHES could be an environment variable used by the application.

PR #675, opened 2016-04-27
fix(scheduler): check if the pods are ready in scale up operations before comparing to desired state

helgi:
The code was naively checking if any pods were in the current RC but did not check if they were actually ready. This poses a problem for deploys

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -855,13 +855,7 @@ class KubeHTTPClient(object):
             pods = self._get_pods(namespace, labels=labels).json()
             for pod in pods['items']:
                 # now that state is running time to see if probes are passing
-                if (
-                    pod['status']['phase'] == 'Running' and
-                    # is the readiness probe passing?
-                    self._pod_readiness_status(pod) == 'Running' and
-                    # is the pod ready to serve requests?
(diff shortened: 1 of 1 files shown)


arschles:
Code LGTM1. I have not (yet) manually tested

rootfs/scheduler/__init__.py
```diff
@@ -1371,6 +1371,19 @@ def _pod_liveness_status(self, pod):
 
         return True
 
+    def _pod_ready(self, pod):
+        """Combines various checks to see if the pod is considered up or not by checking probes"""
+        if (
```
mboersma: This could be just a boolean expression.
```python
return (
    pod['status']['phase'] == 'Running' and
    # is the readiness probe passing?
    self._pod_readiness_status(pod) == 'Running' and
    # is the pod ready to serve requests?
    self._pod_liveness_status(pod)
)
```

helgi: Yeah, I had some other plans (due to a different PR I'm working on) but I like your approach better. I'll update and make my local thing work with this

mboersma:
I tested this on top of workflow-beta3 without any issues. I was able to `curl` immediately after `git push` returns, but I didn't try hard to break that (yet). 👍

PR #678, opened 2016-04-27
fix(readinessprobe): container shouldn't be restarted if the database is unavailable

kmala:
# Summary of Changes

Currently the controller pod restarts even if the database is not responding as the check is part of the liveness probe.It is enough if the controller just doesn't respond to requests than restarting as restarting is not gone help anything to make database up and hence it should be as part of the readiness probe. 

# Testing Instructions

1. Create a Deis Cluster
2. Register an app to check the controller is responding
3. stop the database
4. now controller should still be running but not serving requests
5. start the database again and controller should start serving requests.

# Pull Request Hygiene TODOs

### DIFF
--- a/rootfs/api/views.py
+++ b/rootfs/api/views.py
@@ -26,10 +26,9 @@ import logging
 logger = logging.getLogger(__name__)
 
 
-class HealthCheckView(View):
+class ReadinessCheckView(View):
     """
-    Simple health check view to determine if the server
-    is responding to HTTP requests and DB connection / query.
+    Simple readiness check view to determine DB connection / query.
     """
 
     def get(self, request):
@@ -47,6 +46,17 @@ class HealthCheckView(View):
                 status=status.HTTP_503_SERVICE_UNAVAILABLE
             )
 
+        return HttpResponse("OK")
+    head = get
+
+
+class LivenessCheckView(View):
+    """
+    Simple liveness check view to determine if the server
+    is responding to HTTP requests.
+    """
+
+    def get(self, request):
         return HttpResponse("OK")
     head = get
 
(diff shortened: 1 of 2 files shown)


helgi:
What happens in the situation where I have a Controller that is up and running, the Database is down?

If I run any API operation then will this just bomb out since it can't talk to the DB and a 500 happens? How will I as an operator debug that and also how will we prevent k8s resources to potentially be created (if DB is down it really shouldn't get to that state but well..)

kmala:
if the DB is down and we run any api operation we get a 503 from the router...
```Error: 
503 Service Temporarily Unavailable
<html>
<head><title>503 Service Temporarily Unavailable</title></head>
<body bgcolor="white">
<center><h1>503 Service Temporarily Unavailable</h1></center>
<hr><center>nginx/1.9.6</center>
</body>
</html>

helgi:
We need to improve that, both a better error message (not HTML) and we need to log information to the controller logs to reflect the DB is down so an Operator can figure out what is happening

kmala:
yes better error message need to be done on the router side, i can open a issue in the router and currently in the controller logs we say database health check failed but we don't say the exact error because https://github.com/deis/controller/blob/master/rootfs/api/views.py#L43 is not working

arschles:
If the DB goes down and comes back up, will the controller successfully reconnect to it?

kmala:
yes..it would....had tested executing few deis cli commands

kmala:
rebased.

arschles:
reviewing now. also, is there an issue for improving the error message that comes back through the router, to the CLI?

arschles:
Code LGTM. I have not manually tested this patch

kmala:
issue in router for better error messaging https://github.com/deis/router/issues/174

PR #691, opened 2016-05-03
fix(config): proc type names validation should allow alphanumeric

helgi:
## Summary of Changes

And the way the regex was written it would also match as many valid chars as it could find, so even one a-z char at the start was saying the name was valid

https://docs-v2.readthedocs.io/en/latest/using-workflow/process-types-and-the-procfile/#declaring-process-types

## Testing Instructions

Following the test case should be enough

1. Create a Deis Cluster
2. Register an app
3. Create an app that uses a Procfile, use `we&b` as a name for the proc

## Pull Request Hygiene TODOs

### DIFF
--- a/rootfs/api/serializers.py
+++ b/rootfs/api/serializers.py
@@ -11,7 +11,9 @@ from rest_framework import serializers
 
 from api import models
 
-PROCTYPE_MATCH = re.compile(r'^(?P<type>[a-z]+)')
+# proc type name is alphanumeric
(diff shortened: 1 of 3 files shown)


bacongobbler:
you can't use numbers in a process type. I like this change except for the introduction of numbers. I'd argue that we stick to Heroku's format, since they introduced it.

bacongobbler:
> The syntax is defined as:

> - <process type> – an alphanumeric string, is a name for your command, such as web, worker, urgentworker, clock, etc.

huh. I was wrong about that. Never mind!

Issue #612, opened 2016-04-07
Improve exception handling from models


helgi:
Right now exceptions are either bubbled up all the way to the end user or they are included as part of the validation/error messages

This is not very user friendly. A wholesale rethink of the exception handling is in order, including how we can document the exceptions, show them properly to the admins without having to be in debug mode (log to controller log with a reference code that can be found in the error message itself?) to new a few enhancements

Example of an error message when uploading malformed certificate:

> Could not load certificate: [('PEM routines', 'PEM_read_bio', 'no start line')]

This was wrapped in `{'certificate': <error here>}` which is not very nice to read

helgi:
https://github.com/deis/controller/pull/626 has model::app raise a KubeException, it's kinda weird but it makes the way we deal with the view work smoother. Another reason why we need to rethink things

helgi:
Closing this since things have been cleaned up pretty well. Any future work should have a new ticket with a new proposal

PR #702, opened 2016-05-03
ref(models): use DRF APIException as a base for all exceptions

helgi:
## Summary of Changes

Move to using `DeisException` that extends `APIException` (with 400 as a status code) instead of `EnvironmentError`, have `AlreadyExists` extend `APIException` with 409 and `KubeException` and Logger related exceptions get raised up to the view as `ServiceUnavailable`

Additionally using `from e` when re-raising exception to keep contextual information if that is ever required (https://www.python.org/dev/peps/pep-3134/)

https://github.com/tomchristie/django-rest-framework/blob/master/docs/api-guide/exceptions.md has further info on DRF exceptions

ref #612

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -11,11 +11,12 @@ from urllib.parse import urljoin
 
 from django.conf import settings
 from django.db import models
-from django.core.exceptions import ValidationError
+from rest_framework.exceptions import ValidationError, NotFound
 from jsonfield import JSONField
 
 from deis import __version__ as deis_version
-from api.models import UuidAuditedModel, log_event, AlreadyExists
+from api.models import UuidAuditedModel, log_event, AlreadyExists, \
+    DeisException, ServiceUnavailable
 
 from api.utils import generate_app_name, app_build_type
 from api.models.release import Release
@@ -43,7 +44,7 @@ def validate_app_structure(value):
         if any(int(v) < 0 for v in value.values()):
             raise ValueError("Must be greater than or equal to zero")
     except ValueError as err:
-        raise ValidationError(err)
+        raise ValidationError(str(err))
 
 
 def validate_reserved_names(value):
@@ -152,7 +153,10 @@ class App(UuidAuditedModel):
             )
 
         # create required minimum resources in k8s for the application
-        self._scheduler.create(self.id)
+        try:
+            self._scheduler.create(self.id)
+        except KubeException as e:
+            raise ServiceUnavailable(str(e)) from e
 
         # Attach the platform specific application sub domain to the k8s service
         # Only attach it on first release in case a customer has remove the app domain
@@ -164,8 +168,8 @@ class App(UuidAuditedModel):
         try:
             # attempt to remove application from kubernetes
             self._scheduler.destroy(self.id)
-        except KubeException:
-            pass
+        except KubeException as e:
+            raise ServiceUnavailable(str(e)) from e
 
         self._clean_app_logs()
         return super(App, self).delete(*args, **kwargs)
@@ -213,7 +217,7 @@ class App(UuidAuditedModel):
             while True:
                 # timed out
                 if elapsed >= timeout:
-                    raise KubeException('timeout - 5 minutes have passed and pods are not up')
+                    raise DeisException('timeout - 5 minutes have passed and pods are not up')
 
                 # restarting a single pod behaves differently, fetch the *newest* pod
                 # and hope it is the right one. Comes back sorted
@@ -234,7 +238,6 @@ class App(UuidAuditedModel):
 
                 elapsed += 5
                 time.sleep(5)
-
         except Exception as e:
             err = "warning, some pods failed to start:\n{}".format(str(e))
             log_event(self, err, logging.WARNING)
@@ -261,9 +264,17 @@ class App(UuidAuditedModel):
         self.create()
 
         if self.release_set.latest().build is None:
-            raise EnvironmentError('No build associated with this release')
+            raise DeisException('No build associated with this release')
 
         release = self.release_set.latest()
+
+        # Validate structure
+        try:
+            for target, count in structure.copy().items():
+                structure[target] = int(count)
+            validate_app_structure(structure)
+        except (TypeError, ValueError) as e:
+            raise DeisException('Invalid scaling format: {}'.format(e))
 
         # test for available process types
         available_process_types = release.build.procfile or {}
@@ -272,7 +283,7 @@ class App(UuidAuditedModel):
                 continue  # allow docker cmd types in case we don't have the image source
 
             if container_type not in available_process_types:
-                raise EnvironmentError(
+                raise DeisException(
                     'Container type {} does not exist in application'.format(container_type))
 
         # merge current structure and the new items together
@@ -323,16 +334,15 @@ class App(UuidAuditedModel):
                     command=command,
                     **kwargs
                 )
-
             except Exception as e:
                 err = '{} (scale): {}'.format(self._get_job_id(scale_type), e)
                 log_event(self, err, logging.ERROR)
-                raise
+                raise ServiceUnavailable(e) from e
 
     def deploy(self, release):
         """Deploy a new release to this application"""
         if release.build is None:
-            raise EnvironmentError('No build associated with this release')
+            raise DeisException('No build associated with this release')
 
         # use create to make sure minimum resources are created
         self.create()
@@ -386,7 +396,7 @@ class App(UuidAuditedModel):
             except Exception as e:
                 err = '{} (app::deploy): {}'.format(self._get_job_id(scale_type), e)
                 log_event(self, err, logging.ERROR)
-                raise
+                raise ServiceUnavailable(err) from e
 
         # cleanup old releases from kubernetes
         release.cleanup_old()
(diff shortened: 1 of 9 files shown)


rootfs/api/models/build.py
```diff
@@ -43,10 +43,11 @@ def create(self, user, *args, **kwargs):
         try:
             self.app.deploy(new_release)
             return new_release
-        except Exception:
+        except Exception as e:
             if 'new_release' in locals():
                 new_release.delete()
-            raise
+
+            raise DeisException(str(e)) from e
```
bacongobbler: would this mask any important errors? I feel like we'd be losing the exception class that was raised in a traceback if we're converting it straight to a DeisException.

rootfs/api/models/certificate.py
```diff
@@ -179,15 +180,15 @@ def attach_in_kubernetes(self, domain):
...
             self._scheduler._create_secret(app, name, data)
         else:
             # update cert secret to the TLS Ingress format if required
             if secret != data:
                 try:
                     self._scheduler._update_secret(app, name, data)
-                except KubeHTTPException:
-                    raise
+                except KubeException as e:
+                    raise ServiceUnavailable(str(e)) from e
```
bacongobbler: I feel like this is out of scope of the scheduler and should be handled in the model instead.

rootfs/api/models/certificate.py
```diff
@@ -220,9 +221,8 @@ def detach(self, *args, **kwargs):
                 # We raise an exception when a secret doesn't exist
                 self._scheduler._get_secret(app, name)
                 self._scheduler._delete_secret(app, name)
-            except KubeHTTPException as e:
-                logger.critical(e)
-                raise EnvironmentError("Could not delete certificate secret {} for application {}".format(name, app))  # noqa
+            except KubeException as e:
+                raise ServiceUnavailable("Could not delete certificate secret {} for application {}".format(name, app)) from e  # noqa
```
bacongobbler: same here. The model or template should determine if the error is significant enough to return a SERVICE UNAVAILABLE message, not the scheduler.

rootfs/api/models/certificate.py
```diff
@@ -220,9 +221,8 @@ def detach(self, *args, **kwargs):
                 # We raise an exception when a secret doesn't exist
                 self._scheduler._get_secret(app, name)
                 self._scheduler._delete_secret(app, name)
-            except KubeHTTPException as e:
-                logger.critical(e)
```
bacongobbler: any reason why we're removing this message?

rootfs/api/models/build.py
helgi: We would not, see the from e portion - also you'd get the error message and proper status code up in the view, which is all it would end up getting piped up into and displayed to the end user as

rootfs/api/models/release.py
```diff
@@ -136,36 +134,38 @@ def previous(self):
...
-            build=prev.build,
-            config=prev.config,
-            summary=summary,
-            source_version='v{}'.format(version)
-        )
-
+    def rollback(self, user, version=None):
         try:
+            # if no version is provided then grab version from object
+            version = (self.version - 1) if version is None else int(version)
```
bacongobbler: can we please retain the old logic in this PR? Let's pull this into another PR for review's sake.

rootfs/api/models/certificate.py
helgi: But this is in the certificate model that I am re-raising as a different type, not the scheduler. Can you elaborate what you mean?

rootfs/api/models/certificate.py
helgi: Not consistent with any other exception handling we do

rootfs/api/models/release.py
```diff
@@ -136,36 +134,38 @@ def previous(self):
...
+
             if self.build is not None:
                 self.app.deploy(new_release)
             return new_release
-        except Exception:
+        except Exception as e:
             if 'new_release' in locals():
                 new_release.delete()
-            raise
+            raise DeisException(str(e))
```
bacongobbler: same here as before. Masking the original Exception with a DeisException may mask the original issue (for example a KeyuboardInterrupt or a MemoryError.

rootfs/api/models/certificate.py
helgi: It doesn't when service unavailable, it has the error message. I think you are just looking at the default message

rootfs/api/tests/test_auth.py
```diff
@@ -248,7 +248,7 @@ def test_passwd(self):
         }
         response = self.client.post(url, submit,
                                     HTTP_AUTHORIZATION='token {}'.format(token))
-        self.assertEqual(response.status_code, 400)
+        self.assertEqual(response.status_code, 401)
```
bacongobbler: I'd try to keep the responses the same as they were before for compatibility's sake. Is there a specific reason why we want to change this from a 400 to a 401?

rootfs/api/models/release.py
helgi: i do not see the point of that, as part of this PR I simplified the view file and this is one of those

rootfs/api/tests/test_auth.py
bacongobbler: If we are going forward with this change can you please make a doc PR to the API reference docs so we can keep it up to date? Thanks!

rootfs/api/models/release.py
helgi: It doesn't mask it in any meaningful way, also we never tried catching any of the exceptions you mentions and rarely tried to catch the top level Exception.

We need to not let random exceptions leak out to the user with a massive trace back. We need to handle them

rootfs/api/models/build.py
bacongobbler: You'd get the error message and the status code, but not the parent Exception class. We should handle those on a case-by-case basis moving forward rather than mask them all off with a generic Exception class, IMO.

rootfs/api/tests/test_auth.py
helgi: Because it is a failed authentication and not a bad request - we were using the wrong status code

rootfs/api/models/certificate.py
bacongobbler: Sorry, I meant handled in the views, not in the model. Need more :coffee:

rootfs/api/models/build.py
helgi: I think that's the wrong way to do it and lends for a bad user experience. If you are very concerned by this and the nicer UX then we can look at a custom exception handler that can log more ifnromation about parent exception, since I am attaching the originating exception to the re-raised one

rootfs/api/models/certificate.py
helgi: Views shouldn't be handling exceptions, the way we built it up so far isn't a very clean way

rootfs/api/models/build.py
bacongobbler: Yeah, I guess it's two different ways of approaching the same problem. I'm fine with this.

rootfs/api/models/certificate.py
bacongobbler: I've always been taught that in an MVC world, the views should handle errors and return the appropriate response code, whereas models should retrieve the data and raise errors if it could not retrieve said data. Is this wrong to assume?

helgi: Yes and no - There is more than one way to skin that cat.

One thing to pay close attention to here is that we are still raising exceptions up to the View, the difference here is we are curating what exceptions are raised, and they contain all the *required* information for the DRF exception handler to translate that into a `HttpResponse` object. In that sense we have not broken the paradigm you described.

When we use the models across one another (app calling build, build calling app, etc) then they are still Python Exceptions that we can catch and process on, however the Models are not Libraries. The scheduler is a library the Models use, and the Models are the Contract between the View and the 3rd party services (scheduler + k8s, logger, docker)

That's how I have been approach it in any case and how I ended up with what I have now

bacongobbler: works for me

rootfs/api/models/build.py
helgi: `DeisException` is our generic, like `EnvironmentError` was before - We can always introduce more Exception classes if we want to expose different things up to the API View or generally keep track of things differently

One thing that comes to mind for expansion: `KubeServiceUnavailable` that extends `ServiceUnavailable` and same way `LoggerServiceUnavailable` - Ultimately boils down to a `503` in the View.

rootfs/api/models/release.py
bacongobbler: it's just more code creep that I'd rather keep separate, for example changing the function signature and how we retrieve the version number from that is not important to this PR. It just slows down code review because I have to do a context switch in my mind to review this. It looks fine to me though.

helgi: I can look at splitting it out after lunch if it helps

rootfs/api/tests/test_auth.py
bacongobbler: if this was any time other than beta, I'd say this has to be changed, wrong status code or not. Because we're pre-prod though this is fine.

rootfs/api/models/release.py
bacongobbler: Nah I reviewed and it's ok. Let's just try to refrain from refactoring parallel code in the future so it's easier on the reviewers :)

rootfs/api/tests/test_auth.py
helgi: I agree there, it was the only reason why I touched that

rootfs/api/models/app.py
```diff
@@ -43,7 +44,7 @@ def validate_app_structure(value):
         if any(int(v) < 0 for v in value.values()):
             raise ValueError("Must be greater than or equal to zero")
     except ValueError as err:
-        raise ValidationError(err)
+        raise ValidationError(str(err))
```
arschles: is this a built-in DRF exception?

helgi: Yeah, ValidationError is both Django and DRF, the latter augments it a bit to better play with its View logic

bacongobbler: https://github.com/tomchristie/django-rest-framework/blob/bb56ca46ed6c07db0146dbdc61c672ff25f127de/rest_framework/exceptions.py#L61-L79

PR #703, opened 2016-05-04
ref(scheduler): use a try / except block instead of get_rc_status, allows for more precise handling

helgi:


mboersma:
This is good, but I think the [`retrying`](https://pypi.python.org/pypi/retrying) python library would simplify this type of code. Maybe in another refactoring.

helgi:
@mboersma you've mentioned it in two different tickets now. Maybe time to open up a meta ticket so someone can take it on? (hint hint: you ;), hehe... Lets at least track it somewhere

PR #711, opened 2016-05-08
fix(builds):User should be able to do a deploy an app after a failed attempt

kmala:
# Summary of Changes

delete the build if an exception occurs while doing a new release for the build.

# Testing Instructions

1. Create a Deis Cluster
2. Deploy a docker app without exposing any port and proc file so that the controller will raise an exception and the build fails.
3. change the docker file to expose the port  and do the deploy again.The deploy should succeed this time with the code changes.

# Pull Request Hygiene TODOs

### DIFF
--- a/rootfs/api/models/build.py
+++ b/rootfs/api/models/build.py
@@ -46,6 +46,7 @@ class Build(UuidAuditedModel):
         except Exception as e:
             if 'new_release' in locals():
                 new_release.delete()
+            self.delete()
 
             raise DeisException(str(e)) from e
 


helgi:
I'm all for the change but wondering if we did that inside the create function?

`build.create` is called in another place in the view file so if you moved this up into the Build model as a try / except then https://github.com/deis/controller/blob/master/rootfs/api/views.py#L250 would also benefit from it

kmala:
I din't think we are using it other place..changed the code to move a level up into the Build model

PR #716, opened 2016-05-09
ref(registry): prepend registry url on images up in Release model

helgi:
Removes the need for Deis registry url being known by the scheduler

Also clarifies what is the `image` and what is the `slug_url` in build pack templates

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -32,7 +32,7 @@ POD_BTEMPLATE = """\
     "containers": [
       {
         "name": "$id",
-        "image": "$slugimage",
+        "image": "$image",
         "env": [
         {
             "name":"PORT",
@@ -40,7 +40,7 @@ POD_BTEMPLATE = """\
         },
         {
             "name":"SLUG_URL",
-            "value":"$image"
+            "value":"$slug_url"
         },
         {
             "name": "BUILDER_STORAGE",
@@ -190,7 +190,7 @@ RCB_TEMPLATE = """\
         "containers": [
           {
             "name": "$containername",
-            "image": "$slugimage",
+            "image": "$image",
             "imagePullPolicy": "$image_pull_policy",
             "env": [
             {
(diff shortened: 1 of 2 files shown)


bacongobbler:
yes, this is definitely something the registtry module should know rather than the scheduler. Nice cleanup!

rootfs/scheduler/__init__.py
```diff
@@ -1024,9 +1019,9 @@ def _create_rc(self, namespace, name, image, command, **kwargs):  # noqa
                 secret = self._get_secret('deis', 'objectstorage-keyfile').json()
                 self._create_secret(namespace, 'objectstorage-keyfile', secret['data'])
 
-            l["image"] = image
+            l["slug_url"] = image
```
kmala: we need to remove the registry part for the the image or the slugrunner will fail with invalid path as this is basically the path of the slug to be downloaded.

rootfs/scheduler/__init__.py
```diff
@@ -530,23 +529,21 @@ def run(self, namespace, name, image, entrypoint, command, **kwargs):
...
+            'image': image,
             'image_pull_policy': settings.DOCKER_BUILDER_IMAGE_PULL_POLICY,
             'storagetype': os.getenv("APP_STORAGE"),
             'terminationGracePeriodSeconds': settings.KUBERNETES_POD_TERMINATION_GRACE_PERIOD_SECONDS  # noqa
         }
 
         if entrypoint == '/runner/init':
             POD = POD_BTEMPLATE
-            l["image"] = image
+            l["slug_url"] = image
```
kmala: we need to remove the registry part for the the image or the slugrunner will fail with invalid path as this is basically the path of the slug to be downloaded.

rootfs/scheduler/__init__.py
helgi: That should happen in `Release.image` and not in the scheduler. I can look at doing that, right now I made it so all but External Registries get the Internal Registry prepended. Just to confirm, only build packs should have that excluded, right?

helgi: I made is so `Release.image` does not prepend registry url to buildpack images

bacongobbler:
@kmala does the changes @helgi made seem fair now? Looks like your comments were resolved with changes elsewhere in release.py.

kmala:
@helgi @bacongobbler looks good now

PR #718, opened 2016-05-09
feat(images): use imagePullSecret when pulling a private image

helgi:
use a k8s Secret to encode a `.docker/config.json` to completely bypass the Controller when it comes to images if a private registry is being used

Handles the various possible errors coming from k8s and bubble that up through to the end user, similar to how the deis dockerclient can do.
k8s adds more information on top of what docker gives so an error message bubbled up from this compared to when the Controller was handling the Docker interaction

ref #639

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -338,7 +338,8 @@ class App(UuidAuditedModel):
                 'memory': release.config.memory,
                 'cpu': release.config.cpu,
                 'tags': release.config.tags,
-                'envs': envs,
+                'envs': release.config.values,
+                'registry': release.config.registry,
                 'version': "v{}".format(release.version),
                 'replicas': replicas,
                 'app_type': scale_type,
@@ -390,7 +391,8 @@ class App(UuidAuditedModel):
                 'memory': release.config.memory,
                 'cpu': release.config.cpu,
                 'tags': release.config.tags,
-                'envs': envs,
+                'envs': release.config.values,
+                'registry': release.config.registry,
                 # only used if there is no previous RC
                 'replicas': replicas,
                 'version': "v{}".format(release.version),
@@ -583,6 +585,7 @@ class App(UuidAuditedModel):
             'cpu': release.config.cpu,
             'tags': release.config.tags,
             'envs': release.config.values,
+            'registry': release.config.registry,
             'version': "v{}".format(release.version),
             'build_type': release.build.type,
         }
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -333,12 +333,11 @@ class App(UuidAuditedModel):
             # fetch application port and inject into ENV Vars as needed
             envs['PORT'] = release.get_port(routable)
 
-            image = release.image
             kwargs = {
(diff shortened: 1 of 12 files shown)


jchauncey:
Should we be manually testing this and if so how?

helgi:
## Testing Instructions

* Create a Deis Cluster
* Register an app
* Create a Private Repo on Quay.io
* Set user / pw via `deis registry:set username=<quay_user> password=<quay_token>` for the app
* Do deis pull referencing an image in a private repo

* The app called "abcd" should be deployed using the private image
* Describe the pod to ensure that image pull secret is set

helgi:
One e2e caught a scenario I had not accounted for so I need to investigate that better, looks like I am missing a data guard somewhere

jchauncey:
```
╭─jonathanchauncey at ENG000637 in ~/projects/deis/example-dockerfile-http on master✔ using ‹2.0.0-p645›
╰─± deis pull -a foo quay.io/jchauncey/privateapp
Creating build... done
╭─jonathanchauncey at ENG000637 in ~/projects/deis/example-dockerfile-http on master✔ using ‹2.0.0-p645›
╰─± kgpo --namespace=foo
NAME               READY     STATUS    RESTARTS   AGE
foo-v3-cmd-0tcd7   1/1       Running   0          2m
╭─jonathanchauncey at ENG000637 in ~/projects/deis/example-dockerfile-http on master✔ using ‹2.0.0-p645›
╰─± curl foo.devart.io
Powered by Deis
```

helgi:
Currently blocked on https://github.com/deis/controller/issues/721

jchauncey:
moving to rc1

bacongobbler:
how is this blocked by #721? Just call `registry.publish_release()` to pull the image. The image will be pushed to the internal registry, which then you can introspect and bypass any changes necessary to facilitate #721.

bacongobbler:
https://github.com/deis/controller/blob/b13152ac0bf0c0dfeeca3eefc1937cbadac008aa/rootfs/api/models/release.py#L113 pushes the image into the registry so it should be available for inspection.

helgi:
@bacongobbler that's not the point, I am aware of how the code works and how I could pull the image into the local registry. 

The point here is we are having k8s fully pulling images and dealing with the secrets but then having Controller pull the image into the insecure registry just to inspect a port, causing false sense of security *and* adding in additional download time *and* storage requirements. That's what is being discussed in #721

bacongobbler:
Right, but that's something we can implement later when we come to a conclusion on how we can resolve it. For now let's just follow what we've been doing since v1. Whether or not it's stored in the registry doesn't matter since the image is in the cluster anyways.

helgi:
Maybe what's being lost in translation here is that Authenticated External Registry is net new feature functionality as a fully support part of Deis (not a hack) so this scenario doesn't really compare to v1, which allowed for public only images (unless you copied configs over).

Also, by default the port is `5000` and can be overwritten with `deis config:set PORT=5001`

Doing this post 2.0 wouldn't be very viable since you'd be moving the user from introspection to the above, so we couldn't do this until 3.0

PR #720, opened 2016-05-09
ref(rootfs): rename DEBUG to DEIS_DEBUG

bacongobbler:
refs https://github.com/deis/workflow/issues/178


helgi:
Should we include https://github.com/deis/controller/blob/master/rootfs/deis/settings.py#L10 and the thing below it

mboersma:
> Should we include...

That's just `settings.DEBUG` though. It doesn't map to the env var and I think it would confuse Django to change it.

jchauncey:
moving this to rc1

helgi:
You will want to update this https://github.com/deis/controller/blob/0bc556388b252aca8ad5b1e023f1eb3e6e4045fc/rootfs/api/tests/test_pods.py#L610 i believe

PR #726, opened 2016-05-11
ref(release): use scheduler cleanup release code as it is identical

helgi:


rootfs/api/models/release.py
```diff
@@ -254,16 +254,7 @@ def _delete_release_in_scheduler(self, namespace, version):
...
-            self._scheduler._delete_rc(namespace, controller['metadata']['name'])
-            # Remove stray pods
-            labels = controller['metadata']['labels']
-            pods = self._scheduler._get_pods(namespace, labels=labels).json()
-            for pod in pods['items']:
-                if self._scheduler._pod_deleted(pod):
-                    continue
-
-                self._scheduler._delete_pod(namespace, pod['metadata']['name'])
+            self._scheduler._cleanup_release(controller)
```
kmala: i think namespace also should be passed as the argument.

helgi: Good eye, fixed

PR #727, opened 2016-05-11
ref(build): give Build model awareness of what type it is and what version it is

helgi:


rootfs/api/models/build.py
```diff
@@ -27,17 +27,28 @@ class Meta:
...
+        elif self.sha:
+            return 'buildpack'
+        else:
+            # docker image (or any sort of image) used via deis pull
+            return 'image'
+
+    @property
+    def version(self):
+        return 'git-{}'.format(self.sha) if self.type == 'buildpack' else 'latest'
+
```
kmala: doesn't docker file app also has sha set?

helgi: hmm, yeah I guess they technically could - I'll have a bit of a look see. Was mostly basing it off the function we had in the utils file

rootfs/api/models/build.py
```diff
@@ -27,17 +27,28 @@ class Meta:
...
-        source_version = 'latest'
-        if self.sha:
-            source_version = 'git-{}'.format(self.sha)
-
         new_release = latest_release.new(
             user,
             build=self,
             config=latest_release.config,
-            source_version=source_version
+            source_version=self.version
```
kmala: since source_version is just based on sha and not on the build type shouldn't version also be the same?

helgi:
I updated the PR, I made a new property called `source_based` instead of just checking for `sha`. It feels clearer since there is a nice comment people can cross reference

Issue #193, opened 2015-12-23
dockerize the integration testing environment


arschles:

arschles:
closing, as this is a duplicate of #194

arschles:
I overlooked the fact that #194 is the PR that fixes this issue. Reopening.

PR #731, opened 2016-05-12
chore(Dockerfile,__init__.py): DEIS_RELEASE -> WORKFLOW_RELEASE

vdice:
# Summary of Changes

Updated `DEIS_RELEASE` to be `WORKFLOW_RELEASE` as done similarly elsewhere (see https://github.com/deis/workflow/pull/193)
TODO
- [ ] similar PR in https://github.com/deis/example-go/blob/master/web.go#L24-L27

# Pull Request Hygiene TODOs

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -142,7 +142,7 @@ RCD_TEMPLATE = """\
                 "value":"$id"
             },
             {
-                "name":"DEIS_RELEASE",
+                "name":"WORKFLOW_RELEASE",
                 "value":"$appversion"
             }
             ]
@@ -206,7 +206,7 @@ RCB_TEMPLATE = """\
                 "value":"$id"
             },
             {
-                "name":"DEIS_RELEASE",
+                "name":"WORKFLOW_RELEASE",
                 "value":"$appversion"
             },
             {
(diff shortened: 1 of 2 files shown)


rootfs/scheduler/__init__.py
```diff
@@ -142,7 +142,7 @@
                 "value":"$id"
             },
             {
-                "name":"DEIS_RELEASE",
+                "name":"WORKFLOW_RELEASE",
```
helgi: Does anyone know why we inject this into the RCs? It uses `$appversion` and well... that's the application release version - Can we remove it?

Either way, I don't think this should be renamed if it is not being used for release purposes

bacongobbler: it's used so that the application can gain knowledge of its own release information. It's important!

bacongobbler: we should update example-go to use WORKFLOW_RELEASE: https://github.com/deis/example-go/blob/master/web.go#L24-L27

helgi: Ah, right, okay if all of those get updated then we can make this change :)

vdice: thanks @bacongobbler & @helgi ; example-go PR here https://github.com/deis/example-go/pull/14

PR #732, opened 2016-05-12
ref(scheduler): unify how information is applied to the application container object in a Pod template

helgi:
This also allowed me to move healthchecks (default and otherwise) under the new _set_container and have deis run make use of it

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -565,11 +565,18 @@ class KubeHTTPClient(object):
         else:
             args = [command[1:-1]]
 
-        containers = template['spec']['containers'][0]
-        containers['command'] = [entrypoint]
-        containers['args'] = args
+        spec = template['spec']
 
-        self._set_environment(containers, namespace, **kwargs)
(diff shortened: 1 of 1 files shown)


rootfs/scheduler/__init__.py
```diff
@@ -870,9 +884,9 @@ def _wait_until_pods_terminate(self, namespace, labels, current, desired):
 
             time.sleep(1)
 
-        logger.debug("{} pods in namespace {} are terminated".format(delta, namespace))  # noqa
+        logger.debug("{} pods in namespace {} are terminated".format(delta, namespace))
```
arschles: so is fabric8 now cool with this?

helgi: flake8, yeah - That used to be a line that was too long (over 99 chars). No more. The CI runs flake8 and co so you'd see a red build otherwise

arschles: ah, yea. forgot the name and I guess I forgot that I forgot. good to know RE the line length check

arschles:
Code LGTM

PR #736, opened 2016-05-16
fix(apps): cleanup fails on older env secrets that lacked version info

helgi:


helgi:
Going to pull this one out for now, not useful

PR #738, opened 2016-05-16
ref(scheduler): rework the way deis run is handled, timeouts and more

### LINKED ISSUE #590: When deis run fails show the error message from the container
Right now we output the full JSON from k8s. See #589

### LINKED ISSUE #681: app::run timeout handling needs to be rewritten
`deis run` has a ~30 second timeout and that is not always sufficient as it can take longer to just download and initialise a container

From https://ci.deis.io/job/workflow-test-pr/1305/console

```
00:21:08.748 Creating build... ...o...o...o...o...o...o...o...o...o...o...o...o...o...o...o...o...o...odone
00:21:08.794 $ deis apps:run --app=test-986838140 echo Hello, 世界
00:21:08.794 Running 'echo Hello, 世界'...
00:21:08.794 Error: 
00:21:08.794 503 Service Unavailable
00:21:08.795 detail: failed to Pod start took more than 30 seconds: 200 OK
00:21:08.795 {'apiVersion': 'v1', 'status': {'conditions': [{'message': 'containers with unready status: [test-986838140-v2-run-oe5g4]', 'lastTransitionTime': '2016-04-28T22:31:21Z', 'type': 'Ready', 'reason': 'ContainersNotReady', 'status': 'False', 'lastProbeTime': None}], 'containerStatuses': [{'lastState': {}, 'image': '10.111.250.54:80/test-986838140:v2', 'name': 'test-986838140-v2-run-oe5g4', 'state': {'waiting': {'message': 'Image: 10.111.250.54:80/test-986838140:v2 is ready, container is creating', 'reason': 'ContainerCreating'}}, 'ready': False, 'imageID': '', 'restartCount': 0}], 'phase': 'Pending', 'startTime': '2016-04-28T22:31:21Z', 'hostIP': '10.240.0.11'}, 'metadata': {'creationTimestamp': '2016-04-28T22:31:20Z', 'name': 'test-986838140-v2-run-oe5g4', 'namespace': 'test-986838140', 'selfLink': '/api/v1/namespaces/test-986838140/pods/test-986838140-v2-run-oe5g4', 'uid': 'ee8a5345-0d90-11e6-ac2f-42010a800060', 'resourceVersion': '1539'}, 'kind': 'Pod', 'spec': {'containers': [{'terminationMessagePath': '/dev/termination-log', 'imagePullPolicy': 'IfNotPresent', 'image': '10.111.250.54:80/test-986838140:v2', 'name': 'test-986838140-v2-run-oe5g4', 'volumeMounts': [{'mountPath': '/var/run/secrets/kubernetes.io/serviceaccount', 'name': 'default-token-rbekr', 'readOnly': True}], 'args': ['-c', 'echo Hello, 世界'], 'resources': {}, 'command': ['/bin/bash']}], 'restartPolicy': 'Never', 'dnsPolicy': 'ClusterFirst', 'serviceAccountName': 'default', 'serviceAccount': 'default', 'volumes': [{'secret': {'secretName': 'default-token-rbekr'}, 'name': 'default-token-rbekr'}], 'terminationGracePeriodSeconds': 5, 'securityContext': {}, 'nodeName': 'gke-helm-testing-9df4-default-pool-6ea2ff67-97f5'}}
00:21:08.802 
00:21:08.802 $ deis apps:destroy --app=test-986838140 --confirm=test-986838140
00:21:08.802 Destroying test-986838140...
00:21:08.802 done in 0s
00:21:08.802 $ deis auth:cancel --username=test-37025 --password=asdf1234 --yes
```

In there the container was in the `ContainerCreating` state still when it bailed out

helgi:
# Summary of Changes

Use the same "is the pod ready" function as normal deployments, also give deis run 20 mins max (arbitrary but is somewhat what the gunicorn worker can do)

On failure the whole Pod manifest is no longer returned. This change affects how we do tests here but minimally

Mocking was improved to help move a Pod from Pending to Running and if it is a deis run Pod (not attached to RC) then it also goes to Succeeded state. This helps with test coverage but is not the full Pod state transition route

Fixes #681
Fixes #590

deis run already has e2e

Documentation still needs to be updated to handle the little change in API

# Testing Instructions

1. Create a Deis Cluster
2. Register an app
3. run a `deis run` command
4. things should work!

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -8,7 +8,7 @@ from urllib.parse import urljoin
 import base64
 
 from django.conf import settings
-from .states import JobState
+from .states import PodState
 import requests
 from requests_toolbelt import user_agent
 from .utils import dict_merge
@@ -24,7 +24,13 @@ POD_BTEMPLATE = """\
   "kind": "Pod",
   "apiVersion": "$version",
   "metadata": {
-    "name": "$id"
+    "name": "$id",
+    "labels": {
+      "app": "$app",
+      "version": "$appversion",
+      "type": "$type",
+      "heritage": "deis"
+    }
   },
   "spec": {
     "containers": [
@@ -81,7 +87,13 @@ POD_TEMPLATE = """\
   "kind": "Pod",
   "apiVersion": "$version",
   "metadata": {
-    "name": "$id"
+    "name": "$id",
+    "labels": {
+      "app": "$app",
+      "version": "$appversion",
+      "type": "$type",
+      "heritage": "deis"
+    }
   },
   "spec": {
     "containers": [
@@ -527,6 +539,9 @@ class KubeHTTPClient(object):
         POD = POD_TEMPLATE
         l = {
             'id': name,
+            'app': namespace,
+            'appversion': kwargs.get('version'),
+            'type': 'run',
             'version': self.apiversion,
             'image': image,
             'image_pull_policy': settings.DOCKER_BUILDER_IMAGE_PULL_POLICY,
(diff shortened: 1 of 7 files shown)


helgi:
I also need to update the deis cli - Originally I wasn't really hot on updating the API which would require a CLI update but `rc` means little to nothing compared to `exit_code`

helgi:
Backing out the API change in this PR and moving it to a separate one that will depend on this one

rootfs/scheduler/__init__.py
```diff
@@ -568,42 +583,53 @@ def run(self, namespace, name, image, entrypoint, command, **kwargs):
...
+        self._wait_until_pods_are_ready(namespace, container, labels, desired=1)
+
+        try:
+            # give pod 20 minutes to execute (after it got into ready state)
+            # this is a fairly arbitrary limit but the gunicorn worker / LBs
+            # will make this timeout around 20 anyway.
+            # TODO: Revisit in the future so it can run longer
+            state = 'up'  # pod is still running
+            waited = 0
+            timeout = 1200  # 20 minutes
```
mboersma: Could this re-use the actual gunicorn timeout?
```python
from deis import gconf

timeout = gconf.timeout
```

rootfs/scheduler/__init__.py
```diff
@@ -568,42 +583,53 @@ def run(self, namespace, name, image, entrypoint, command, **kwargs):
...
 
-        data = ''
-        duration = 30
-        iteration = 1
-        while (iteration < duration):
-            try:
+        labels = {
+            'app': namespace,
+            'type': 'run',
+            'version': kwargs.get('version'),
```
bacongobbler: shouldn't this be for `appversion` instead?

rootfs/scheduler/__init__.py
helgi: We can't do that since that's the overall timeout for the worker and there are many other pod waits ahead of this `1200` second limit. If we used the gunicorn timeout then the worker would always timeout before the `while` loop would catch - I felt it was better to have a hardcoded one, not based on the gunicorn worker limit.

rootfs/scheduler/__init__.py
helgi: No, `run` in `App` defines it as `version` (https://github.com/deis/controller/blob/master/rootfs/api/models/app.py#L582)

rootfs/scheduler/__init__.py
```diff
@@ -568,42 +583,53 @@ def run(self, namespace, name, image, entrypoint, command, **kwargs):
...
+            if waited == timeout:
+                raise KubeException('Timed out (20 mins) while running')
 
-        return 0, data
+            # grab log information
+            log = self._pod_log(namespace, name)
+            log.encoding = 'utf-8'  # defaults to "ISO-8859-1" otherwise...
+
+            return exit_code, log.text
+        except KubeException:
```
mboersma: `except/raise` like this amounts to a no-op, so I think it should be removed.

helgi: Does that play with `finally`? I can give it a go just to see, usually have it coupled with `except`

rootfs/scheduler/states.py
```diff
@@ -1,7 +1,34 @@
...
+        return NotImplemented
+
+    def __lt__(self, other):
+        if self.__class__ is other.__class__:
+            return self.value < other.value
+
+        return NotImplemented
+
+
+@unique
```
mboersma: Cool, I didn't know about the [@unique Enum decorator](https://docs.python.org/3/library/enum.html#enum.unique).

rootfs/scheduler/__init__.py
bacongobbler: Sorry, I'm referring to the label instead of the keyword argument. `POD_TEMPLATE` [uses `$appversion`](https://github.com/helgi/controller/blob/7e44adb167eff9504774bb8283473ebadd66fd88/rootfs/scheduler/__init__.py#L93) for the application version number. `$version` is used for the API version number.

rootfs/scheduler/__init__.py
mboersma: The `finally` will happen regardless and then any `Exception` will be raised. This code just catches and re-raises `KubeException` explicitly but the behavior would be the same if that were omitted. Unless it's there as documentation to clue in someone reading the code that `KubeException` is likely here.

helgi: Okay, cool, convinced, going to change

rootfs/scheduler/__init__.py
helgi: That is a label selector which I am using for filtering and if you look at the `POD_TEMPLATE` then `$appversion` is for the templating but the key it self is `version`

I really need to just rename `$version` to `$apiversion` to remove this confusino

helgi:
This also fixes #590

PR #742, opened 2016-05-17
ref(logging): remove syslog logging since deis-logger handles all aggregation

helgi:


bacongobbler:
Funny enough [this code is fairly old](https://github.com/deis/deis/pull/457). As long as messages still end up in /dev/stdout (which in turn, `kubectl logs`) then I'm fine with this change.

PR #745, opened 2016-05-20
ref(exceptions/logging): unify logging and make sure exceptions go into logs

### LINKED ISSUE #568: Turning on DEBUG env var does not always spit out details in logs as they should
For example I had problems with the DB health check and it should be outputting the exception in `debug` mode but nothing got outputted when I turned it on.

Many other extra debugs did get outputted however so it may be a case where Views in Django do not get included (that's where the health check is happening) but we'd have to review that so we can assist people on the debug front going forward

helgi:
## Summary of Changes

Add a custom exception handler to deal with uncaught exceptions so they get properly logged to the stdout logging, and also return a 500 with a JSON body like everything else does.
A lot of scheduling logging has moved up from debug to info (the new default value) and now DEBUG=True (in the django config file) will cause debug information to outputting

`tbgrep` was added, usage would be `k logs -f <controller pod> | tbgrep`

One thing I started doing in this PR is to reduce the `str(e)` as the msg to Exceptions, instead writing more generic message and chaining the Exceptions to together which in turn get logged.

The idea for that is to show end users less internal error messages in API responses but still make it available to the Operator

Please list the issue(s) that this PR closes, similar to the below:

fixes #568
ref #612

### DIFF
--- a/rootfs/deis/gconf.py
+++ b/rootfs/deis/gconf.py
@@ -1,4 +1,6 @@
 import os
+import faulthandler
+faulthandler.enable()
 
 bind = '0.0.0.0'
 try:
--- a/rootfs/api/exceptions.py
+++ b/rootfs/api/exceptions.py
@@ -1,3 +1,38 @@
-class HealthcheckException(Exception):
+import logging
+from rest_framework.compat import set_rollback
+from rest_framework.exceptions import APIException, status
+from rest_framework.response import Response
+from rest_framework.views import exception_handler
+
+
+class HealthcheckException(APIException):
     """Exception class used for when the application's health check fails"""
     pass
+
+
+class DeisException(APIException):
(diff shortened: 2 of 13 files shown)


aledbf:
@helgi thank you for this!!!!

rootfs/deis/gconf.py
```diff
@@ -1,4 +1,6 @@
 import os
+import faulthandler
+faulthandler.enable()
```
mboersma: [`faulthandler`](https://docs.python.org/dev/library/faulthandler.html) is handy. I didn't know that was in Python 3.3+.

helgi: Yeah, haven't looked at if I can replace the `traceback` below but we will see in the future

PR #749, opened 2016-05-20
fix(scheduler): don't match container names in "deis run"

### LINKED ISSUE #748: "deis run boguscmd" sometimes returns 0
Run a nonexistent `deis run` command, or one that returns non-zero. It's effectively random whether the `deis` CLI returns 0 or the correct exit code.

This is blocking merge of deis/workflow-e2e#212.

mboersma:
# Summary of Changes

Skips the container name comparison in the case of a `deis run`-spawned pod, since they won't match. (`name` will be something like "nonfat-jailbird-run" but the container's name might be "nonfat-jailbird-v2-run-6atqv".) Using the first (only) container in the pod should be fine in this case.

This change fixes the error in my testing, and I always get the proper exit code back from `deis` now.

Closes #748.

refs deis/workflow-e2e#212

# Testing Instructions

Run a nonexistent deis run command, or one that returns non-zero, then query the `deis` CLI return code:
```console
$ deis run boguscmd
$ echo $?
```

A non-zero exit code should *always* be returned. This error isn't deterministic so you should repeat the test many times.

# Pull Request Hygiene TODOs

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -1490,7 +1490,7 @@ class KubeHTTPClient(object):
         name = '{}-{}'.format(pod['metadata']['labels']['app'], pod['metadata']['labels']['type'])
         for container in pod['status']['containerStatuses']:
             # find the right container in case there are many on the pod
-            if container['name'] != name:
+            if container['name'] != name and not name.endswith("-run"):
                 continue
 
             if not container['ready']:


bacongobbler:
LGTM

helgi:
I would prefer we fixed the container name for run templates so we don't need to do this kind of special case in a generic function

helgi:
I made PR #751 that handles the `run` Pod without adding specific logic into any of the check methods since I'd prefer not to special case types in those. Give that a go and see if it works for you

PR #750, opened 2016-05-21
chore(Dockerfile): Refactor image to use ubuntu-slim

krancour:


helgi:
We should not be changing the whole repository to match the Docker images. The `rootfs` directory was already a bit of a stretch but adding another depth via `app` seems a bit too much.

Can we re-add the 2-3 lines in the `build.sh` scrip that made it so?

krancour:
I could agree that this change shouldn't be bundled with the transition to Ubuntu slim. I'll undo it. I do, however, feel this should still happen, even if as its own PR. All the other projects follow the "rootfs" pattern, but controller only _sort of_ follows it currently, because contrary to what is implied by the directory name `rootfs`, that directory is not copied to `/`. I think we have too many inconsistencies between projects and this is one obvious one that's really easy to pay down.

helgi:
I would argue that a project shouldn't even have a `rootfs` directory - Laying out a project in the format that it will distributed in feels messy to me and doesn't serve much benefit beyond the Dockerfile has to do 1 or 2 less things.

Look at RPM and DEB packages, that whole logic is usually kept in a `debian` folder in the root or `specs/debian` or `pkg` (and same for whatever else format). That's my preferred setup, ditching `rootfs` and having scripts that can do the right thing.

krancour:
It's a fine argument, but even if your position is correct (and it probably is), all the other components follow the rootfs pattern. Whether this project should change to match them or they should move off of that pattern is probably a subject best saved for later. Right now, we're just bikeshedding. I've amended the PR to not include a shift toward that pattern.

jchauncey:
we should definitely talk about that pattern. For some repos like fluentd we have to create a crazy amount of empty directories just to satisfy this pattern for 1 file.

rootfs/Dockerfile
```diff
@@ -1,21 +1,40 @@
...
+		git \
+		libffi-dev \
+		libpq-dev \
+		python3-dev \
+	&& apt-get clean \
+	&& rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/man /usr/share/doc
 
 # define execution environment
+WORKDIR /app
 CMD ["/app/bin/boot"]
```
arschles: this can change to `bin/boot` since the above `WORKDIR` is set

smothiki:
@krancour passed jenkins test. Please merge this one if there aren't any issues

rootfs/Dockerfile
krancour: I had tried that at one point and gotten a poor result at boot time.

krancour: I didn't dig too deep, but something in the boot script must make an assumption about what the cwd is.

PR #751, opened 2016-05-22
fix(scheduler): run pod container name did not match how app containers are done

### LINKED ISSUE #748: "deis run boguscmd" sometimes returns 0

helgi:
# Summary of Changes

Instead of using the `Pod` name for `run` templates then instead use the same container name logic as normal application `Pods` have on their app `Container`

Fixes #748
Replaces #749

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -36,7 +36,7 @@ POD_BTEMPLATE = """\
   "spec": {
     "containers": [
       {
-        "name": "$id",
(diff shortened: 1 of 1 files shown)


mboersma:
This fixes #748 in my manual testing. 👍

PR #757, opened 2016-05-30
fix(gunicorn): silence healthcheck points in access logs

### LINKED ISSUE #252: Silence /healthz requests in log
Django logs all HTTP transactions by default, but ideally GET requests on /health-check would be silenced or retargeted for DEBUG level only. Otherwise that constant traffic will crowd out useful log info.

helgi:
If `DEIS_DEBUG` is set then they show up again

Testing Plan:
* `make deploy` on this branch
* `kubectl logs -f <controller pod> --namespace deis`
* See if `/healtz` or `/readiness` show up in the logs (they should not)
* Do any `deis` CLI action and ensure those endpoints show up in the access log

If `DEIS_DEBUG` is set then the those requests should show

Fixes #252

### DIFF
--- a/rootfs/deis/gconf.py
+++ b/rootfs/deis/gconf.py
@@ -2,6 +2,7 @@ import os
 import faulthandler
 faulthandler.enable()
 
+
 bind = '0.0.0.0'
 try:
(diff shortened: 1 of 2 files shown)


sgoings:

PR #758, opened 2016-05-30
ref(gunicorn): unify gunicorn config and logging in a directory

helgi:
Depends on #757


sgoings:

PR #759, opened 2016-05-30
Move various pieces out of `deis` to `api`

helgi:
The separation between `deis` and the `api` component is not required anymore.

This PR moves a few of the pieces around to fix that and also namespaces all settings under `api.settings.<type>` for better distinction of what's what


sgoings:

PR #764, opened 2016-06-01
ref(image): handle Dockerfile build deps via variables

helgi:
Also speed up apt-get update by disabling source repos

If https://github.com/deis/docker-base/pull/2 gets merged then the optimisation can be removed from this repo

### DIFF
--- a/rootfs/Dockerfile
+++ b/rootfs/Dockerfile
@@ -9,28 +9,20 @@ RUN adduser --system \
 
 COPY . /app
 
-RUN apt-get update \
-	&& apt-get install -y \
-		gcc \
-		git \
-		libffi-dev \
-		libpq5 \
-		libpq-dev \
-		python3 \
-		python3-dev \
-		sudo \
-	&& ln -s /usr/bin/python3 /usr/bin/python \
-	&& curl -sSL https://bootstrap.pypa.io/get-pip.py | python - pip==8.1.2 \
-	&& mkdir -p /configs && chown -R deis:deis /configs \
-	&& pip install --disable-pip-version-check --no-cache-dir -r /app/requirements.txt \
-	&& apt-get remove -y --auto-remove --purge \
-		gcc \
-		git \
-		libffi-dev \
-		libpq-dev \
-		python3-dev \
-	&& apt-get clean \
-	&& rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/man /usr/share/doc
+RUN buildDeps='gcc git libffi-dev libpq-dev python3-dev'; \
+	apt-get update && \
+	apt-get install -y --no-install-recommends \
+        $buildDeps \
+        libpq5 \
+        python3 \
+        sudo && \
+	ln -s /usr/bin/python3 /usr/bin/python && \
+	curl -sSL https://bootstrap.pypa.io/get-pip.py | python - pip==8.1.2 && \
+	mkdir -p /configs && chown -R deis:deis /configs && \
+	pip install --disable-pip-version-check --no-cache-dir -r /app/requirements.txt && \
+	apt-get purge -y --auto-remove $buildDeps && \
+	apt-get clean && \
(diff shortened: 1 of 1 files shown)


rootfs/Dockerfile
```diff
@@ -9,28 +9,18 @@ RUN adduser --system \
...
-		python3-dev \
-	&& apt-get clean \
-	&& rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/man /usr/share/doc
+RUN buildDeps='gcc git libffi-dev libpq-dev libpq-dev python3-dev'; \
+	# disable source repos (speeds up apt-get update)
+	sed -i -e 's/^deb-src/#deb-src/' /etc/apt/sources.list && \
+	apt-get update && \
+	apt-get install -y $buildDeps python3 sudo --no-install-recommends && \
+	ln -s /usr/bin/python3 /usr/bin/python && \
+	curl -sSL https://bootstrap.pypa.io/get-pip.py | python - pip==8.1.1 && \
```
mboersma: `pip` 8.1.2 is current.

helgi: Yeah, but that is not what this PR is about (pip that is). 

8.1.2 has been the current for a long time and we haven't updated, we can do that as a separate PR.

helgi: https://github.com/deis/controller/pull/765 is the update PR for that

mboersma: Sorry @helgi, took my eye off the 🏈 there.

rootfs/Dockerfile
```diff
@@ -9,28 +9,18 @@ RUN adduser --system \
...
-	&& pip install --disable-pip-version-check --no-cache-dir -r /app/requirements.txt \
-	&& apt-get remove -y --auto-remove --purge \
-		gcc \
-		git \
-		libffi-dev \
-		libpq-dev \
-		python3-dev \
-	&& apt-get clean \
-	&& rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/man /usr/share/doc
+RUN buildDeps='gcc git libffi-dev libpq-dev libpq-dev python3-dev'; \
```
mboersma: `libpq-dev` is repeated here. And when I run this container I get:
```
django.core.exceptions.ImproperlyConfigured: Error loading psycopg2 module: libpq.so.5: cannot open shared object file: No such file or directory
```

I think `libpq5` needs to stay installed. (Maybe also `python3` and `sudo`, looking at the diff.)

helgi: `libpq5` being ommitted is a mistake. `python3` and `sudo` are kept since they are on the `apt-get install` line, I'm going to rejigger a bit for clarity.

Give it a go

mboersma:
I tested this with [example-dockerfile-http](https://github.com/deis/example-dockerfile-http) and [example-dockerfile-python](https://github.com/deis/example-dockerfile-python). :+1:

rootfs/Dockerfile
```diff
@@ -9,28 +9,22 @@ RUN adduser --system \
...
-		gcc \
-		git \
-		libffi-dev \
-		libpq-dev \
-		python3-dev \
-	&& apt-get clean \
-	&& rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/man /usr/share/doc
+RUN buildDeps='gcc git libffi-dev libpq-dev python3-dev'; \
+	# disable source repos (speeds up apt-get update)
+	sed -i -e 's/^deb-src/#deb-src/' /etc/apt/sources.list && \
```
krancour: If a new docker-base image does this already (i.e. after https://github.com/deis/docker-base/pull/2 is merged, I assume the `FROM` would be updated), wouldn't this line be redundant unless additional sources had been added (which in this case, they have not)?

helgi: Yeah, this PR is pretty much also a test for the docker-base one

PR #769, opened 2016-06-01
fix(scheduler): update instead of appending env vars so hardcoded values can be overwritten

### LINKED ISSUE #766: Unable to overwrite ENV Vars via deis config:set if they were originally set in a template
Example of this is if a user wants to overwrite `PORT` in a `buildpack` app or any of the Deis set variables in the RC templates.

Right now things are appended instead of merged. Danger here is if we fix this then people can break their apps by overwriting certain important values but flexibility wins over contrived.

https://github.com/deis/workflow/issues/275 has a lot of information

helgi:
# Summary of Changes

An example of this would be a user wanting to overwrite a hardcoded `PORT` or one of the other values Deis may set. Generally people should not do that, however the flexibility should be there in case they want to play with fire

if outputting the information for a given Pod this would be seen - 2x `PORT`
```
      env:
      - name: PORT
        value: "5000"
      - name: DEIS_APP
        value: finest-rosewood
      - name: PORT
        valueFrom:
          secretKeyRef:
            key: port
            name: finest-rosewood-v2-env
```

This should only show up once

Fixes #766

# Testing Instructions

1. Create a Deis Cluster
2. Register an app
3. push a buildpack app

2. `kubectl get pod --namespace=abcd` should show 1 pod running
3. Get pod information and make sure only one `PORT` definition is available

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -664,7 +664,7 @@ class KubeHTTPClient(object):
                 self._create_secret(namespace, secret_name, secrets_env, labels=labels)
 
             for key in env.keys():
-                data["env"].append({
+                item = {
                     "name": key,
                     "valueFrom": {
                         "secretKeyRef": {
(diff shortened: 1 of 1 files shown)


kmala:
does this fix https://github.com/deis/workflow/issues/275

helgi:
@kmala It addresses part of it but doesn't fix it - See the ticket this references

PR #771, opened 2016-06-01
fix(app): only add a PORT env var if there is a value

### LINKED ISSUE #767: PORT is being set regardless of what get_port returns for the Release object
When an application is not routable then `get_port` in the Release model returns `None`, however in the `App` model we use that return value regardless of what is there

https://github.com/deis/workflow/issues/275 has a lot of information on it and #766 plays a part in this as well

helgi:
# Summary of Changes

Only set the `PORT` env var if any was found.  By default it seems that k8s / docker will populate `$PORT` based on the service anyway unless we overwrite with our own value

Depends on #769 and #770

Fixes #767 
Fixes https://github.com/deis/workflow/issues/275

# Testing Instructions

It is best to follow the testing steps in the previous tickets and/or what @slack outlined in https://github.com/deis/workflow/issues/275

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -331,7 +331,9 @@ class App(UuidAuditedModel):
             # http://docs.deis.io/en/latest/using_deis/process-types/#web-vs-cmd-process-types
             routable = True if scale_type in ['web', 'cmd'] else False
             # fetch application port and inject into ENV Vars as needed
-            envs['PORT'] = release.get_port(routable)
+            port = release.get_port(routable)
+            if port:
+                envs['PORT'] = port
 
             kwargs = {
                 'memory': release.config.memory,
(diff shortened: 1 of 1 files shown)


slack:
Just tested #771 with image quay.io/jhansen/controller:git-ac7e296

```
$ kubectl --namespace=kabuki-duckling get secret kabuki-duckling-v3-env -o yaml
apiVersion: v1
data:
  port: NTAwMA==
  powered-by: Y29udHJvbGxlci9wdWxsLzc3MQ==
kind: Secret
metadata:
  creationTimestamp: 2016-06-01T23:19:53Z
  labels:
    app: kabuki-duckling
    type: env
    version: v3
  name: kabuki-duckling-v3-env
  namespace: kabuki-duckling
  resourceVersion: "162465"
  selfLink: /api/v1/namespaces/kabuki-duckling/secrets/kabuki-duckling-v3-env
  uid: 5882766b-284f-11e6-9aba-0a45229c9ef3
type: Opaque
```

Same basic worker type before. LGTM

PR #776, opened 2016-06-03
fix(scheduler): cast ports to int before passing them on to k8s

helgi:
when doing config:set PORT=80 then k8s complains that 80 is not a valid port / name

```
deis config:set PORT=80 -a zanier-zirconia
Creating config... Error:
400 Bad Request
detail: zanier-zirconia-v3-cmd (app::deploy): failed to create ReplicationController "zanier-zirconia-v3-cmd" in Namespace "zanier-zirconia": 422 status code 422
{'metadata': {}, 'kind': 'Status', 'reason': 'Invalid', 'status': 'Failure', 'message': 'ReplicationController "zanier-zirconia-v3-cmd" is invalid: spec.template.spec.containers[0].readinessProbe.tcpSocket.port: Invalid value: "80": must be an IANA_SVC_NAME (at most 15 characters, matching regex [a-z0-9]([a-z0-9-]*[a-z0-9])*, it must contain at least one letter [a-z], and hyphens cannot be adjacent to other hyphens): e.g. "http"', 'code': 422, 'apiVersion': 'v1', 'details': {'causes': [{'message': 'Invalid value: "80": must be an IANA_SVC_NAME (at most 15 characters, matching regex [a-z0-9]([a-z0-9-]*[a-z0-9])*, it must contain at least one letter [a-z], and hyphens cannot be adjacent to other hyphens): e.g. "http"', 'field': 'spec.template.spec.containers[0].readinessProbe.tcpSocket.port', 'reason': 'FieldValueInvalid'}], 'name': 'zanier-zirconia-v3-cmd', 'kind': 'ReplicationController'}}
```

Test is to run the above `config:set` and not get the error

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -464,7 +464,7 @@ class KubeHTTPClient(object):
                 for pos, item in enumerate(service['spec']['ports']):
                     if item['port'] == 80 and port != item['targetPort']:
                         # port 80 is the only one we care about right now
-                        service['spec']['ports'][pos]['targetPort'] = port
+                        service['spec']['ports'][pos]['targetPort'] = int(port)
 
             self._update_service(namespace, namespace, data=service)
         except Exception as e:
@@ -1181,7 +1181,7 @@ class KubeHTTPClient(object):
                 # an http probe
                 'httpGet': {
                     'path': path,
-                    'port': port
+                    'port': int(port)
                 },
                 # length of time to wait for a pod to initialize
                 # after pod startup, before applying health checking
@@ -1195,7 +1195,7 @@ class KubeHTTPClient(object):
                 # an http probe
                 'httpGet': {
                     'path': path,
-                    'port': port
+                    'port': int(port)
                 },
                 # length of time to wait for a pod to initialize
                 # after pod startup, before applying health checking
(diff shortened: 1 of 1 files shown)


bacongobbler:
sounds we do some validation that the port is an int first? For example I imaging I could break this via `deis config:set PORT=dog`

helgi:
We could but we never did test specific config options for the types - if you had done your example then k8is would throw a fit, with the PR it will just fail earlier

The main reason for the change is so k8s knows it's an int - otherwise it seems to see it as a string and wants to validate it as a name

bacongobbler:
I guess as long as the controller won't blow up and a 500 error is returned on a ValueError I'm fine with that.

helgi:
It will blow up with a 400 ;)

sstarcher:
I tested this out and with my app in the bad state I did the following and it came back up.
* deis config:set PORT=80 - the result was that the service was set to port 80, but the pods still health checked on the wrong port and failed
* deis config:unset HEALTHCHECK_PORT - svc and liveness probe were on the correct port
* App was functional and HEALTHCHECK_PORT is now gone for the first time ever ;)

helgi:
That's good to hear :-) Curious why the `unset` was required, will keep an eye out on that one!

party on

sstarcher:
@helgi I don't know if it was the unset that fixed it or if it was just forcing it to kick off a second build, but certainly going from the broken state to setting the PORT did not bring it up correctly.  It may of worked if I randomly changed another config or did a fresh pull.

PR #783, opened 2016-06-04
fix(logs): retry fetching logs from the logger 3 times if the services is unavailable

### LINKED ISSUE #676: connection between logger and controller seems flaky
Try enough times and you get this:

```
><> deis logs
INFO [go]: domain go removed
INFO [go]: bacongobbler created initial release
INFO [go]: domain go added

><> deis logs
INFO [go]: domain go removed
INFO [go]: bacongobbler created initial release
INFO [go]: domain go added

><> deis logs
Error: There are currently no log messages. Please check the following things:
1) Logger and fluentd pods are running.
2) The application is writing logs to the logger component by checking that an entry in the ring buffer was created: kubectl logs <logger pod> --namespace=deis
3) Making sure that the container logs were mounted properly into the fluentd pod: kubectl exec <fluentd pod> --namespace=deis ls /var/log/containers
><> deis logs
INFO [go]: domain go removed
INFO [go]: bacongobbler created initial release
INFO [go]: domain go added

```

helgi:
Using `backoff` instead of `retrying` since it offers the ability (built in) to retry on certain exceptions and is also still maintained

Fixes #676

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -1,3 +1,4 @@
+import backoff
 from collections import OrderedDict
 from datetime import datetime
 import logging
@@ -541,6 +542,7 @@ class App(UuidAuditedModel):
             level=logging.DEBUG
         )
 
+    @backoff.on_exception(backoff.expo, ServiceUnavailable, max_tries=3)
     def logs(self, log_lines=str(settings.LOG_LINES)):
         """Return aggregated log data for this application."""
         try:
--- a/rootfs/requirements.txt
+++ b/rootfs/requirements.txt
@@ -1,4 +1,5 @@
 # Deis controller requirements
+backoff==1.2.1
 Django==1.9.7
 django-cors-headers==1.1.0
 django-guardian==1.4.4


rootfs/api/models/app.py
```diff
@@ -541,6 +542,7 @@ def verify_application_health(self, **kwargs):
             level=logging.DEBUG
         )
 
+    @backoff.on_exception(backoff.expo, ServiceUnavailable, max_tries=3)
```
kmala: i think you can use @retry(retry_on_exception= ServiceUnavailable, stop_max_attempt_number=3)

helgi: Yeah, I specifically decided not to use the `retrying` lib as it has not had a release since 2014 where as `backoff` was released a little bit ago. See https://github.com/deis/controller/issues/708

kmala: may be we can use backoff in all places instead of retrying

helgi: Yeah, was going to bring it up tomorrow but now the cat is out of the bag!

PR #787, opened 2016-06-06
feat(config): validate PORT and HEALTHCHECK_* values for config:set operations

### LINKED ISSUE #774: HEALTHCHECK_URL url encoding
When supplying a HEALTHCHECK_URL to deis config:set that contains a ```?``` it gets encoded when past into the Kubernetes liveness probe.

Steps:
* deis config:set HEALTHCHECK_URL=/servlet/status?test_datasources=0
* deis config:get HEALTHCHECK_URL - shows the value as not encoded
* kubectl describe pod --namespace=app APP - shows the value as encoded and the health check never passes
```http-get http://:8080/servlet/status%3Ftest_datasources=0 delay=120s timeout=120s period=10s #success=1 #failure=10```

### LINKED ISSUE #777: Validate that PORT config option is in fact an int
Right now it is possible to do `PORT=dog` and we'll happily pass that along everywhere. While `PORT` is in the very broad `config:set` space then we can still do some special validation and throw a nice error message

See https://github.com/deis/controller/pull/776 for more info

### LINKED ISSUE #781: Validate all health check inputs
https://github.com/deis/controller/issues/774 is an example where `URL` has to be validated as not containing query parameters

Most of the others are a numerical validation

helgi:
# Summary of Changes

This validates that the `PORT` is an int and within the acceptable port range.

`HEALTHCHECK_URL` has a special validation due to Kubernetes constraints we found in #774 - Basically only supports `path`

Please list the issue(s) that this PR closes, similar to the below:

Fixes #774
Fixes #777
Fixes #781

https://github.com/deis/workflow/pull/286

# Testing Instructions

The Controller tests take care of this

### DIFF
--- a/rootfs/api/serializers.py
+++ b/rootfs/api/serializers.py
@@ -4,6 +4,7 @@ Classes to serialize the RESTful representation of Deis API models.
 
 import json
 import re
+from urllib.parse import urlparse
 
 from django.contrib.auth.models import User
 from django.utils import timezone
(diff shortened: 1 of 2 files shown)


bacongobbler:
validation's always nice

PR #789, opened 2016-06-06
feat(release): throw 409 when identical release is done sequantially

### LINKED ISSUE #321: Applying the same tag/config + value = new release
Same tag/value can be applied many times and each one is a new release but shows up as "helgi changed nothing"

This means kicking off a new release with all the associated building/pulling/pushing and all other operations k8s does.

Should be detected and be a noop. no release.

helgi:
# Summary of Changes

This replaces logging that nothing changed for the release and still going ahead with the release. Now it is a noop operation.

Fixes #321

# Testing Instructions

1. Create a Deis Cluster
2. Register an app
3. `deis pull deis/example-go -a <app>
4. `deis config:set foo=bar -a <app>`
5. `deis config:set foo=bar -a <app>`

A `409` response should be seen

### DIFF
--- a/rootfs/api/models/release.py
+++ b/rootfs/api/models/release.py
@@ -5,7 +5,8 @@ from django.db import models
 
 from registry import publish_release, get_port as docker_get_port, RegistryException
 from api.utils import dict_diff
-from api.models import UuidAuditedModel, DeisException
+from api.models import UuidAuditedModel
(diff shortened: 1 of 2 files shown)


helgi:
I need to investigate e2e a bit better on this one now that I have e2e working locally

helgi:
This depends on #826

PR #790, opened 2016-06-06
fix(scale):get the desired number of replicas from the rc

### LINKED ISSUE #756: If pods are in crashloop you cant scale down
```
╰─± kgpo --namespace=foo
NAME               READY     STATUS             RESTARTS   AGE
foo-v2-web-ouxev   0/1       Error              0          3s
foo-v2-web-rcy89   0/1       CrashLoopBackOff   1          11s
foo-v2-web-tdl9g   1/1       Running            0          11h
╭─jonathanchauncey at ENG000637 in ~/projects/deis/example-go on master✔ using ‹2.2.2›
╰─± deis scale web=1
done in 0s
=== foo Processes
--- web:
foo-v2-web-ouxev error (v2)
foo-v2-web-rcy89 error (v2)
foo-v2-web-tdl9g up (v2)
```

kmala:
# Summary of Changes

Get the desired number of replicas from the rc instead of counting.

closes #756 

# Testing Instructions

1. Create a Deis Cluster
2. Deploy an app
3. Now kill the minio pod so that the image is not present
4. scale the rc so that they are in crash loop back status(this is bug which will be fixed in another PR) and now if you scale it does nothing.
5. after the fixes when you scale down it should do it.

# Pull Request Hygiene TODOs

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -1004,12 +1004,7 @@ class KubeHTTPClient(object):
             'version': rc['spec']['selector']['version']
         }
 
-        # Are there any pods running (and verified as ready) available?
-        pods = self._get_pods(namespace, labels=labels).json()['items']
-        current = 0
-        for pod in pods:
-            if self._pod_ready(pod):
-                current += 1
+        current = int(rc['spec']['replicas'])
 
         if desired == current:
             logger.info("Not scaling RC {} in Namespace {} to {} replicas. Already at desired replicas".format(name, namespace, desired))  # noqa


rootfs/scheduler/__init__.py
```diff
@@ -1004,12 +1004,7 @@ def _scale_rc(self, namespace, name, desired):
             'version': rc['spec']['selector']['version']
         }
 
-        # Are there any pods running (and verified as ready) available?
-        pods = self._get_pods(namespace, labels=labels).json()['items']
-        current = 0
-        for pod in pods:
-            if self._pod_ready(pod):
-                current += 1
+        current = int(rc['spec']['replicas'])
```
helgi: This is not going to cut it - This is looking at how much we have available, not how many we want.

By doing this you are basically saying that `current = desired`

kmala: I din't get what you meant by current = desired....
rc['spec']['replicas'] gives the current desired number of replicas and when we do a deis scale the value we give is the desired number of replicas we want and if both are not equal we scale otherwise we don't

helgi: The `desired` value is already passed in, which is the value that is saved as `rc['spec']['replicas']`

See https://github.com/deis/controller/blob/master/rootfs/scheduler/__init__.py#L1019 and then https://github.com/deis/controller/blob/master/rootfs/scheduler/__init__.py#L1050

Your code is then doing `current = rc['spec']['replicas']` which is just the same value as `desired` will already be using

helgi: Maybe you are looking for `rc['status']['replicas']` ? That will give the current state, however that may be deceiving as well since we are specifically looking for pods that are ready

kmala: the issue is happening at https://github.com/deis/controller/blob/master/rootfs/scheduler/__init__.py#L1015 which happens before the code we are setting the desired and updating the rc
here the issue is getting the value of the current before doing an update

helgi: Ah right, that's true. However that is still not going to work. We need to figure out if there are *working* pods ready, not how many pods the RC thinks it should have (that's the difference between `spec/replicas` and `status/replicas`)

kmala: when i do deis scale web/cmd=<some number> i meant that the number is the desired number of replicas i want eventually which is spec/replicas.
whereas status/replicas is observed number of replicas at that time....which can be different during the scheduling of the pod but will eventually be same as spec/replicas

helgi: Correct but look at bb844717170d355a6e75ebe371a5efd090defd88 and 22feb4478f4a0c0b9c7e1d751448ba2a5dbd44cd which are the progression behind why `_scale_rc` is the way it is.

I just want to be very cautious about what we are doing in `_scale_rc`

kmala: i saw those PR's and it was mentioned in #675 as it poses problems for deploys but i don't see any reason why that would pose a problem.We need to check whether the pods are ready or not after the scale and which we are doing.

Issue #785, opened 2016-06-05
Controller Scaling pods out of service without health pods


sstarcher:
`deis limits:set --cpu cmd=200000`

Will of course cause the pods to not be able to schedule.  Upon doing this the controller waits for the new pods to become healthy and after the timeout, while the pods are still pending it will shutdown the functioning pods.

```
INFO [APP]: config APP-1b1f384 updated
INFO [APP]: sstarcher changed nothing
INFO [APP]: quay.io/image/image:latest exists in the target registry. Using image for release 32 of app APP
INFO deploy APP-v32-cmd, img quay.io/image/image:latest, cmd ""
INFO scaling release APP-v32-cmd to 4 out of final 4
INFO adding 30s on to the original 120s timeout to account for the initial delay specified in the readiness probe
INFO waiting for 4 pods in APP namespace to be in services (150 timeout)
INFO waited 20s and 0 pods are in service
INFO waited 10s and 0 pods are in service
INFO waited 30s and 0 pods are in service
INFO waited 20s and 0 pods are in service
INFO waited 40s and 0 pods are in service
INFO waited 30s and 0 pods are in service
INFO waited 50s and 0 pods are in service
INFO waited 40s and 0 pods are in service
INFO waited 60s and 0 pods are in service
INFO waited 50s and 0 pods are in service
INFO waited 70s and 0 pods are in service
INFO waited 60s and 0 pods are in service
INFO waited 80s and 0 pods are in service
INFO waited 70s and 0 pods are in service
INFO waited 90s and 0 pods are in service
INFO waited 80s and 0 pods are in service
INFO waited 100s and 0 pods are in service
INFO waited 90s and 0 pods are in service
INFO waited 110s and 0 pods are in service
INFO waited 100s and 0 pods are in service
INFO waited 120s and 0 pods are in service
INFO waited 110s and 0 pods are in service
INFO waited 130s and 0 pods are in service
INFO waited 120s and 0 pods are in service
INFO waited 140s and 0 pods are in service
INFO waited 130s and 0 pods are in service
INFO waited 150s and 0 pods are in service
INFO waited 140s and 0 pods are in service
INFO waited 160s and 0 pods are in service
INFO 0 out of 4 pods in namespace APP are in service
INFO scaling old release APP-v31-cmd from original 4 to 0
INFO Not scaling RC APP-v31-cmd in Namespace APP to 0 replicas. Already at desired replicas
[...]

PR #792, opened 2016-06-07
fix(scheduler): rollback scale / deploy when the desired number of pods can not be brought up in a timely manner

### LINKED ISSUE #706: Failed health checks still deploy
_From @sstarcher on May 3, 2016 17:19_

I created an app that without proper configuration would fail health checks.  The following steps were followed.

* Deis create app
* git push deis
* First version of app is deployed and in a crashback loop as expected
* deis config:set HEALTHCHECK_URL=/
* Second app version deployed and in a crashback loop as expected.  At this point ```kubectl get rcs --namespace=app``` shows 2 app versions are in a crashback loop
* deis config:set rest of data to make app function
* Third app version claims it has passed the health check, but the app takes several minutes to start up.  At this point the deis router routes to the app and gives a 503 even though the deis command line shows the app is good.  ```deis info``` shows version 3 of the app is up
* Third version of the app finally comes up after several minutes and the router routes correctly to the app.
* deis config:set is used to break the application configuration.  
* Fourth version of the app comes up and deis believes it has passed the health check.  Deis removes the third version of the app and now the router is back to a 503.

End result is
```
NAME                 READY     STATUS             RESTARTS   AGE
hound-v1-cmd-2fe47   0/1       CrashLoopBackOff   20         1h
hound-v2-cmd-qggwg   0/1       Error              5          33m
hound-v4-cmd-dw6u6   0/1       CrashLoopBackOff   6          9m
```

Old RCs are still hanging around and the app is down due to the third version being deleted prior to the 4th version passing health checks.

deis config:list
```
HEALTHCHECK_FAILURE_THRESHOLD      3
HEALTHCHECK_INITIAL_DELAY          1080
HEALTHCHECK_PERIOD_SECONDS         10
HEALTHCHECK_PORT                   5000
HEALTHCHECK_SUCCESS_THRESHOLD      1
HEALTHCHECK_TIMEOUT                400
HEALTHCHECK_URL                    /
```

_Copied from original issue: deis/workflow#214_

helgi:
# Summary of Changes

Prior to this if a deploy failed at bringing up pods then it would still scale down the old release instead of rolling back. This would cause users to basically have a broken app if the new pods for whatever reason stick around in Pending mode

If the timeout period is not long enough for users then increasing the `HEALTHCHECK_INITIAL_DELAY` is the way to go.

Fixes #706
Refs #785

# Testing Instructions

1. Create a Deis Cluster
2. Register an app
4. `deis limits:set -c cmd=2000` (this should fail)
5. `deis config:set foo=bar` (this should succeed)
6. `deis limits` (this should show Unlimited and not 2000`

2. `k get po --namespace <app>` should still show v2 and not v3
3. Log should contain the following:
```
INFO waited 110s and 0 pods are in service
INFO waited 120s and 0 pods are in service
INFO 0 out of 1 pods in namespace quartz-zoologer are in service
INFO Not scaling RC quartz-zoologer-v3-cmd in Namespace quartz-zoologer to 0 replicas. Already at desired replicas
INFO Not scaling RC quartz-zoologer-v2-cmd in Namespace quartz-zoologer to 1 replicas. Already at desired replicas
ERROR [quartz-zoologer]: quartz-zoologer-v3-cmd (app::deploy): Could not scale quartz-zoologer-v3-cmd to 1. Deleting and going back to old release
10.244.1.10 "POST /v2/apps/quartz-zoologer/config/ HTTP/1.1" 400 134 "Deis Client v2.0.0-dev"
```

While it doesn't show the Exception message that I added in this PR it is one of the higher level messages in the `scheduler::deploy`, and the application is not town down

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -1010,7 +1010,11 @@ class KubeHTTPClient(object):
         if waited > timeout:
             self.log(namespace, 'timed out ({}s) waiting for pods to come up in namespace {}'.format(timeout, namespace))  # noqa
 
-        self.log(namespace, "{} out of {} pods are in service".format(count, desired))  # noqa
+        self.log(namespace, "{} out of {} pods are in service".format(count, desired))
+        if count != desired:
+            # raising to allow operations to rollback
+            raise KubeException('Not enough pods in namespace {} came into service. '
+                                '{} out of {}'.format(namespace, count, desired))
 
     def _scale_rc(self, namespace, name, desired, timeout):
         rc = self.get_rc(namespace, name).json()


rootfs/scheduler/__init__.py
```diff
@@ -993,6 +993,10 @@ def _wait_until_pods_are_ready(self, namespace, container, labels, desired):  #
             logger.info('timed out ({}s) waiting for pods to come up in namespace {}'.format(timeout, namespace))  # noqa
 
         logger.info("{} out of {} pods in namespace {} are in service".format(count, desired, namespace))  # noqa
+        if count != desired:
+            # raising to allow operations to rollback
+            raise KubeException('Not enough pods in namespace {} came into service. '
+                                '{} out of {}'.format(namespace, count, desired))
```
kmala: it would be better if we can get the reason why the pod din't come to service than just saying not enough pods

helgi: Yeah, however in the current setup this is not bubbled up and rather a message in the deploy function is used.

I can look into improving the message but I didn't want to change the deploy command since it handles more than just this failure

helgi: @kmala do you have any suggestions what information we could bubble up? Note that this can be more than one pod that can fail as we do deploy batches now

kmala: If the pods are in a crash loop back/if they are atleast started the pods logs can be anyway found using the deis logs....i just want to see the possibility if we can get the reason for pods not starting in  other scenarios may be getting it from the events.The reason we might want to show is because users may be in confused state as in why the release or scale failed.
Also i have one more question regarding this change....do we want to think the operation as failure even if just the readiness probe failed because it might be waiting for some external dependency?

helgi: Okay, that's fair. Most people will be able to use `deis ps` to do diagnostics but maybe there is a way to get more info into the Exception. Only issue could be overloading the Exception message with a lot of information since you might be deploying a batch of 10 Pods and they all fail.

Before I (personally) didn't see it as a failure but as seen in #706 and #785 that not treating this as a failure is causing problems. The scenario here is this:

I have many releases done via Deis Workflow, I'm at v28 - I do some config setting and v29 kicks off. The Pods for v29 never come up and the timeout kicks in (when waiting for the pods to be ready) - After the timeout the deploy system is made to just go start scaling down v28

At the end of all of that I lost v28 and v29 is all in a Pending state. Now my business is down.

Because of the above I've started considering this a failure state. If your readiness probe is still waiting for external dependencies then you should probably increase `HEALTHCHECK_INITIAL_DELAY` - I have also considered making the 2 minute wait for pod timeout as something people can configure, in addition to the `INITIAL_DELAY`

Does that help explain why I'm opting for failure mode?

kmala: we definitely should opt for failure mode when it is actually fails to starts.....both those bugs show the scenario of pods starting to fail or liveness probe failing which are failing modes...my question is regarding the readiness check where the pod actually starts running but isn't just ready to serve requests(this is definitely not ready but i don't think a failure) also we should have separate liveness check and readiness check instead of single health check
My point is it would be better if we have discussion on what is considered a failure mode though considering all scenarios as failure modes isn't gonna harm anything

helgi: That's a good point - Until we have both liveness and readiness as independently configurable then we will have to consider failure in either as a failure mode.

Even if we let people configure both then I think it is a bit of a problem if your readiness is still working but is beyond the timeout, we can't really start spinning working stuff down until we have enough pods to keep the app running. That's the biggest issue - Suppose we could add time to the timeout when we detect it is in an almost ready state, but only up to a point

sstarcher:
* deis limits:set -c cmd=2000
```
Applying limits... Error:
400 Bad Request
detail: public-v9-cmd (app::deploy): Could not scale public-v9-cmd to 1. Deleting and going back to old release
```

Deis public-v8 was still deployed as expected

* deis limits
```
=== public Limits

--- Memory
Unlimited

--- CPU
Unlimited
```

Also what I would expect since limits:set 2000 failed

* ```deis config:set HELLO=test``` - to cause an additional build

This additional deployment still contains the 
```
    resources:
      limits:
        cpu: 2k
      requests:
        cpu: 2k
```

rootfs/scheduler/__init__.py
kmala: I just saw that we hard coding the timeout to 2 minutes https://github.com/deis/controller/pull/792/files#diff-3b6e1f09121075ad67f216aa2bbfc9a6L942
shouldn't that be configurable as some deploys take more than 2 mins?

helgi: That's something that has been around for a longgggg time (probably v1 days) - That's just the waiting aspect for each *scale* operation, of which can be many in a single deploy. So far `HEALTHCHECK_INITIAL_DELAY` can extend that timeout tho. So if you have a delay of 120 seconds then it is a total of 240 seconds timeout per scale operation.

kmala: yes i know that it has been hardcoded since its inception but now that we are starting to fail the operation based on the timeout i think we should start allowing the user to configure it failing which it might become problem for them....since HEALTHCHECK_INITIAL_DELAY is not mandatory to be set for every pod we shouldn't depend on it

helgi: I agree on doing that but lets do it as part of a different PR. Can you open a PR and I will tackle it this week?

bacongobbler:
This concerns me a little considering that we didn't error out if a `deis scale` failed before due to pending pods. I'd be okay if we failed knowing that the pod's boot time is taking longer than expected, but what if the container is really big (i.e. ruby+python container) and just takes a long time to be deployed on a worker node? Is there a way we can differentiate between the two situations?

helgi:
Right now the Controller will add 10 minutes if your image is a slow downloading image, it does that after 1 minute of downloading, this we can also make configurable (tl;dr; yes we can understand pretty well where in the process your are stalling), also you have `HEALTHCHECK_INITAL_DELAY` to affect the timeouts even more (but that's applied after a download) if your app will take a long time to come up.

Thing is, if we don't do this, you may end up nixing a working app in favour of bunch of Pending pods. This change would kill the "in flight" RC / Pods and never put them live.

helgi:
I've created https://github.com/deis/controller/issues/817 to address the configurable timeout aspect

bacongobbler:
this PR has 2 LGTMs, just needs to pass CI to merge.

helgi:
Due to a few of the e2e tests being a bit sensitive around pod timeout limits then this PR is not really working out for RCs

https://github.com/deis/controller/pull/942 will most likely replace this one when I pull the trigger on Deployments being on by default

helgi:
Turns out Deployments may not fully solve this but only because I ended up getting this

> Applying limits... Error: Post http://deis.helgi.engineyard.com/v2/apps/bamboo-ladybird/config/: EOF

That left a Deployment hanging around and the following `config:set` and related commands won't work because we have a "Deployment in progress" protection - I need to figure out something here

helgi:
Closing this in favour of https://github.com/deis/controller/pull/978

PR #793, opened 2016-06-07
chore(version): update platform version to 2.0.0

helgi:


bacongobbler:
I was looking for this the other day. Turns out I was just looking at the API version. LGTM

PR #794, opened 2016-06-07
ref(gunicorn): use gaiohttp workers

mboersma:
# Summary of Changes

Replaces the default `sync` [gunicorn worker class](http://docs.gunicorn.org/en/latest/settings.html#worker-class) with `gaiohttp`. This [async worker](http://docs.gunicorn.org/en/latest/design.html#asyncio-workers) class is based on `aiohttp` for Python 3. It may offer better responsiveness and lower CPU load.

My preliminary testing shows it holds up well and uses less CPU, but I'll produce more specific benchmark results to justify this soon.

TODO:
- [ ] produce more specific benchmark results
- [x] continue long-term testing to reassure myself about stability of this change

# Testing Instructions

There are no specific testing instructions, since this should be an invisible backend change. However, monitoring for response times and CPU load over a long period of time would assure this does not introduce new problems.

# Pull Request Hygiene TODOs

### DIFF
--- a/rootfs/deis/gunicorn/config.py
+++ b/rootfs/deis/gunicorn/config.py
@@ -11,7 +11,7 @@ try:
         raise ValueError()
 except (NameError, ValueError):
     workers = (os.cpu_count() or 4) * 2 + 1
-
+worker_class = 'gaiohttp'
 pythonpath = dirname(dirname(dirname(realpath(__file__))))
 timeout = 1200
 pidfile = '/tmp/gunicorn.pid'
--- a/rootfs/requirements.txt
+++ b/rootfs/requirements.txt
@@ -1,4 +1,5 @@
 # Deis controller requirements
+aiohttp==0.21.6
 backoff==1.2.1
 Django==1.9.7
 django-cors-headers==1.1.0


bacongobbler:
>  - Supports both client and server Web-Sockets out-of-the-box.

This could be game changing. We've been looking for an websocket-compatible HTTP server that works with django for a while. This could mean that we could potentially support `deis logs -f`, right?

helgi:
Should be able to, tho the alternative could have always been to have a logger endpoint (public) we could pass on the connection to the end user which could allow for it instead of piping via the API

jgmize:
While benchmarking, it might be interesting to also try out https://github.com/MagicStack/uvloop via the implementation in https://github.com/KeepSafe/aiohttp/pull/878. See http://magic.io/blog/uvloop-blazing-fast-python-networking/ for some existing benchmarks.

mboersma:
> it might be interesting to also try out https://github.com/MagicStack/uvloop...

I did spend some time using the `KeepSafe/aiohttp` tip and trying to get it working as advertised, but ran into a variety of configuration issues I couldn't get past. I love the benchmarking numbers on `uvloop`, so let's revisit this when it's had time to mature. Hopefully just this `gaiohttp` change will help.

mboersma:
This seems promising, especially when paired with the potential of `uvloop`. But my `boom` and workflow-e2e test benchmarks have shown it doesn't perform quite as well as the trusty gunicorn `sync` worker.

So I'm closing this, but it would be smart for someone to revisit this territory in the future after #833 or #835 are (probably) merged.

PR #797, opened 2016-06-07
feat(django): explicitically define on_delete for ForeignKey fields for 2.0 compat

helgi:
Also define all owner fields as PROTECT instead of delete see https://docs.djangoproject.com/en/1.9/ref/models/fields/#django.db.models.ForeignKey

https://docs.djangoproject.com/en/1.9/releases/1.9/#foreignkey-and-onetoonefield-on-delete-argument


helgi:
I changed it so that `Key` is deleted when a user is deleted. Reason is `User` and `Key` are the only thing really properly connected in a 1:1 way

PR #808, opened 2016-06-10
fix(serializer): skip None values in config and registry

helgi:
Other config items use this and there was a probelm where config:unset was getting caught on trying to validate that PORT was numeric


slack:
worth a unit test?

helgi:
Yeah, good point - This won't be merged until e2e is better (caught it via that work) so I'll add a unit test in here first

slack:
Cool, LGTM pending unit test.

helgi:
Added a test specifically testing `PORT`, registry isn't doing any value validation right now

PR #812, opened 2016-06-10
ref(scheduler): move a few high level scheduling functions to the App model

helgi:
Just some housekeeping. It's logic more pertaining to the App model than the scheduler

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -151,7 +151,7 @@ class App(UuidAuditedModel):
         """
         logger.log(level, "[{}]: {}".format(self.id, message))
 
-    def create(self, *args, **kwargs):
+    def create(self, *args, **kwargs):  # noqa
         """
         Create a application with an initial config, release, domain
         and k8s resource if needed
@@ -171,9 +171,28 @@ class App(UuidAuditedModel):
             )
 
         # create required minimum resources in k8s for the application
+        namespace = self.id
+        service = self.id
         try:
-            self._scheduler.create(self.id)
+            self.log('creating Namespace {} and services'.format(namespace), level=logging.DEBUG)
+            # Create essential resources
+            try:
+                self._scheduler.get_namespace(namespace)
+            except KubeException:
+                self._scheduler.create_namespace(namespace)
+
+            try:
+                self._scheduler.get_service(namespace, service)
+            except KubeException:
+                self._scheduler.create_service(namespace, service)
         except KubeException as e:
+            # Blow it all away only if something horrible happens
+            try:
+                self._scheduler.delete_namespace(namespace)
+            except KubeException as e:
+                # Just feed into the item below
+                pass
+
             raise ServiceUnavailable('Kubernetes resources could not be created') from e
 
         # Attach the platform specific application sub domain to the k8s service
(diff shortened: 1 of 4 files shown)


rootfs/api/models/app.py
```diff
@@ -171,10 +171,27 @@ def create(self, *args, **kwargs):
...
 
         # create required minimum resources in k8s for the application
+        namespace = self.id
+        service = self.id
         try:
-            self._scheduler.create(self.id)
+            self.log('creating Namespace {} and services'.format(namespace), level=logging.DEBUG)
+            # Create essential resources
+            try:
+                self._scheduler._get_namespace(namespace)
```
bacongobbler: technically speaking we shouldn't be using private functions (functions with `_` before the name) outside of the package; only public functions. If we want to migrate these over, we should make `_get_namespace`, `_create_namespace`, etc. as public functions.

helgi: Yeah, however we've used scheduling code in the models (like that) for a few months but mostly due to one of my "rewrite the scheduler" PRs ended up in a disaster.

I'll do a diff PR that opens that all up and will expose all `_get`, `_create`, `_update` and `_delete` as I want to have the models interacting with the scheduling client a lot more (like this PR)

bacongobbler:
looks like travis is failing on the unit tests for legit reasons.

helgi:
@bacongobbler indeed - hence the `in progress` label

kmala:
I feel those should still be in scheduler as creating of namespace and service are related to scheduler and not app.

helgi:
I'm looking at it from the aspect that the Business logic (our models) should own the fact we are weaving together multiple Kubernetes resources and the Scheduler should just be a client interface to Kubernetes

Technically I could make `client.py` for those interactions and keep `scheduler` owning the very few interactions that are not pure "I am talking to the k8s API" but I have not felt the need just yet

helgi:
Another way to look at it: If we decide to stop creating `Services` up front when an application is created (vs deployed) then `create` in the `scheduler` would only have the `Namespace` logic. Makes for a very tiny function

kmala:
Yes..i agree the should be in the Business logic (our models) if we are thinking it(init.py) as client.py and not scheduler.py but as per me we still assume it to be scheduler as we do a lot business logic in it and not just client because of which i think the logic shouldn't be in app model how much ever the code might be.

helgi:
The scheduler has a lot of legacy from when we did many schedulers, to me this would be one of those.

The business logic shouldn't be part of the scheduler, as much as possible. There will be cases where it has to be but I don't count this as one of those

rootfs/api/models/app.py
```diff
@@ -185,8 +204,14 @@ def delete(self, *args, **kwargs):
         """Delete this application including all containers"""
         self.log("deleting environment")
         try:
-            # attempt to remove application from kubernetes
-            self._scheduler.destroy(self.id)
+            self._scheduler.delete_namespace(self.id)
+
+            # wait 30 seconds for termination
```
mboersma: Does this actually wait for 30 seconds or just try 30 times?

helgi: little bit of column A, little bit of column B

It's 30 seconds + whatever latency each request adds :) each place has a `1` second sleep. I mostly copied this over from the scheduler but I can change the language if it is something we care about

PR #813, opened 2016-06-11
ref(scheduler): move scheduler templates from JSON to YAML

helgi:
Reasoning behind this is everyone looks at k8s information via YAML. It is much easier to read

Uses https://yaml.readthedocs.io/en/latest/index.html instead of PyYAML as it is basing off it anyway

Decided not to change the `string.Template` approach, yet.

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -10,6 +10,7 @@ import base64
 from django.conf import settings
 from docker.auth import auth as docker_auth
 from .states import PodState
+import ruamel.yaml
 import requests
 from requests_toolbelt import user_agent
 from .utils import dict_merge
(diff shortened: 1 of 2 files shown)


bacongobbler:
Won't this break v2.0.0 apps, or is the upgrade path ok with this change? I guess since we are fetching the pod manifest from the kube API it wouldn't affect upgrades, right?

helgi:
Yeah this is totally internal, no affect on apps, we still convert the yaml to dict and post it as json

This only changes our base templates to YAML

rootfs/scheduler/__init__.py
```diff
@@ -21,277 +22,183 @@
...
+  labels:
+    app: $id
+    version: $appversion
+    type: $type
+    heritage: deis
+spec:
+  replicas: $replicas
+  selector:
+    app: $id
+    version: $appversion
```
slack: This will always be `v<INT>` yeah?

helgi: Yeah, it will

slack:
LGTM, looks like all the spots where things may be inty when we were getting JSON strings are appropriately quoted in the YAML form.

helgi:
I had to quote (turn into a string) the minio port since k8s requires env vars to come in as strings and the objectkeyfile mounts due to the dash (`-`)

PR #814, opened 2016-06-13
ref(scheduler): ditch error() in favour of using KubeHTTPException

helgi:


rootfs/api/tests/test_release.py
```diff
@@ -320,11 +320,13 @@ def test_release_rollback_failure(self, mock_requests):
...
                 # instead of full request mocking, fake it out in a simple way
                 class Response(object):
-                    pass
+                    def json(self):
+                        return '{}'
 
                 response = Response()
                 response.status_code = 404
-                kube_exception = KubeHTTPException(response=response)
+                response.reason = "Not Found"
```
bacongobbler: can't we just return a `django.http.HttpResponseNotFound` instead to simplify this code?

helgi: Ah! hadn't thought of that but I did give it a go just now and it won't work because of `AttributeError: 'HttpResponseNotFound' object has no attribute 'reason'` 

`KubeException` is expecting a `python-requests` response object

PR #815, opened 2016-06-13
ref(scheduler): use observedGeneration instead of resourceVersion for RC readiness

helgi:
https://github.com/kubernetes/kubernetes/blob/master/docs/devel/api-conventions.md#metadata explains the resourceVersion and generation pretty well

resourceVersion is the internal version representation where as generation deals with desired state

Some resources report the observedGeneration, which is the generation most recently observed by the component responsible for acting upon changes to the desired state of the resource. This can be used, for instance, to ensure that the reported status reflects the most recent desired status.

This is also how to determine if Deployments are ready so it's good to match it all up


mboersma:
Thanks for linking to the relevant docs. 👍

PR #818, opened 2016-06-14
ref(scheduler): move a lot of scheduler functions from private to public

helgi:
Many were already being used in the Models - Time to make it more official


helgi:
e2e passed, there is simply some error in there unrelated to the actual ginkgo tests

PR #820, opened 2016-06-16
ref(scheduler): introduce a build pod template function to reduce duplication in scheduling code

helgi:
ReplicationController can take a fully formed Pod Template (include Kind / apiVerison) within its spec/template section

This allows reuse of the Pod Template between deis run and RCs (and Deployments in the future).

The distinction between buildpack and "the rest" templates was eliminated by moving more of the env var and volume logic into the build pod template and set container functions.

This may not seem like a lot but it will help reduce code duplication (and template duplication) going forward, especially when I add Deployments

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -20,157 +20,10 @@ from deis import __version__ as deis_version
 
 logger = logging.getLogger(__name__)
 
-# Used for one off command runs on pods
-POD_BTEMPLATE = """\
-kind: Pod
-apiVersion: $version
-metadata:
-  name: $id
-  labels:
-    app: $app
-    version: $appversion
-    type: $type
-    heritage: deis
-spec:
-  containers:
-    - name: $containername
-      image: $image
-      env:
-        - name: SLUG_URL
-          value: $slug_url
-        - name: BUILDER_STORAGE
-          value: $storagetype
-        - name: DEIS_MINIO_SERVICE_HOST
-          value: $mHost
-        - name: DEIS_MINIO_SERVICE_PORT
-          value: "$mPort"
-      volumeMounts:
-        - name: objectstorage-keyfile
-          mountPath: /var/run/secrets/deis/objectstore/creds
-          readOnly: true
-  volumes:
-    - name: "objectstorage-keyfile"
-      secret:
-        secretName: objectstorage-keyfile
-  terminationGracePeriodSeconds: $terminationGracePeriodSeconds
-  restartPolicy: Never
-"""
-
-POD_TEMPLATE = """\
-kind: Pod
-apiVersion: $version
-metadata:
-  name: $id
-  labels:
-    app: $app
-    version: $appversion
-    type: $type
-    heritage: deis
-spec:
-  containers:
-    - name: $containername
-      image: $image
-      env: []
-  terminationGracePeriodSeconds: $terminationGracePeriodSeconds
-  restartPolicy: Never
-"""
-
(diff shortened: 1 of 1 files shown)


krancour:
I like this, but one thing I think could be potentially confusing is that the `_build_pod_template` method isn't actually building a template. As far as I can tell, it's building a fully-formed manifest that's ready for submission to k8s. Maybe renaming this method and variables whose value is returned from this method might be in order?

helgi:
To what? It is constructing / building a template to be used by other functions for submissions or inclusion as part of their template

krancour:
For example:

```
template = self._build_pod_template(namespace, name, image, **kwargs)
  
url = self._api("/namespaces/{}/pods", namespace)
response = self.session.post(url, json=template)
```

`self.session.post` submitting the `json` directly to k8s, right? If I'm reading that correctly, then what you're submitting isn't a actually "template," but rather a fully-formed manifest.

helgi:
Correct, so you'd prefer calling it `manifest`? btw if you look at the previous code base we always referred to it as `template`, both in the variable and the global one we ingested

krancour:
So `POD_TEMPLATE`, for instance, _was_ a template-- it had placeholders in it for data that was substituted in later by rendering that template, like so:

```
POD = POD_TEMPLATE
...
template = ruamel.yaml.load(string.Template(POD).substitute(l))
```

I suppose what you're referring to is the fact that the final result of that was, even then, assigned to a variable named "template." So I guess you'd be right in saying this isn't a new problem. Point taken. 

I still think it makes sense to fix it in this pr and improve the accuracy by calling it a manifest. Method and variable names that accurately denote their purpose are a great service to whoever looks at the code next.

helgi:
I can change it since e2e is being a pain anyway. Consider this tho, for RC and Deployments (and ReplicaSet I guess) the `pod` addition is called `Pod Template` and not `Pod Manifest` even if generally individual configurations are called manifests

helgi:
Changed to `manifest`

krancour:
> Consider this tho, for RC and Deployments (and ReplicaSet I guess) the pod addition is called Pod Template

Fair point. There's some overloading of the term `template` at play, I guess.

rootfs/scheduler/__init__.py
```diff
@@ -190,7 +43,7 @@
 
 SECRET_TEMPLATE = """\
 kind: Secret
-apiVersion: $version
+apiVersion: v1
```
kmala: why do we want to hard code the version..wouldn't it be simpler during an upgrade using the existing class variable

helgi: Because deployments for example does not use v1 - the global variable buys us nothing and should rather be treated per resource object

rootfs/scheduler/__init__.py
```diff
@@ -1358,23 +1268,25 @@ def _pod_pending_status(self, pod):
             return 'Pending', ''
 
         name = '{}-{}'.format(pod['metadata']['labels']['app'], pod['metadata']['labels']['type'])
-        for container in pod['status']['containerStatuses']:
-            # find the right container in case there are many on the pod
-            if container['name'] != name:
-                continue
+        # find the right container in case there are many on the pod
+        container = self._find_container(name, pod['status']['containerStatuses'])
```
bacongobbler: nice cleanup. :+1:

krancour:
I meant to LGTM this after you fixed the nit I picked, and apparently I did not. Now I've been beaten to it. Sorry. fwiw, LGTM3.

PR #822, opened 2016-06-16
fix(api): remove command escaping from v1

### LINKED ISSUE #821: cannot run process with multiple arguments
Given the following Procfile for [example-dockerfile-procfile-http](https://github.com/deis/example-dockerfile-procfile-http):

```
web: /bin/boot
worker: while true; do echo hello world; sleep 10; done
sleeper: sleep 900
```

I can scale the web type, but `sleeper` and `worker` both fail after scaling:

```
><> k --namespace=www get po
NAME                   READY     STATUS             RESTARTS   AGE
www-v2-sleeper-a59d2   0/1       CrashLoopBackOff   1          12s
www-v2-web-8ofdm       1/1       Running            0          5m
www-v2-worker-wdok9    0/1       Error              3          45s
><> k --namespace=www logs www-v2-worker-wdok9
true;: -c: line 0: unexpected EOF while looking for matching `''
true;: -c: line 1: syntax error: unexpected end of file
```

And the actual invocation of the command in the k8s manifest looks like this:

```
spec:
  containers:
  - args:
    - bash
    - -c
    - '''while'
    - true;
    - do
    - echo
    - hello
    - world;
    - sleep
    - 10;
    - done'
```

Looks like we need to fix up some of the command parsing logic from v1.

FYI this is on workflow-v2.0.0

bacongobbler:
This code used to be necessary to prevent systemd abuse. Now that kubernetes handles the escaping
for us, there is no need to be super defensive in our own code.

closes #821 
closes https://github.com/deis/deis/issues/5060

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -133,23 +133,33 @@ class App(UuidAuditedModel):
         return "{app}-{version}-{container_type}".format(**locals())
 
     def _get_command(self, container_type):
+        """
+        Return the kubernetes "container arguments" to be sent off to the scheduler.
+
+        In reality this is the command that the user it attempting to run.
+        """
         try:
-            # if this is not procfile-based app, ensure they cannot break out
-            # and run arbitrary commands on the host
             # FIXME: remove slugrunner's hardcoded entrypoint
             release = self.release_set.latest()
             if release.build.dockerfile or not release.build.sha:
-                return "bash -c '{}'".format(release.build.procfile[container_type])
+                return [release.build.procfile[container_type]]
 
-            return 'start {}'.format(container_type)
+            return ['start', container_type]
         # if the key is not present or if a parent attribute is None
         except (KeyError, TypeError, AttributeError):
             # handle special case for Dockerfile deployments
-            return '' if container_type == 'cmd' else 'start {}'.format(container_type)
+            return [] if container_type == 'cmd' else ['start', container_type]
 
-    def _get_command_run(self, command):
-        # SECURITY: shell-escape user input
-        command = command.replace("'", "'\\''")
+    def _get_entrypoint(self, container_type):
+        """
+        Return the kubernetes "container command" to be sent off to the scheduler.
+
+        In this case, it is the entrypoint for the docker image. Because of Heroku compatibility,
+        Any containers that are not from a buildpack are run under /bin/bash.
+        """
+        # handle special case for Dockerfile deployments
+        if container_type == 'cmd':
+            return []
 
         # if this is a procfile-based app, switch the entrypoint to slugrunner's default
         # FIXME: remove slugrunner's hardcoded entrypoint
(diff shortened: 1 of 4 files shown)


bacongobbler:
pinging @helgi on this one. Whenever I test this PR against https://github.com/deis/example-dockerfile-procfile-http and scale up the worker process, I hit this interesting bug:

```
><> deis scale worker=1
Error: 
503 Service Unavailable
detail: Single '}' encountered in format string
```

The controller logs aren't very helpful either...

```
INFO scale www-v2-worker, img 10.247.225.28:80/www:git-0c483886, cmd "while true; do echo hello world; sleep 10; done"
ERROR [www]: www-v2-worker (scale): Single '}' encountered in format string
10.246.64.14 "POST /v2/apps/www/scale/ HTTP/1.1" 503 52 "Deis Client v2.0.0"
```

bacongobbler:
back up for review again. Now whenever I run `deis run echo hi` I hit #862 which is solved by running `deis config:set PORT=5000` before running that command.

Manual testing:

1. deploy a deis cluster with this PR
2. deploy [example-dockerfile-procfile-http](https://github.com/deis/example-dockerfile-procfile-http)
3. `deis scale worker=1` should work and the process should be responsive
4. `deis run echo hi` should also work

bacongobbler:
Hold up; still seeing the same error when scaling the worker process...

```
detail: Single '}' encountered in format string
```

bacongobbler:
Aha, now I'm actually getting logs this time!

```
ERROR:root:Single '}' encountered in format string
  File "/app/scheduler/__init__.py", line 255, in scale
    self.create_rc(namespace, name, image, entrypoint, command, **kwargs)
  File "/app/scheduler/__init__.py", line 907, in create_rc
    manifest['spec']['template'] = self._build_pod_manifest(namespace, name, image, **kwargs)
  File "/app/scheduler/__init__.py", line 357, in _build_pod_manifest
  File "/app/scheduler/__init__.py", line 470, in _set_container
    self.update_secret(namespace, secret_name, secrets_env)
  File "/app/scheduler/__init__.py", line 1107, in update_secret
    name, namespace
scheduler.KubeHTTPException: ('failed to failed to update Secret "www-v2-env" in Namespace "www": 400 Bad Request\n{\'metadata\': {}, \'message\': \'Secret in version "v1" cannot be handled as a Secret: [pos 398]: json: error decoding base64 binary \\\'80\\\': illegal base64 data at input byte 0\', \'apiVersion\': \'v1\', \'status\': \'Failure\', \'kind\': \'Status\', \'code\': 400, \'reason\': \'BadRequest\'}', 'www-v2-env', 'www')
```

bacongobbler:
K, I have gone through the manual steps and all looks good now. Manual steps have been updated and I have fixed and/or otherwise filed issues for problems I have encountered. Ready for review!

bacongobbler:
aha, good thing we have e2e tests. e2e found an issue with `deis pull` apps with this change. Fixing

bacongobbler:
fixed. c'mon jenkins!

bacongobbler:
Hmm, now this is an interesting error:

> 13:57:57 Error: Unknown Error (503): {"detail":"(list pods): Release matching query does not exist."}

Don't think that's related so I'll re-run CI but I'll take a look.

bacongobbler:
@smothiki mentioned to me in slack that this is unrelated

helgi:
Manually tested. LGTM

PR #833, opened 2016-06-21
ref(gunicorn/config.py): double default worker count

mboersma:
# Summary of Changes

Doubles the default value for gunicorn workers, which increases responsiveness to a point where e2e tests seem to pass reliably with `GINKGO_NODES=30`, at the cost of more memory.

See http://docs.gunicorn.org/en/stable/design.html#how-many-workers and http://docs.gunicorn.org/en/stable/settings.html#workers.

Refs deis/workflow-e2e#215.
Replaces #835.

# Pull Request Hygiene TODOs

### DIFF
--- a/rootfs/deis/gunicorn/config.py
+++ b/rootfs/deis/gunicorn/config.py
@@ -10,7 +10,7 @@ try:
     if workers < 1:
         raise ValueError()
 except (NameError, ValueError):
-    workers = (os.cpu_count() or 4) * 2 + 1
+    workers = (os.cpu_count() or 4) * 4 + 1
 
 pythonpath = dirname(dirname(dirname(realpath(__file__))))
 timeout = 1200


helgi:
So from ~8 to ~16 or so? Should probably be fine as long as we don't see OOM

mboersma:
> Should probably be fine as long as we don't see OOM

It's been happy on my cluster, but I'll try to get some memory consumption numbers so we can compare.

mboersma:
This same change could be accomplished by setting `GUNICORN_WORKERS` to a larger value in the jenkins job or somesuch. But I think the workflow-e2e tests in CI are effectively a basic load test that most users would expect to perform well.

bacongobbler:
just gotta make sure we update the documentation, but LGTM

mboersma:
Here is a comparison of memory usage and performance for the default (cpu * 2 + 1 `sync` workers), this PR which doubles that, and #835 which adds 2 threads / worker:

| | VSZ | RSS | Req/s |
|---|---|---|---|---|
| Default | 945132 | 277928 | 262 |
| Double Workers | 1681912 | 515460 | 413 |
| Triple Threads | 2068676 | 303176 | 410 |

Methodology: I ran the e2e tests and then several rounds of `boom -n 10000 -c 150 <hostname>` on the same host to produce these numbers. I measured memory usage by shelling into the container and totalling the python processes listed by `ps aux`.

PR #835, opened 2016-06-22
ref(gunicorn/config.py): triple default thread count

mboersma:
# Summary of Changes

Triples the default thread count for gunicorn workers, which increases responsiveness to a point where e2e tests seem to pass reliably with `GINKGO_NODES=30`, at the cost of slightly more memory.

See http://docs.gunicorn.org/en/stable/design.html#how-many-threads and http://docs.gunicorn.org/en/stable/settings.html#threads.

Refs deis/workflow-e2e#215.
Replaces #833.

# Pull Request Hygiene TODOs

### DIFF
--- a/rootfs/deis/gunicorn/config.py
+++ b/rootfs/deis/gunicorn/config.py
@@ -11,6 +11,7 @@ try:
         raise ValueError()
 except (NameError, ValueError):
     workers = (os.cpu_count() or 4) * 2 + 1
+threads = 3
 
 pythonpath = dirname(dirname(dirname(realpath(__file__))))
 timeout = 1200


mboersma:
Here is a comparison of memory usage and performance for the default (cpu * 2 + 1 `sync` workers), #833 which doubles that, and this PR which adds 2 threads / worker:

| | VSZ | RSS | Req/s |
|---|---|---|---|---|
| Default | 945132 | 277928 | 262 |
| Double Workers | 1681912 | 515460 | 413 |
| Triple Threads | 2068676 | 303176 | 410 |

Methodology: I ran the e2e tests and then several rounds of `boom -n 10000 -c 150 <hostname>` on the same host to produce these numbers. I measured memory usage by shelling into the container and totalling the python processes listed by `ps aux`.

helgi:
So to paraphrase, double workers improves req/s but doubles memory in use and bumps virtual memory by 50% but then adding more thread brings down the memory usage while keeping reqs count, however ballons virtual memory a bit

So off hand I'd say we have to get both the PRs instead of only one or the other

mboersma:
It appears that either increasing the threads or the worker count is good enough to make e2e pass more consistently. I think the memory / speed tradeoff is more acceptable for this "triple threads" PR, but I wanted to get more feedback.

bacongobbler:
+1 for triple threads then! Nice mini chart.

helgi:
What does the combo of the two look like?

mboersma:
> What does the combo of the two look like?

@helgi I cherry-picked both changes and repeated the test. I think we're getting some memory thrashing because performance actually degraded a lot with this combination:

| | VSZ | RSS | Req/s |
|---|---|---|---|---|
| Default | 945132 | 277928 | 262 |
| Double Workers | 1681912 | 515460 | 413 |
| Triple Threads | 2068676 | 303176 | 410 |
| DoubleW + TripleT | 3653172 | 478880 | 120 |

helgi:
Nuts, well more workers should pretty much mean double memory requirement. How much memory was available?

mboersma:
> How much memory was available?

I'm not sure if there are other limits, but I've been testing on a GKE cluster with n1-standard-2 nodes (2 vCPUs, 7.5 GB memory). The actual CI test nodes are fatter: n1-standard-4 (4 vCPUs, 15 GB memory).

helgi:
Would be curious to see the combo ran on a bigger box but strictly not required for merging this

jgmize:
Increasing the thread count instead of the process count may save some memory, but it increases the risk of hitting threading-related bugs at high concurrency.  Given the choice I'd rather spend the extra memory for more processes.

mboersma:
> increases the risk of hitting threading-related bugs

The thread implementation for `sync` workers has been around for a long time in gunicorn and doesn't appear particularly buggy according to their issue list. But I hear you--it's not as tested in this environment as the existing 1 thread / worker.

Another argument for #833 is that it's simply changing a default value on a knob that users can already turn themselves, whereas adding `threads = 3` is a hidden default that can't be customized (although I could add that if needed).

PR #823, opened 2016-06-16
feat(healthcheck): Add healthchecks as separate configuration

kmala:
# Summary of Changes

Consume health checks as separate configuration.

# Issue(s) that this PR Closes

fixes #603 

# Associated [End To End Test](https://github.com/deis/workflow-e2e) PR(s)

# Associated [Documentation](https://github.com/deis/workflow) PR(s)

https://github.com/deis/workflow/pull/322

# Associated [Design Document](http://docs.deis.io/en/latest/contributing/design-documents)(s)

# Testing Instructions

https://github.com/deis/workflow/pull/322

# Pull Request Hygiene TODOs

Please make sure the below checklist is complete.

- [x] Your pull request is concise and to the point (make another PR for refactoring nearby code)
- [x] Your commits are squashed into logical units of work
- [x] Your commits follow the [commit style guidelines](http://docs.deis.io/en/latest/contributing/standards/#commit-style-guide)

:cherry_blossom: Thank you! :cherry_blossom:

Issue #322, opened 2016-01-29
Document Platform SSL


slack:
Complete: https://github.com/deis/workflow/blob/master/docs/src/managing-deis/platform-ssl.md

helgi:
Closing in favour of deis/docs-v2#25

PR #836, opened 2016-06-22
feat(api): Add healthcheck field to Config

### LINKED ISSUE #599: Add HTTP Headers as an option to k8s healthchecks
* http://kubernetes.io/docs/user-guide/liveness/
* http://kubernetes.io/docs/api-reference/v1/definitions/#_v1_httpgetaction

### LINKED ISSUE #603: Move Health Checks to a top level configuration object
Right now all health check logic lives as a normal config value with a `HEALTHCHECK_*` env var like naming convention - given how important this element has come and the expansion of its functionalities in the future then we should consider moving this to a top level configuration (like tags) and add new CLI endpoints for it.

`deis health:set url=/`, `deis health:set port=4999` and so on

A migration can be pretty easily written to move things within the config object to represent the new location.

While we are at it there should be a discussion around introducing the separation between liveness and readiness probes (but keep ability to set a global one to apply to both for a given application) and how we want to arch the introduction of `Exec` and 'TCPSocket' in addition to `HTTPGet`

http://kubernetes.io/docs/user-guide/pod-states/#container-probes

bacongobbler:
# Summary of Changes

This PR adds a `healthcheck` field to the controller to handle the `deis healthchecks` command, and to eventually deprecate the old way of setting healthchecks via `deis config:set HEALTHCHECK_URL`. This is a continuation of #823.

closes #603
closes #599

TODO

deis/workflow#322

deis/workflow#322

# Testing Instructions

TODO

# Pull Request Hygiene TODOs


bacongobbler:
so for others to be aware of, this PR passes when setting `exec` and `tcpSocket` checks, but for some reason I cannot set `httpGet` probes... Will have to look more into it, because all I get back from the server is

```
><> deis healthchecks:set readiness httpGet 5000
Applying readinessProbe healthcheck... Error: 400 Bad Request

><> kd logs deis-controller-abc12
10.246.72.17 "POST /v2/apps/go/config/ HTTP/1.1" 400 47 "Deis Client v2.1.0-dev"
```

No traceback :cry:

helgi:
I believe you are being hit by https://github.com/deis/controller/issues/816

helgi:
Can we change the migration so that `HEALTHCHECK_` gets migrated from `values` into `healthchecks` with it's appropriate defaults under `httpGet` ?

Same basically needs to be done when people do `config:set HEALTHCHECK*`

Doing that would reduce the various duplications in the code that spider outside of the `Config` model

bacongobbler:
> Can we change the migration so that HEALTHCHECK_ gets migrated from values into healthchecks with it's appropriate defaults under httpGet ?

This might be easier to tackle via the client. We can convert HEALTHCHECK_* in the client to send a real `healthcheck` object, but keep this code in for old time's sake until we eventually deprecate it. I'm not sure we can easily do a data migration from config to healthcheck, and we still need to support existing `config` objects that handle HEALTHCHECK_*. would that be reasonable, or how would you imagine we could tackle this issue?

helgi:
We can't really guarantee that a user is on the latest client, right?

bacongobbler:
also what if for some crazy reason the user also required that envvar in their app? Migrating it to the healthcheck field would mean the envvar is no longer present in their app (considered a breaking change)

helgi:
I think that might be a stretch that in the 2.0 timeframe someone would start depending on the healthcheck env var within their own app

helgi:
Another option is to populate `healthchecks` but also keep it in `values` that way the env var stays but internally we always use `healthchecks`

bacongobbler:
that would be a good tradeoff. I can tackle that, most likely.

bacongobbler:
this is finally ready for review. @helgi I implemented the change you requested and added unit tests to cover that. Now any time a config object is saved (including existing ones), a healthcheck object is created if the legacy envvars exist.

rootfs/scheduler/__init__.py
```diff
@@ -936,62 +936,6 @@ def delete_rc(self, namespace, name):
 
         return response
 
-    def _healthcheck(self, namespace, container, routable=False, path='/', port=5000,
```
helgi: I find it a little dangerous when we are not keeping the healthcheck logic inside the scheduler but it is rather now outside in serializer class... Thoughts?

rootfs/api/models/config.py
```diff
@@ -137,13 +111,12 @@ def save(self, **kwargs):
...
                         if key not in data:
                             raise UnprocessableEntity('{} does not exist under {}'.format(key, attr))  # noqa
-
                         data.pop(key)
                     else:
                         data[key] = value
                 setattr(self, attr, data)
 
-            self.set_healthchecks()
+            self._migrate_legacy_healthcheck()
```
helgi: Nice! We just need to aggressively ensure that people don't have legacy health check and then try to set new checks. On controller restart it would always overwrite. Maybe add in a "if livenessProbe in self.healthcheck" is true then skip setting legacy

rootfs/scheduler/__init__.py
bacongobbler: this is essentially doing the same job as `jsonschema` is in the serializer, but it fails much later in the process *after* someone has already created the healthcheck. I think the way we validate the healthcheck through the serializer is much better, but I'm open to opinions.

rootfs/api/tests/test_config.py
```diff
@@ -849,3 +849,125 @@ def test_healthchecks_validations(self, mock_requests):
...
+
+        # Set healthcheck URL to get defaults set
+        response = self.client.post(
+            '/v2/apps/{app.id}/config'.format(**locals()),
+            {'values': json.dumps({'HEALTHCHECK_URL': '/health'})}
+        )
+        self.assertEqual(response.status_code, 201, response.data)
+        self.assertIn('HEALTHCHECK_URL', response.data['values'])
+        self.assertEqual(response.data['values']['HEALTHCHECK_URL'], '/health')
+        # legacy defaults
```
bacongobbler: A note I should bring up... Are these really the defaults we should be using? I referenced them based off of https://github.com/deis/controller/blob/master/rootfs/api/models/config.py#L40-L45 but the scheduler has conflicting options.

helgi: Yeah the Model always overwrote the defaults in the scheduler, the former may have evolved while the latter was just left behind with an egg on its face

rootfs/api/models/config.py
bacongobbler: Good thinking. My current logic is that if someone is using legacy healthchecks, they're gonna have to unset them before continuing on to set a livenessProbe healthcheck.

Shall I also add a deprecation notice here to eventually rip this out after 2.4 or something?

rootfs/scheduler/__init__.py
helgi: Yes, it sure is and I think the serializer should still do the validation but I was mostly wondering given how we consider the scheduler a separate module that maybe serializer should pull in "validation" from the scheduler plus we may have a different API vs Scheduler structure in the health check department, either due to differences in k8s versions or changes in internal design.

Trying to make sure we don't cause us debt later :) I'm not fully sold on any of the above ideas particularly, yet.

rootfs/api/models/config.py
helgi: Yeah, I think we may have to keep it to the next major release given how strict we are usually on breakages. @slack thoughts on how long we should keep it deprecated before removal?

rootfs/scheduler/__init__.py
bacongobbler: so you want to move the k8s healthcheck schema over to the `scheduler` module instead of sticking it in the serializer? That makes sense. Just trying to grok what you wanna change :)

helgi: Yeah, that was **one** of the things rattling around. Was also looking for other ideas if you see a better way :) A bit of pre-caffeine rubber ducking.

If we moved it over then I wouldn't want to lose the error message handling in the Serializer. It's a little funky when we have such big scope shared in two places but the nature of the beast I guess :)

rootfs/api/models/config.py
kmala: The reason I din't migrate before I left is how are we gonna handle port.... current environment health checks doesn't provide port and we use to get it from releases in the scheduler logic.... Port is the only mandatory field and we won't have handle of it in the config object

helgi: We could pull the service port for now to solve for that

bacongobbler: The existing behaviour of pulling in the service port should work in that case. I'll see if I can write some tests for that.

kmala: You can set config before creating a build and the service port is actually overridden during deploy of the app but the migration happens during a new config....
What-if the client does all his config first and does a build next...then the health check might have wrong port

bacongobbler:
I am not proud of this code, but https://github.com/deis/controller/pull/836/commits/6fd0add9908272116500beb540676386bdd968cd should do it.

rootfs/api/models/config.py
bacongobbler: I fixed this in 6fd0add by only injecting the port if it is not already defined (such as when using `deis config:set HEALTHCHECK_URL`). That should alleviate your concern with 

> What-if the client does all his config first and does a build next...then the health check might have wrong port

helgi:
I added a `liveness` probe of `httpGet` to an app (which worked) and then later I added a `readiness` probe of `exec` and I got this exceptions:

```
ERROR [oddest-lungfish]: oddest-lungfish-v18-cmd (app::deploy): 'livenessProbe'
ERROR:root:oddest-lungfish-v18-cmd (app::deploy): 'livenessProbe'
  File "/app/scheduler/__init__.py", line 114, in deploy
    self.get_rc(namespace, name)
  File "/app/scheduler/__init__.py", line 709, in get_rc
    'get ReplicationController "{}" in Namespace "{}"', name, namespace
scheduler.KubeHTTPException: ('failed to get ReplicationController "oddest-lungfish-v18-cmd" in Namespace "oddest-lungfish": 404 Not Found\n{\'message\': \'replicationcontrollers "oddest-lungfish-v18-cmd" not found\', \'kind\': \'Status\', \'reason\': \'NotFound\', \'code\': 404, \'status\': \'Failure\', \'apiVersion\': \'v1\', \'metadata\': {}, \'details\': {\'name\': \'oddest-lungfish-v18-cmd\', \'kind\': \'replicationcontrollers\'}}', 'oddest-lungfish-v18-cmd', 'oddest-lungfish')

During handling of the above exception, another exception occurred:

  File "/app/api/models/app.py", line 445, in deploy
    **kwargs
  File "/app/scheduler/__init__.py", line 120, in deploy
    new_rc = self.create_rc(namespace, name, image, command, replicas=0, **kwargs).json()
  File "/app/scheduler/__init__.py", line 889, in create_rc
    manifest['spec']['template'] = self._build_pod_manifest(namespace, name, image, **kwargs)
  File "/app/scheduler/__init__.py", line 348, in _build_pod_manifest
  File "/app/scheduler/__init__.py", line 516, in _set_container
    healthchecks['livenessProbe'].get('httpGet') is not None and
KeyError: 'livenessProbe'

The above exception was the direct cause of the following exception:

  File "/app/api/views.py", line 261, in post_save
    config.app.deploy(self.release)
  File "/app/api/models/app.py", line 457, in deploy
    raise ServiceUnavailable(err) from e
api.exceptions.ServiceUnavailable: oddest-lungfish-v18-cmd (app::deploy): 'livenessProbe'

The above exception was the direct cause of the following exception:

  File "/usr/local/lib/python3.5/dist-packages/rest_framework/views.py", line 463, in dispatch
  File "/app/api/views.py", line 171, in create
    return super(AppResourceViewSet, self).create(request, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/rest_framework/mixins.py", line 21, in create
    self.perform_create(serializer)
  File "/app/api/viewsets.py", line 21, in perform_create
    self.post_save(obj)
  File "/app/api/views.py", line 264, in post_save
    raise DeisException(str(e)) from e
api.exceptions.DeisException: oddest-lungfish-v18-cmd (app::deploy): 'livenessProbe'
```

bacongobbler:
this was actually part of the fix I made this morning for the auto-port populate. thanks for catching that!

bacongobbler:
fix'd

PR #841, opened 2016-06-29
chore(rootfs): bump platform version to 2.1.0

bacongobbler:


rootfs/Dockerfile
```diff
@@ -31,4 +31,4 @@ WORKDIR /app
 CMD ["/app/bin/boot"]
 EXPOSE 8000
 
-ENV WORKFLOW_RELEASE 2.0.0
+ENV WORKFLOW_RELEASE 2.1.0
```
mboersma: I think this env var is obsolete (it was originally `DEIS_RELEASE` and was created to force docker image diffs as I know you know @bacongobbler). It might still be useful documentation and I'm not suggesting this PR remove it, just noting.

I'll create a meta-issue to investigate and possibly remove `WORKFLOW_RELEASE` from Dockerfiles.

bacongobbler: sounds good. This was also semi-useful for determining what version of the server was installed using `deisctl`, but now users can just lookup the image tag in the kubernetes pod manifest making this moot. A meta issue to remove these would be great. :)

PR #843, opened 2016-06-29
chore(rootfs): bump version to v2.2.0-dev

bacongobbler:


helgi:
Shouldn't it be 2.2? :P

bacongobbler:
right, good catch.

PR #844, opened 2016-06-30
fix(scale): return error message in proper format

### LINKED ISSUE #831: Array response when trying to scale a process to less than 0
`POST /v2/apps/<id>/scale/` Body: `{"foo": -1}`

will return 400: `["Must be greater than or equal to zero"]`

This error is in a different format from every other API error I know of.

I would expect the error to be: `{"foo":["Must be greater than or equal to zero"]}` or  `{"details":["Must be greater than or equal to zero"]}`

kmala:
# Summary of Changes

Handle the case of validation error instead of value error as it value error re-thrown as validation error in the validate function.

fixes #831

# Testing Instructions

1. Create a Deis Cluster
2. Register an app
3. Try to call the scale endpoint as mentioned in the #831 and you should get a proper json error response.

# Pull Request Hygiene TODOs

### DIFF
--- a/rootfs/api/tests/test_pods.py
+++ b/rootfs/api/tests/test_pods.py
@@ -415,6 +415,8 @@ class PodTest(APITransactionTestCase):
         body = {'web': -1}
         response = self.client.post(url, body)
         self.assertEqual(response.status_code, 400, response.data)
+        self.assertEqual(response.data, {"detail": "Invalid scaling format: "
+                                         "['Must be greater than or equal to zero']"})
 
         # scale to something other than a number
         url = "/v2/apps/{app_id}/scale".format(**locals())
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -317,7 +317,7 @@ class App(UuidAuditedModel):
             for target, count in structure.copy().items():
                 structure[target] = int(count)
             validate_app_structure(structure)
-        except (TypeError, ValueError) as e:
+        except (TypeError, ValidationError) as e:
             raise DeisException('Invalid scaling format: {}'.format(e))
 
         # test for available process types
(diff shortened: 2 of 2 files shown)


bacongobbler:
can we add some unit tests here to validate this?

rootfs/api/tests/test_pods.py
```diff
@@ -415,6 +415,8 @@ def test_scale_errors(self, mock_requests):
         body = {'web': -1}
         response = self.client.post(url, body)
         self.assertEqual(response.status_code, 400, response.data)
+        self.assertEqual(response.data, {'detail': "Invalid scaling format: "
+                                         "['Must be greater than or equal to zero']"})
```
Joshua-Anderson: Is this valid JSON?

bacongobbler: Yes, it just looks confusing. the key is `detail`, the value is a string: `Invalid scaling format: ['Must be greater than or equal to zero']`. It just looks off here,

Joshua-Anderson: Having a value with datatypes `Invalid scaling format: ['Must be greater than or equal to zero']` is a little tricky to parse, and also doesn't really follow standard json usage. Something like this makes the error much easier to work with and is a lot less confusing for people trying to understand it.

```json
{
    "detail": {
        "scaling_format": ["Must be greater than or equal to zero"]
    }
}
```

bacongobbler: I agree, but that's out of the scope of this issue. Mind filing an issue for it?

helgi:
We could embed the "invalid scaling format" as part of the actual error message so that we don't have a string + list and instead just get a list. Thoughts?

helgi:
ah wait that might be hard since we are transforming ValidationError into a DeisException...

PR #846, opened 2016-06-30
fix(scheduler): if one RC fails to scale then ensure all other RCs are at the right level

helgi:
Previously only the failing RC would get fixed up but now the whole state will be reverted if one RC fails

I marked it as a `fix` as I felt the potential inconsistency in the app state was broken

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -331,7 +331,8 @@ class App(UuidAuditedModel):
                     'Container type {} does not exist in application'.format(container_type))
 
         # merge current structure and the new items together
-        new_structure = self.structure.copy()
+        old_structure = self.structure
+        new_structure = old_structure.copy()
         new_structure.update(structure)
 
         if new_structure != self.structure:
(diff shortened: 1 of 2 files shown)


rootfs/api/models/app.py
```diff
@@ -331,15 +331,21 @@ def scale(self, user, structure):  # noqa
...
         if new_structure != self.structure:
             # save new structure to the database
             self.structure = new_structure
             self.save()
 
-            self._scale_pods(structure)
+            try:
+                self._scale_pods(structure)
+            except ServiceUnavailable:
+                # scaling failed, go back to old scaling numbers
```
kmala: below we are handling kubeexception but here it is serviceunavailable.....are re raising it anywhere

helgi: `_scale_pods` takes `KubeException` (and by exension `KubeHTTPException`) and wraps it into a `ServiceUnavailable`

Prior to this change the call to `_scale_pod` would let `ServiceUnavailable` bubble all the way up

PR #850, opened 2016-07-01
fix(passwd): raise 400 when password is not a parameter

Joshua-Anderson:
# Summary of Changes

Previously, passwd didn't check if `password` or `new_password` was set resulting in server errors when trying to access them.
I haven't wrote any tests, cause I wanted to make sure this was the right approach.
I was also looking at serializers, but it doesn't seem like you can override the serializers on a per-view basis. I'm also concerned about using a `DeisException`, because I would rather have a field specific error, like serializers do.

# Testing Instructions

1. Create a normal user.
2. Try `POST /v2/auth/passwd` without `new_password` or `password` and make sure 500 errors are returned.


helgi:
Can you rebase? And add some tests

Joshua-Anderson:
@helgi Rebased and added some tests.

PR #853, opened 2016-07-06
docs(github): simplify pull request template

bacongobbler:
We've found that half of the engineering team either hates or doesn't even bother using the PR
templates. Refactoring it down to the bare minimum should suffice.


bacongobbler:
If this sounds like a good approach I'll go ahead and remove the other templates from the rest of the repositories

PR #854, opened 2016-07-07
feat(*): add Deployments support behind a feature flag

### LINKED ISSUE #19: Use k8s deployments?
Disclaimer: this is an alpha feature, but it looks interesting.  I think this could dramatically simplify some of what's in `scheduler/k8s.py` if it sticks around.

http://kubernetes.io/v1.1/docs/user-guide/deployments.html

helgi:
TODO:
- [x] There is one flakey tests for Deployments (mock related)
- [x] [docs](https://github.com/deis/workflow)
- [x] [e2e](https://github.com/deis/workflow-e2e)

Closes #19 
Requires deis/workflow-e2e#251
Requires deis/workflow#369

This adds Deployment object support (replaces ReplicationController) when a feature flag is enabled. This can be done by either settings a global ENV Var on the Controller or setting a config per app. The variable is `DEIS_KUBERNETES_DEPLOYMENTS=1`

A few new features (Deployments only) have been introduced or specifically expose Deployment functionalities:
* `KUBERNETES_DEPLOYMENTS_REVISION_HISTORY_LIMIT` (global and per app setting) specifies how many revisions (ReplicaSets) to keep
* Stop concurrent deployments (triggered by push, pull, config:set and so on)

If Deployments is enabled on an existing app then a whole new copy will be spun up (behind the scenes) as Deployment and then the old ReplicationController is spun down. This is a one way conversion.

Note that pod naming has changed due to how Deployments adds a hash to the name, among other things.

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -121,9 +121,16 @@ class App(UuidAuditedModel):
     def _get_job_id(self, container_type):
         app = self.id
         release = self.release_set.latest()
+
+        # see if there is a global or app specific setting to specify Deployments usage
+        deployments = bool(release.config.values.get('DEIS_KUBERNETES_DEPLOYMENTS', settings.DEIS_KUBERNETES_DEPLOYMENTS))  # noqa
+
+        # deployments does not need version in the job
+        if deployments:
+            return "{app}-{container_type}".format(**locals())
+
         version = "v{}".format(release.version)
-        job_id = "{app}-{version}-{container_type}".format(**locals())
-        return job_id
+        return "{app}-{version}-{container_type}".format(**locals())
 
     def _get_command(self, container_type):
         try:
@@ -237,27 +244,40 @@ class App(UuidAuditedModel):
 
     def restart(self, **kwargs):  # noqa
         """
-        Restart found pods by deleting them (RC will recreate).
-        Wait until they are all drained away and RC has gotten to a good state
+        Restart found pods by deleting them (RC / Deployment will recreate).
+        Wait until they are all drained away and RC / Deployment has gotten to a good state
         """
         try:
-            # Resolve single pod name if short form (worker-asdfg) is passed
-            if 'name' in kwargs and kwargs['name'].count('-') == 1:
-                if 'release' not in kwargs or kwargs['release'] is None:
-                    release = self.release_set.latest()
-                else:
-                    release = self.release_set.get(version=kwargs['release'])
+            if kwargs.get('release', None) is None:
+                release = self.release_set.latest()
+            else:
+                release = self.release_set.get(version=kwargs['release'])
 
-                version = "v{}".format(release.version)
-                kwargs['name'] = '{}-{}-{}'.format(kwargs['id'], version, kwargs['name'])
+            # see if there is a global or app specific setting to specify Deployments usage
+            deployments = bool(release.config.values.get('DEIS_KUBERNETES_DEPLOYMENTS', settings.DEIS_KUBERNETES_DEPLOYMENTS))  # noqa
 
-            # Iterate over RCs to get total desired count if not a single item
+            if deployments:
+                # Resolve single pod name if short form (cmd-1269180282-1nyfz) is passed
+                if 'name' in kwargs and kwargs['name'].count('-') == 2:
+                    kwargs['name'] = '{}-{}'.format(kwargs['id'], kwargs['name'])
+            else:
+                # Resolve single pod name if short form (worker-asdfg) is passed
+                if 'name' in kwargs and kwargs['name'].count('-') == 1:
+                    version = "v{}".format(release.version)
+                    kwargs['name'] = '{}-{}-{}'.format(kwargs['id'], version, kwargs['name'])
+
+            # Iterate over RCs / RSs to get total desired count if not a single item
             desired = 1
             if 'name' not in kwargs:
                 desired = 0
                 labels = self._scheduler_filter(**kwargs)
-                controllers = self._scheduler.get_rcs(kwargs['id'], labels=labels).json()['items']
-                for controller in controllers:
+                # fetch RS (which represent Deployments) / RCs
+                if deployments:
+                    controllers = self._scheduler.get_replicasets(kwargs['id'], labels=labels)
+                else:
+                    controllers = self._scheduler.get_rcs(kwargs['id'], labels=labels)
+
+                for controller in controllers.json()['items']:
                     desired += controller['spec']['replicas']
         except KubeException:
             # Nothing was found
@@ -265,7 +285,7 @@ class App(UuidAuditedModel):
 
         try:
             for pod in self.list_pods(**kwargs):
-                # This function verifies the delete. Gives pod 30 seconds
+                # This function verifies the delete
                 self._scheduler.delete_pod(self.id, pod['name'])
         except Exception as e:
             err = "warning, some pods failed to stop:\n{}".format(str(e))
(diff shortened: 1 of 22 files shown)


helgi:
Note that I copied all relevant tests into `api/tests/deployments` to test deployments by enabling the feature flag - we can remove that wholesale when Deployments are the default

slack:
Image: quay.io/jhansen/controller:git-8f948c7

So far looking pretty good. Manually tested (3 node cluster, AWS):
* scale=15, default deploy worked
* scale=15, interrupted deis cli, deployment finished!, subsequent config:set properly 400s
* scale=15 custom batch size of 5 worked well

bacongobbler:
holy mother of PRs... Gonna need copius amounts of coffee to review this. :coffee: :coffee: :coffee:

bacongobbler:
also note that I added "needs docs" because we need documentation on how to enable this, how it all works, etc.

helgi:
To test migration:

* Create an app, scale to 10
* `kubectl get rc,po --namespace <app>` to verify it is an RC
* set `DEIS_KUBERNETES_DEPLOYMENTS=1`
* `kubectl get rc,po,deployments --namespace <app>` to verify a Deployment exists and the RC is gone

bacongobbler:
Once a formal doc proposal is made for this I'll be happy to test. Until then I have no idea how to begin reviewing this PR. :cyclone:

helgi:
The test plan for this PR is pretty broad so I am only going to describe parts of the test plan

* create application (`deis create --no-remote deployment-pull`)
* enable deployment on the app (`deis config:set DEIS_KUBERNETES_DEPLOYMENTS=1 -a deployment-pull`)
* deploy app (`deis pull deis/example-go -a deployment-pull`)
* verify with kubernetes (`kubectl get rc,deployments,rs,pods --namespace deployment-pull`), there should be no RC but instead Deployments
* scale to 10 and do the above verification
* run `ps:list`
* run `ps:restart`
* use `deis run`

And any other normal application action that would interact with Kubernetes

The same exercise can be applied to BuildPacks and DockerFile/Procfile

Also, see previous comment to verify that RC gets migrated to Deployment

helgi:
The duplication of tests has been separated out into a secondary commit - This should make it easier to review the main code changes

rootfs/api/models/app.py
```diff
@@ -457,14 +496,23 @@ def deploy(self, release):
...
             }
 
         # Sort deploys so routable comes first
         deploys = OrderedDict(sorted(deploys.items(), key=lambda d: d[1].get('routable')))
 
         for scale_type, kwargs in deploys.items():
             try:
+                # Is there an existing deployment in progress?
+                name = self._get_job_id(scale_type)
+                if not force_deploy and release.deployment_in_progress(self.id, name):
```
arschles: seems like the `deployment_in_progress` function uses an in-memory boolean, which would mean that we can't run multiple controller replicas. I know we have other functionality that uses in-memory state, so this particular one isn't a dealbreaker, but do you have any aspirations or plans to refactor to be able to run 2+ replicas in the future?

helgi: Nope, haven't thought about it for the overall Controller. But it's fine as an in-memory boolean since it is checking the state of the Deployment object (owned by Kubernetes) so there is no stateful information to be concerned about for 2+ replicas of the controller as they all hit the same source of truth (kubernetes)

arschles: cool - thanks for letting me know. :+1:

rootfs/scheduler/mock.py
```diff
@@ -274,6 +278,104 @@ def upsert_pods(controller, url):
...
+        deployment['status']['updatedReplicas'] = current
+        deployment['status']['availableReplicas'] = current
+        deployment['status']['unavailableReplicas'] = deployment['spec']['replicas'] - current
+        cache.set(url, deployment, None)
+
+        # all ready and matching the intent
+        if current == deployment['spec']['replicas']:
+            break
+
+        time.sleep(0.5)
```
arschles: what's this for? a comment would be good here as well

helgi: Just the sleep before it starts the next go-around of hitting `cleanup_pods` and `pod_state_transitions`. Needed to do it so the system didn't get out of whack. Can add some comments

arschles: makes sense. thanks for adding comments

rootfs/api/models/release.py
```diff
@@ -330,6 +370,20 @@ def _delete_release_in_scheduler(self, namespace, version):
...
 
+    def deployment_in_progress(self, namespace, name):
+        # see if there is a global or app specific setting to specify Deployments usage
+        deployments = bool(self.config.values.get('DEIS_KUBERNETES_DEPLOYMENTS', settings.DEIS_KUBERNETES_DEPLOYMENTS))  # noqa
+        if not deployments:
+            return False
+
+        try:
+            ready, _ = self._scheduler.are_deployment_replicas_ready(namespace, name)
+            return (False if ready else True)
```
arschles: is there an equivalent to `return (!ready)` in python?

helgi: Actually yeah I could make this `return not ready` - I'll amend that

rootfs/scheduler/__init__.py
```diff
@@ -612,7 +687,19 @@ def pod_state(self, pod):
 
     def _api(self, tmpl, *args):
         """Return a fully-qualified Kubernetes API URL from a string template with args."""
-        url = "/api/{}".format(self.apiversion) + tmpl.format(*args)
+        # FIXME better way of determining API version based on requested component
+        # extensions use apis and not api
+        # TODO this needs to be aware that deployments / rs could be top level in future releases
```
arschles: are there issues for the fix and the todo?

helgi: Not yet, I'll make create a ticket - I have a fix in mind, but it isn't urgent enough

rootfs/scheduler/__init__.py
```diff
@@ -1420,5 +1506,292 @@ def get_node(self, name, **kwargs):
...
+        """
+        Looks at status/observedGeneration and metadata/generation and
+        waits for observedGeneration >= generation to happen
+
+        http://kubernetes.io/docs/user-guide/deployments/#the-status-of-a-deployment
+        More information is also available at:
+        https://github.com/kubernetes/kubernetes/blob/master/docs/devel/api-conventions.md#metadata
+        """
+        logger.debug("waiting for Deployment {} to get a newer generation (30s timeout)".format(name))  # noqa
+        for _ in range(30):
```
arschles: seems like this isn't really a 30 second timeout - it's more like a `30 * time(get_deployment)` second timeout. does that matter?

rootfs/scheduler/__init__.py
```diff
@@ -1420,5 +1506,292 @@ def get_node(self, name, **kwargs):
...
+        # if replicas are higher than maxSurge then the old deployment is never scaled down
+        # maxSurge can't be 0 when maxUnavailable is 0 and the other way around
+        if replicas > 0 and replicas < maxSurge:
+            maxSurge = replicas
+
+        # http://kubernetes.io/docs/user-guide/deployments/#strategy
+        manifest['spec']['strategy'] = {
+            'rollingUpdate': {
+                'maxSurge': maxSurge,
+                # TODO potentially support this
```
arschles: issue for this?

rootfs/scheduler/__init__.py
helgi: Not inherently, I never figure out a good way to express it was more than 30 seconds but a variable of how long each `get_deployment` would take. I guess it's less of a timeout and more of a 30 attempts

arschles:
@helgi I added some comments, mostly nits & not blocking this PR. aside from them, LGTM

rootfs/scheduler/__init__.py
helgi: Coming up, had it written down on on a piece of paper. Transcribing to tickets now

PR #861, opened 2016-07-09
ref(*): use requests.Session as a singleton

mboersma:
The scheduler's k8s API access and app.py's healthcheck calls would create a new [`requests.Session`](http://docs.python-requests.org/en/master/user/advanced/#session-objects) with each class used, resulting in many `urllib3` allocations as printed in the controller logs. Since `Session` persists parameters and cookies and uses connection pooling, creating and destroying it often isn't optimal usage.

This refactoring moves the two usages of `Session` into module-level objects with a `get`ter method--the cheap way of making an object a singleton in python.

After this change, I don't see large blocks of connection allocations in the controller logs any more, but I do see this sometimes, which means it's working:
```console
INFO:requests.packages.urllib3.connectionpool:Resetting dropped connection: 10.203.253.73
```

TODO:
- [x] basic benchmarking to show this isn't slower than before

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -27,6 +27,23 @@ from scheduler import KubeHTTPException, KubeException
 
 logger = logging.getLogger(__name__)
 
+session = None
+
+
+def get_session():
+    global session
+    if session is None:
+        session = requests.Session()
+        session.headers = {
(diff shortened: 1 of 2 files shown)


helgi:
Nice, there were times I wondered about connection galore! Especially against the k8s API

PR #863, opened 2016-07-10
ref(scheduler): generate secrets manifest for create and update secrets

### LINKED ISSUE #862: deis run: 'utf-8' codec can't decode byte 0xe7 in position 0: invalid continuation byte
I'm getting the following error on 2.1.0 when trying to do a `deis run /bin/bash` (command doesn't matter):

```
Running '/bin/bash'...
Error: Unknown Error (503): {"detail":"shop-backend-v7-run-ivbxw (run): 'utf-8' codec can't decode byte 0xe7 in position 0: invalid continuation byte"}
```

The logs for the deis-controller pod show:

```
INFO [shop-backend]: buenemann on shop-backend-v7-run-ivbxw runs ''/bin/bash''
INFO:api.models.app:[shop-backend]: buenemann on shop-backend-v7-run-ivbxw runs ''/bin/bash''
INFO run shop-backend-v7-run-ivbxw, img home/shop-backend:git-5cb6a5fe/push/slug.tgz, entrypoint /runner/init, cmd "'/bin/bash'"
INFO:scheduler:run shop-backend-v7-run-ivbxw, img home/shop-backend:git-5cb6a5fe/push/slug.tgz, entrypoint /runner/init, cmd "'/bin/bash'"
INFO:requests.packages.urllib3.connectionpool:Starting new HTTPS connection (1): 10.3.0.1
ERROR:root:'utf-8' codec can't decode byte 0xe7 in position 0: invalid continuation byte
  File "/app/api/models/app.py", line 631, in run
    **kwargs
  File "/app/scheduler/__init__.py", line 377, in run
    manifest = self._build_pod_manifest(namespace, name, image, **kwargs)
  File "/app/scheduler/__init__.py", line 348, in _build_pod_manifest
  File "/app/scheduler/__init__.py", line 461, in _set_container
    self.get_secret(namespace, secret_name)
  File "/app/scheduler/__init__.py", line 1031, in get_secret
    secrets['data'][key] = value.decode(encoding='UTF-8')
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe7 in position 0: invalid continuation byte
The above exception was the direct cause of the following exception:
  File "/usr/local/lib/python3.5/dist-packages/rest_framework/views.py", line 463, in dispatch
  File "/app/api/views.py", line 225, in run
    rc, output = app.run(self.request.user, request.data['command'])
  File "/app/api/models/app.py", line 637, in run
    raise ServiceUnavailable(err) from e
api.exceptions.ServiceUnavailable: shop-backend-v7-run-ivbxw (run): 'utf-8' codec can't decode byte 0xe7 in position 0: invalid continuation byte
10.2.21.6 "POST /v2/apps/shop-backend/run HTTP/1.1" 503 123 "Deis Client v2.1.0"
```

I'm running deis/workflow-v2.1.0 on a kube-aws 1.2.4 cluster.

helgi:
Move to the same model as Pod building. Also moves update secret to generate a whole new object to overwrite the previous one instead of just updating the data element
Unifies the base64 encoding and how labels are applied

This may help fix #862

### DIFF
--- a/rootfs/api/models/certificate.py
+++ b/rootfs/api/models/certificate.py
@@ -173,27 +173,27 @@ class Certificate(AuditedModel):
         # only create if it exists - We raise an exception when a secret doesn't exist
         try:
             name = '%s-cert' % self.name
-            app = domain.app
+            namespace = domain.app.id
             data = {
                 'tls.crt': self.certificate,
                 'tls.key': self.key
             }
 
-            secret = self._scheduler.get_secret(app, name).json()['data']
+            secret = self._scheduler.get_secret(namespace, name).json()['data']
         except KubeException:
-            self._scheduler.create_secret(app, name, data)
+            self._scheduler.create_secret(namespace, name, data)
         else:
             # update cert secret to the TLS Ingress format if required
             if secret != data:
                 try:
-                    self._scheduler.update_secret(app, name, data)
+                    self._scheduler.update_secret(namespace, name, data)
                 except KubeException as e:
                     msg = 'There was a problem updating the certificate secret ' \
-                          '{} for {}'.format(name, app)
+                          '{} for {}'.format(name, namespace)
                     raise ServiceUnavailable(msg) from e
 
         # get config for the service
-        config = self._load_service_config(app, 'router')
+        config = self._load_service_config(namespace, 'router')
 
         # See if certificates are available
         if 'certificates' not in config:
@@ -206,7 +206,7 @@ class Certificate(AuditedModel):
             certificates.append(cert)
         config['certificates'] = ','.join(certificates)
 
-        self._save_service_config(app, 'router', config)
+        self._save_service_config(namespace, 'router', config)
 
     def detach(self, *args, **kwargs):
         # remove the certificate from the domain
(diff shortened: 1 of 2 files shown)


rootfs/api/models/certificate.py
```diff
@@ -173,7 +173,7 @@ def attach_in_kubernetes(self, domain):
         # only create if it exists - We raise an exception when a secret doesn't exist
         try:
             name = '%s-cert' % self.name
-            app = domain.app
+            app = domain.app.id
```
bacongobbler: should probably change this to `app_id` instead of `app` so we're not confused what this variable holds in future PRs

rootfs/scheduler/__init__.py
```diff
@@ -1023,7 +1016,8 @@ def _default_dockerapp_readiness_probe(self, port, delay=5, timeout=5, period_se
         return readinessprobe
 
     # SECRETS #
```
bacongobbler: double secret comment

rootfs/api/models/certificate.py
helgi: Change it to namespace

rootfs/scheduler/__init__.py
```diff
@@ -1053,25 +1047,39 @@ def get_secrets(self, namespace, **kwargs):
 
         return response
 
-    def create_secret(self, namespace, name, data, secret_type='Opaque', labels={}):
+    def _build_secret_manifest(self, namespace, name, data, secret_type='Opaque', labels=None):
```
bacongobbler: Why is the change to `labels` to a NoneType required? Why not keep it as an empty dictionary?

helgi: It's considered an anti-pattern in Python (from what I can tell)

rootfs/scheduler/__init__.py
helgi: Fixed

rootfs/scheduler/__init__.py
bacongobbler: It just seems unnecessary that we now have to determine the if `labels` is a `dict` type a little further down, whereas before we didn't have to because it was guaranteed to be a dictionary.

helgi: Since Python isn't strongly typed (or we aren't using it that way right now) then we anyway should be doing checks like that if we expect a dict. You can pass in anything as `labels` regardless of what I default it to

helgi: I couldn't find the source of that so I just changed it back for now. Will rather tackle this kind of change if we go strictly typed

PR #866, opened 2016-07-11
fix(registry): use the proxy port and local host to connect to registry

kmala:
this depends on 
https://github.com/deis/dockerbuilder/pull/78
https://github.com/deis/builder/pull/386
https://github.com/deis/charts/pull/305


arschles:
@kmala can you add to the PR description what PRs this one depends on?

slack:
Tested via my comment in https://github.com/deis/charts/pull/305

PR #870, opened 2016-07-13
ref(scheduler): separate deploy batch / step calculations into functions

helgi:


rootfs/scheduler/__init__.py
```diff
@@ -122,22 +122,10 @@ def deploy(self, namespace, name, image, command, **kwargs):  # noqa
...
-        # figure out what kind of batches the deploy is done in - 1 in, 1 out or higher
-        if desired < steps:
-            # do it all in one go
-            batches = [desired]
-        else:
-            # figure out the stepped deploy count and then see if there is a leftover
-            batches = [steps for n in set(range(1, (desired + 1))) if n % steps == 0]
-            if desired - sum(batches) > 0:
-                batches.append(desired - sum(batches))
+        batches = kwargs.get('deploy_batches', None)
```
kmala: should this be deploy_batches or batches as the removed code has it as batches.

helgi: Good catch - I separated this out from the Deployments PR so it changed to `deploy_batches`. Will update this PR to reflect that

helgi: Updated the PR. Changes is in `app.py` where the information is passed in

PR #871, opened 2016-07-13
feat(scheduler): sort env vars and secrets by keys for easier hashing

helgi:
Makes it easier to debug when order stays consistent, especially with something like `deepdiff`


bacongobbler:
hopefully this also solves the random unit test failures :+1:

PR #872, opened 2016-07-13
fix(release): return port from get_port for non-routable process types

helgi:
Env Vars are stored in a Secret, where there is 1 secret per Release. This causes a situation where non-web/cmd proc types will overwrite the Secret data during deploys.
This is the only way possible (for now) since env vars are per Release and not per proc type

### DIFF
--- a/rootfs/api/models/release.py
+++ b/rootfs/api/models/release.py
@@ -134,7 +134,7 @@ class Release(UuidAuditedModel):
         deis_registry = bool(self.build.source_based)
         publish_release(source_image, self.image, deis_registry, self.get_registry_auth())
 
-    def get_port(self, routable=False):
+    def get_port(self):
         """
         Get application port for a given release. If pulling from private registry
         then use default port or read from ENV var, otherwise attempt to pull from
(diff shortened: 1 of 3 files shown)


bacongobbler:
Will this fix the error I'm seeing in https://github.com/deis/controller/pull/822#issuecomment-232209776?

rootfs/api/models/release.py
```diff
@@ -134,7 +134,7 @@ def publish(self):
         deis_registry = bool(self.build.source_based)
         publish_release(source_image, self.image, deis_registry, self.get_registry_auth())
 
-    def get_port(self, routable=False):
```
bacongobbler: glad we're removing the `routable` flag. That always threw me off when making calls to it

bacongobbler:
Either way LGTM

helgi:
@bacongobbler Ah not sure if it fixes #822 (haven't specifically tested it) but I think it may help since it won't constantly be added / removed anymore

Had the best of intentions originally with this function and how it worked but well... the shared config throws off the usefulness :)

PR #880, opened 2016-07-14
fix(scheduler): only attach healthcheck if app is routable

### LINKED ISSUE #879: workers are deployed with a healthcheck
For example, deploying example-dockerfile-procfile-http:

```
><> deis scale worker=1
done in 128s
=== www Processes
--- worker:
www-v2-worker-sy6c2 starting (v2)
--- web:
www-v2-web-dp2h0 up (v2)
```

From checking the pod:

```
readinessProbe:
      failureThreshold: 1
      initialDelaySeconds: 5
      periodSeconds: 5
      successThreshold: 1
      tcpSocket:
        port: 80
      timeoutSeconds: 5
```

The worker process does not run the API server, therefore it will always fail the health check. Non-web processes should not have a healthcheck.

bacongobbler:
fixes #879

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,12 +67,12 @@ test-style:
 
 test-unit:
 	cd rootfs \
-		&& coverage run manage.py test --settings=api.settings.testing --noinput registry api \
+		&& coverage run manage.py test --settings=api.settings.testing --noinput registry api scheduler.tests \
 		&& coverage report -m
 
 test-unit-quick:
 	cd rootfs \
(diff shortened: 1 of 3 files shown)


bacongobbler:
Scheduler module, now with tests!

bacongobbler:
Gonna merge this only because the e2e suite passed, but for some reason the e2e suite couldn't fetch the junit results.

> 14:58:01 ERROR: Step ‘Publish JUnit test result report’ failed: No test report files were found. Configuration error?

PR #882, opened 2016-07-14
feat(scheduler) prepend [namespace] to Scheduler log message for better traceability

helgi:
Look at the e2e controller logs to see it in its glory or do any action locally that triggers a scheduler action, it should now be prepended with `[foobar] hello i am a message`


arschles:
seems like a good addition to me. as we discussed offline, this would bring the namespace into the `deis logs` output, which I believe is a good addition since we generally assume that the namespace is equal to the app name

PR #884, opened 2016-07-14
fix(healthchecks): when migrating from old healthchecks also add in readiness probe

### LINKED ISSUE #874: controller assumes the application is ready based soley on the readiness probe
If the application has a liveness probe, but not a readiness probe the application will immediately become ready.

```
--- Liveness
Initial Delay (seconds): 50
Timeout (seconds): 50
Period (seconds): 10
Success Threshold: 1
Failure Threshold: 3
Exec Probe: N/A
HTTP GET Probe: Path="/servlet/status" Port=8080 HTTPHeaders=[]
TCP Socket Probe: N/A

--- Readiness
No readiness probe configured.
```

I'm not 100% certain rather is is an oddity in k8s or deis, but with a liveness probe and an initial delay of 50 seconds I would expect the application to not be live for a minimum of 50 seconds.  Instead as soon as the pods are in a Running state the old application is terminated.

After creating a readiness probe with an initial delay of 50 seconds it works as expected.

helgi:
Also get the max() between liveness and readiness initial seconds delay to use in deploy timeout. It is smart enough to deal with if one or both are missing as well.

Fixes #874

### DIFF
--- a/rootfs/api/models/config.py
+++ b/rootfs/api/models/config.py
@@ -56,6 +56,20 @@ class Config(UuidAuditedModel):
             }
         }
 
+        self.healthcheck['readinessProbe'] = {
+            'initialDelaySeconds': delay,
+            'timeoutSeconds': timeout,
+            'periodSeconds': period_seconds,
+            'successThreshold': success_threshold,
+            'failureThreshold': failure_threshold,
+            'httpGet': {
+                'path': path,
+            }
+        }
+
+        # Unset all the old values
+        self.values = {k: v for k, v in self.values.items() if not k.startswith('HEALTHCHECK_')}
+
     def set_registry(self):
         # lower case all registry options for consistency
         self.registry = {key.lower(): value for key, value in self.registry.copy().items()}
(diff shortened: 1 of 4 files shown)


bacongobbler:
Could we maybe document this behaviour as well?

helgi:
Do we have any docs yet that talk about timeout delays and such? I assume you are talking about the `max()` portion of the code is what you wanted to see docs on?

bacongobbler:
Yeah exactly. We don't talk much about the timeout delays in the healthcheck docs but I would really appreciate it! I'll give you one Canadian maple leaf in exchange for docs.

helgi:
How about I document that as part of https://github.com/deis/controller/issues/817 ? Planning on getting to that pretty soon

bacongobbler:
sure, not a blocker for this PR

bacongobbler:
...Just a maple leaf on the line is all. :maple_leaf:

slack:
We need to make sure we delete the old HEALTCHECK_ otherwise that'll cause wacky behavior on config:set or controller restarts.

helgi:
Yeah, agreed - Changing tests / code right now

helgi:
That's been fixed now

Issue #374, opened 2016-02-16
Run docker build on all PR builds


arschles:
Similar to what builder does in its [.travis.yml file](https://github.com/deis/builder/blob/master/.travis.yml#L16), workflow should run `make ... docker-build` to ensure that the Dockerfile is correct

Joshua-Anderson:
We run docker build to build the image pre-e2e runs, so this has been resolved.

PR #887, opened 2016-07-15
feat(app): make deploy timeout configurable globally/per-app via DEIS_DEPLOY_TIMEOUT, default is 2 minutes

### LINKED ISSUE #817: Make the main deployment loop timeout value configurable
Instead of the core timeout being 2 minutes then that should be configurable

Note that there are many other timeouts involved *and* the 2 minute timeout is padded by `HEALTHCHECK_INITIAL_DELAY` and if the image download takes 60 seconds then the Controller auto adds a 10 minute buffer on top of it all.

See conversation as part of #792

helgi:
This will behave slightly differently between the old deploy method (RC) and Deployments

RC workflow: `DEIS_DEPLOY_TIMEOUT` refers to the base timeout *per* each `DEIS_DEPLOY_BATCHES`

Deployment workflow: `DEIS_DEPLOY_TIMEOUT` is the base timeout which then gets multipled with `DEIS_DEPLOY_BATCHES` to get an overall timeout

closes #817

[docs](https://github.com/deis/workflow/pull/374)

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -404,6 +404,9 @@ class App(UuidAuditedModel):
         # see if the app config has deploy batch preference, otherwise use global
         batches = release.config.values.get('DEIS_DEPLOY_BATCHES', settings.DEIS_DEPLOY_BATCHES)
 
+        # see if the app config has deploy timeout preference, otherwise use global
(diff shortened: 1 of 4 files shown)


helgi:
https://github.com/deis/workflow/pull/374 is the docs

PR #890, opened 2016-07-16
fix(tests): sort domains in tests to get past occasional ordering problems which cause test failures

### LINKED ISSUE #886: unit tests fail often due to unsorted lists
Tracking this because I'm sick of seeing red travis runs for this simple issue:

```
System check identified no issues (0 silenced).

cd rootfs \

		&& coverage run manage.py test --settings=api.settings.testing --noinput registry api scheduler.tests \

		&& coverage report -m

Creating test database for alias 'default'...

.............................................................F..............................................................................................................................................................................................................................................................................

FAIL: test_manage_idn_domain (api.tests.deployments.test_domain.DomainTest)

  File "/home/travis/build/deis/controller/rootfs/api/tests/deployments/test_domain.py", line 114, in test_manage_idn_domain

    self.assertEqual([self.app_id, ace_domain], expected, msg)

AssertionError: Lists differ: ['candid-organism', 'xn--80ahd1agd.ru'] != ['xn--80ahd1agd.ru', 'candid-organism']

First differing element 0:

candid-organism

xn--80ahd1agd.ru

- ['candid-organism', 'xn--80ahd1agd.ru']

+ ['xn--80ahd1agd.ru', 'candid-organism'] : failed on 'домена.ru'

Ran 332 tests in 605.528s

FAILED (failures=1)

Destroying test database for alias 'default'...

make: *** [test-unit] Error 1
```

helgi:
Fixes #886

### DIFF
--- a/rootfs/api/tests/deployments/test_domain.py
+++ b/rootfs/api/tests/deployments/test_domain.py
@@ -111,7 +111,7 @@ class DomainTest(APITestCase):
             url = '/v2/apps/{app_id}/domains'.format(app_id=self.app_id)
             response = self.client.get(url)
             expected = [data['domain'] for data in response.data['results']]
-            self.assertEqual([self.app_id, ace_domain], expected, msg)
+            self.assertEqual(sorted([self.app_id, ace_domain]), sorted(expected), msg)
 
             # Verify creation failure for same domain with different encoding
             if ace_domain != domain:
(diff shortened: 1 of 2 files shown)


vdice:
e2e success in https://ci.deis.io/job/workflow-test-pr/3667/console; merging.

nathansamson:
In all honesty it would make more sense if the API would return this resultset sorted...

bacongobbler:
I'd argue against that logic simply because it requires more computation from the server side to sort a large list of domains. Personally I believe it's better to make the client do the sorting instead to offload some of the computation there.

nathansamson:
Wel if the client does the sorting that would give the save effect for me.

But still I believe this should be in the database

1) first I would like to see the benchmark that shows a significant slowdown because you sort in the worst case a few hundred domains
2) I am not sure this actually works because the result set is downloaded paginated AFAIK. So letting the sort happen on client side means it needs to wait till the last page before it can display anything, partially defeating the point of doing it paginated.

Anyhow I might actually create a merge request for displaying some sets (eg. Apps as well) paginated, because seeing the everytime in a different order is kind of painful

helgi:
Feel free to open up a new ticket / PR - it's a valid point and should be taken care of in the serializer :)

This particular PR was to solve a problem with our CI failing from time to time with a very small subset of domains. Basically stabilising some tests before going into bigger set of work like what you are talking about

PR #892, opened 2016-07-16
chore(requirements): update Django REST Framework to 3.4.0

helgi:
http://www.django-rest-framework.org/topics/3.4-announcement/

Not making use of any of the new functionalities in this change. Those can come later

Requires deis/workflow-e2e#272


helgi:
e2e actually passed but one of the phases after the actual tests is what failed. Merging

PR #896, opened 2016-07-20
chore(requirements): update Django to 1.9.8

mboersma:
See docs.djangoproject.com/en/1.9/releases/1.9.8/

These changes don't appear to affect Deis Workflow at first glance, so I don't think this is a high priority to merge. It can go in v2.3.0 IMHO.


mboersma:
This passed all tests but failed for some CI reporting technicality, so I'm merging it.

PR #904, opened 2016-07-21
fest(registry):Add support for private registry

kmala:


rootfs/scheduler/__init__.py
```diff
@@ -639,29 +639,40 @@ def _set_image_secret(self, data, namespace, **kwargs):
...
+                hostname, _ = docker_auth.split_repo_name(kwargs.get('image'))
+            username = registry.get('username')
+            password = registry.get('password')
+        elif settings.REGISTRY_LOCATION == 'off-cluster':
+            secret = self.get_secret('deis', 'registry-secret').json()
+            username = secret['data']['username']
+            password = secret['data']['password']
+            hostname = secret['data']['server']
+            if hostname == '':
+                hostname = "https://index.docker.io/v1/"
```
helgi: Why is the hostname hardcoded? this will come to bite us at some point

kmala: hostname is hardcoded only if user doesn't specify it where in we assume it to be dockerhub.

helgi: What if they change hostnames? move to v2 and v1 becomes deprecated / removed?

kmala: like i said its only when they don't specify it as part of the configuration....if anything what you said happens they can just change the details in the param.toml file.

rootfs/scheduler/__init__.py
```diff
@@ -639,29 +639,40 @@ def _set_image_secret(self, data, namespace, **kwargs):
...
         http://kubernetes.io/docs/user-guide/images/#specifying-imagepullsecrets-on-a-pod
         """
+        secret_name = settings.REGISTRY_SECRET_PREFIX
         registry = kwargs.get('registry', {})
-        if not registry:
+        if not registry and settings.REGISTRY_LOCATION == 'on-cluster':
+            return
+        elif registry:
+            # try to get the hostname information
+            hostname = registry.get('hostname', None)
```
helgi: Do we still want to keep this? `hostname` is currently not used but was kept as an option to help instruct `builder` to push to other places (and in turn let Controller know that is happening).

kmala: I had kept it thinking it might be useful in future when we want to expand to supporting per app private registry basis

helgi:
I would like to see some tests in this PR, ideally if you develop some for https://github.com/deis/controller/pull/920 then you can bring them over here as this PR has a fair bit of redness in it (when it comes to new lines and lack of coverage)

bacongobbler:
I agree. We should be holding the controller to a higher standard now that we are out of pre-production.

rootfs/api/models/release.py
```diff
@@ -37,6 +37,8 @@ def __str__(self):
 
     @property
     def image(self):
+        if (settings.REGISTRY_LOCATION != 'on-cluster'):
```
bacongobbler: needs a unit test

rootfs/api/models/release.py
```diff
@@ -148,7 +150,7 @@ def get_port(self):
                 return 5000
 
             # application has registry auth - $PORT is required
-            if creds is not None:
+            if (creds is not None) or (settings.REGISTRY_LOCATION != 'on-cluster'):
```
bacongobbler: needs a unit test

rootfs/scheduler/__init__.py
```diff
@@ -643,15 +643,31 @@ def _set_container(self, namespace, container_name, data, **kwargs):  # noqa
             self._default_readiness_probe(data, kwargs.get('build_type'), env.get('PORT', None))
 
     def _get_private_registry_config(self, registry, image):
-        # try to get the hostname information
-        hostname = registry.get('hostname', None)
-        if not hostname:
-            hostname, _ = docker_auth.split_repo_name(image)
-        if hostname == docker_auth.INDEX_NAME:
-            hostname = "https://index.docker.io/v1/"
+        secret_name = settings.REGISTRY_SECRET_PREFIX
```
bacongobbler: can we remove this? as far as I can tell `secret_name` cannot be changed, so we might as well reference `settings.REGISTRY_SECRET_PREFIX` directly rather than waste memory allocating a local variable.

rootfs/scheduler/__init__.py
```diff
@@ -643,15 +643,31 @@ def _set_container(self, namespace, container_name, data, **kwargs):  # noqa
...
+        registryLocation = settings.REGISTRY_LOCATION
```
bacongobbler: same here, no need to assign a variable for a constant.

rootfs/api/models/release.py
kmala: added

rootfs/api/models/release.py
kmala: added

rootfs/scheduler/__init__.py
kmala: it is being changed if the registry location is off-cluster.

PR #911, opened 2016-07-25
fix(release): old release shouldn't be deleted if new release fails

kmala:
e2e test: https://github.com/deis/workflow-e2e/pull/273

### DIFF
--- a/rootfs/api/models/release.py
+++ b/rootfs/api/models/release.py
@@ -235,7 +235,7 @@ class Release(UuidAuditedModel):
     def delete(self, *args, **kwargs):
         """Delete release DB record and any RCs from the affect release"""
         try:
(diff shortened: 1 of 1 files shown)


rootfs/api/models/release.py
```diff
@@ -356,7 +356,7 @@ def _delete_release_in_scheduler(self, namespace, version):
         labels = {
             'heritage': 'deis',
             'app': namespace,
-            'version': version
+            'version': "v{}".format(version)
```
helgi: Look at how `_delete_release_in_scheduler` is used. It seems in some cases it gets a pure version (lacking the `v`) and in other cases it is getting the version from kubernetes labels so it would include the `v`

PR #912, opened 2016-07-25
fix(registry): User should be able to pull the image from internal registry

### LINKED ISSUE #894: deis pull of image from internal registry should work
You should be able to do a `deis pull <regsitry-ip>:<port>/iamge:tag` basically pull an image already in registry.
we need to remove the striping logic https://github.com/deis/controller/blob/64547741ae6e80f8299992fc57b87d9c10d2c5bd/rootfs/api/models/release.py#L58-L59 as we had removed the registry appending logic in https://github.com/deis/controller/pull/716

kmala:
fixes #894

### DIFF
--- a/rootfs/api/models/release.py
+++ b/rootfs/api/models/release.py
@@ -54,9 +54,7 @@ class Release(UuidAuditedModel):
             self.build.image.startswith(settings.REGISTRY_HOST) or
             self.build.image.startswith(settings.REGISTRY_URL)
         ):
-            # strip registry information off first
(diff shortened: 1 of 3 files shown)


bacongobbler:
there's a typo in the commit title: should be "registry". This won't impact us in any way, right?

mboersma:
Could you amend the second commit title? Maybe:
```
tests(build): fix for internal registry access
```

kmala:
all the tests are passing https://ci.deis.io/job/workflow-test-pr/4108/console

bacongobbler:
Doesn't look like Jenkins agrees with that ^ ;)

PR #914, opened 2016-07-26
fix(pull): release and build should be deleted on failed deis pull

kmala:
e2e: https://github.com/deis/workflow-e2e/pull/273


helgi:
Did you move this inside the `try` block to catch failures in `release.new` or to expose `new_release` variable inside the `except` block?

kmala:
@helgi to catch the failures in the `release.new` which occurs during an pulling of invalid image

PR #926, opened 2016-07-28
fix(validation): success threshold can only be equal to one when setting liveness probe. Error properly

### LINKED ISSUE #873: liveness probe success threshold
When setting success-threshold to anything, but 1 deis errors.

```
 deis healthchecks:set liveness httpGet -p /status --success-threshold=3 -- 8080
```

```
Applying livenessProbe healthcheck... Error: Unknown Error (400): map[detail:app-v47-cmd (app::deploy): ('failed to create ReplicationController "app-v47-cmd" in Namespace "app": 422 status code 422\n{\'status\': \'Failure\', \'metadata\': {}, \'message\': \'ReplicationController "app-v47-cmd" is invalid: spec.template.spec.containers[0].livenessProbe.successThreshold: Invalid value: 3: must be 1\', \'reason\': \'Invalid\', \'code\': 422, \'kind\': \'Status\', \'apiVersion\': \'v1\', \'details\': {\'causes\': [{\'field\': \'spec.template.spec.containers[0].livenessProbe.successThreshold\', \'reason\': \'FieldValueInvalid\', \'message\': \'Invalid value: 3: must be 1\'}], \'kind\': \'ReplicationController\', \'name\': \'app-v47-cmd\'}}', 'app-v47-cmd', 'app')]
```

helgi:
Kubernetes has this restriction, catch it before it ever makes it that far.

Introducing jmespath now to later use in other places as well

Fixes #873

### DIFF
--- a/rootfs/api/serializers.py
+++ b/rootfs/api/serializers.py
@@ -3,6 +3,7 @@ Classes to serialize the RESTful representation of Deis API models.
 """
 
 import json
+import jmespath
 import re
 import jsonschema
 import idna
@@ -336,6 +337,15 @@ class ConfigSerializer(serializers.ModelSerializer):
             except jsonschema.ValidationError as e:
                 raise serializers.ValidationError(
                     "could not validate {}: {}".format(value, e.message))
+
+        # http://kubernetes.io/docs/api-reference/v1/definitions/#_v1_probe
+        # liveness only supports successThreshold=1, no other value
+        # This is not in the schema since readiness supports other values
+        threshold = jmespath.search('livenessProbe.successThreshold', data)
+        if threshold is not None and threshold != 1:
+            raise serializers.ValidationError(
+                'livenessProbe successThreshold can only be 1'
+            )
 
         return data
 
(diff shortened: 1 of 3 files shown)


helgi:
```
deis healthchecks:set -a test5 liveness httpGet -p /status --success-threshold=3 -- 8080
Applying livenessProbe healthcheck... Error: Unknown Error (400): map[healthcheck:[livenessProbe successThreshold can only be 1]]
```

rootfs/api/serializers.py
```diff
@@ -337,6 +338,15 @@ def validate_healthcheck(self, data):
                 raise serializers.ValidationError(
                     "could not validate {}: {}".format(value, e.message))
 
+        # http://kubernetes.io/docs/api-reference/v1/definitions/#_v1_probe
```
kmala: i think you can validate this using jsonschema itself
"successThreshold": { "enum": [ 1 ] },

helgi: I looked and yes it is possible but then that would apply to both `readiness` and `liveness` probes.

Only `liveness` has this restriction, `readiness` appears not to have it

PR #929, opened 2016-07-28
chore(api): bump api version to v2.2.0

mboersma:
The auth/whoami endpoint was added.

### DIFF
--- a/rootfs/api/__init__.py
+++ b/rootfs/api/__init__.py
@@ -2,4 +2,4 @@
 The **api** Django app presents a RESTful web API for interacting with the **deis** system.
 """
 
-__version__ = '2.1.0'
+__version__ = '2.2.0'


bacongobbler:
refs https://github.com/deis/workflow/issues/412

bacongobbler:
note that the SDK PR is at https://github.com/deis/controller-sdk-go/pull/61

helgi:
`DEIS_PLATFORM_VERSION` is now based on this version; Should we follow lock step with component version or should API version be what it used to be up till now?

If we do the latter we will have a 2.2.0 for 2.3 and maybe even 2.4 of the workflow / component release

mboersma:
> Should we follow lock step with component version or should API version be what it used to be up till now?

I think we let the API version stay at 2.2.0 until it needs to change, regardless of workflow release tag. I could be persuaded otherwise.

helgi:
Given we now use it for `DEIS_PLATFORM_VERSION` then I think bumping version feels more natural, unless we totally want to deprecate `DEIS_PLATFORM_VERSION` and from 3.0 onward only communicate the API version (major.minor and not patch). Also I think if we are constantly (as in every release or more) changing the API version then we are more inclined to update the right docs, etc. 

That's a bit of a sore spot where we have to remember to update a less known documentation (API doc) part of the overall docs.

bacongobbler:
I agree with @mboersma's comment that we should let the API version stay regardless of workflow product release, however I do believe we are in a bit of a pickle at the moment given that we have released a v2.2.1 controller with an API version of v2.1. This is because of the old differentiation of platform and API version in the controller, but now that we are doing individual product releases the platform version is unused.

Since we've released v2.2.1 with an API version of v2.1, perhaps we should merge this and cut a v2.2.2 for the controller? That way we get the best of both worlds: the controller and tagged release versions are in parity (2.2.x), and we can cut the controller separate from the Workflow product releases. Does that sem fair?

helgi:
When we do a 2.3.x of the Controller but doesn't have public facing API changes, we'd still say the API version is 2.2 instead of 2.3?

bacongobbler:
I guess there would be circumstances where we could release a 2.3 without bumping the API version, but yes, we'd still say the API version is 2.2.

helgi:
Closing in favour of #943

PR #930, opened 2016-07-28
ref(*): update DEIS_PLATFORM_VERSION to be identical to DEIS_API_VERSION since components now have an independant release cycle

helgi:


Joshua-Anderson:
I think the change is good, but breaking backwards compatibility is a little concerning. That being said, I would hope that people would be using api version rather than platform version for checks.

We may want to expose the controller version, because if there is a bug in a specific release, people's apis might want to be able to account for it.

bacongobbler:
Yeah personally I'd keep this in for backwards compatibility reasons, especially since it's still useful for features like https://github.com/deis/workflow-cli/issues/25

helgi:
We have to key it off something else then since we are not longer releasing Controller lockstep with Workflow.

My suggestion was to use `api.__version__` and bump that every time we do a release - That would cause API versions that basically have no public changes but would hold the overall API (Controller) to a single release

If we did that we could keep this header intact, albeit only showing the minor release (like the API version) or we could go even more granular, but even that to me is its own security risk

helgi:
I'm overall open to ideas - All I know is that right now the platform version would never stay in true sync with the world, given we are talking about a potential 2.4 Controller release for the 2.3 Workflow release :)

helgi:
Yeah sounds like a plan - I amended a commit locally (will fix commit message) already to work that way. I wanted to try to pull in the git sha if people did a checkout but didn't get far

rootfs/api/middleware.py
```diff
@@ -46,5 +46,7 @@ def process_response(self, request, response):
         a response header.
         """
         # clients shouldn't care about the patch release
-        response['DEIS_API_VERSION'] = __version__.rsplit('.', 1)[0]
+        version = __version__.rsplit('.', 1)[0]
+        response['DEIS_API_VERSION'] = version
+        response['DEIS_PLATFORM_VERSION'] = version
```
bacongobbler: Sorry, I forgot that the api version is short-form. I'd say that we should make PLATFORM_VERSION use `__version__` without the rsplit so we know the patch release running. For the API we don't really care about patch releases but it would be handy for `deis version --all` in the future.

mboersma:
> Okay let's keep both the API and PLATFORM headers available, but let's just change the two to be identical.

I think that's the best compromise that maintains backward-compatibility. 👍

rootfs/api/middleware.py
helgi: I can do that, tho do we really want to expose the exact version of the Controller? Feels like a security issue to me

Joshua-Anderson: It could be used to scan for vulnerable deis installs, but it also can be useful for checking if the controller has a required bugfix. If you think it's a large concern, we could only expose the header to authenticated requests.

bacongobbler: considering we did that before, I don't see an issue

helgi: I know we did do that before but that's not always the best reason to keep doing it either :-)

I'll probably just do the full version then and hide it behind auth another time

helgi: I updated it to use the full version

PR #931, opened 2016-07-29
fix(models): add more sensible ordering to various API responses

helgi:


helgi:
Let me know if any of those ordering fields need to be adjusted

PR #932, opened 2016-07-29
chore(requirements): Update to DRF 3.4.1

helgi:
http://www.django-rest-framework.org/topics/release-notes/#341

Requires deis/workflow-e2e#279


bacongobbler:
ping @Joshua-Anderson, as I think this also affects the SDK: https://github.com/deis/controller-sdk-go/blob/e1894e9b89ddc961bf12b543d60be9ebf32cae7d/errors.go#L32

Joshua-Anderson:
It does, I'll make a PR for it soon.

PR #934, opened 2016-07-29
feat(models): add routable flag to App

bacongobbler:
If an app is considered non-routable, it is removed from the router mesh.

refs https://github.com/deis/deis/issues/4391
refs https://github.com/deis/deis/issues/2715

TODO:

 - [x] add docs
 - [x] add to the SDK
 - [x] add a --routable flag in the CLI


helgi:
In its current implementation, there shouldn't be any problem doing a `POST` and updating things from routable True -> False or the other way around. I don't see anything that should prevent it but wanted to make sure

bacongobbler:
In that case we should make it part of the config. I was under the impression that we never wanted to change this field after creating the app. I'll make the change.

bacongobbler:
removing from 2.3 because I know there won't be enough time to get this in.

PR #938, opened 2016-08-01
feat(apps): scale / deploy run all process types in parallel instead of sequantially

helgi:
This also moves the restart function to use api.utils.async_run  - it is a slightly modified version of the original to be exception aware and not require an internal function

Eventually I may move us to use https://github.com/observerss/aioutils but right now that is not required


helgi:
Getting https://github.com/deis/controller/pull/939 reviewed first but getting a review on this would be good. The first commit is the what this PR is all about

Issue #424, opened 2016-02-26
Connection closed error when pushing build


jchauncey:
```
╰─± git push deis master
Counting objects: 106, done.
Delta compression using up to 8 threads.
Compressing objects: 100% (71/71), done.
Writing objects: 100% (106/106), 21.10 KiB | 0 bytes/s, done.
Total 106 (delta 57), reused 74 (delta 30)
remote: 2016/02/26 21:09:33 Running in debug mode
Running git hook
read [0000000000000000000000000000000000000000,6df4194662366c3d586eff8e086f3803fd3e1f54,refs/heads/master]
Workflow request /v2/hooks/push (body elided)
Workflow request POST /v2/hooks/config
{"receive_user":"jchauncey","receive_repo":"foo"}
got the following config back for app foo: {Owner:jchauncey App:foo Values:map[] Memory:map[] CPU:map[] Tags:map[] UUID:a6f78cf7-f131-4d36-b17e-43fca5288630 Created:2016-02-26 20:43:59 +0000 UTC Updated:2016-02-26 20:43:59 +0000 UTC}
running [git archive --format=tar.gz --output=foo.tar.gz 6df41946] in directory /home/git/foo.git
running [tar -xzf foo.tar.gz -C /home/git/foo.git/build/tmp691074999/] in directory /home/git/foo.git
Uploading tar to http://10.3.0.197:9000/git/home/foo:git-6df41946/tar
Starting build... but first, coffee!
Starting pod slugbuild-foo-6df41946-eb733769
Pod spec: {
  "metadata": {
    "name": "slugbuild-foo-6df41946-eb733769",
    "namespace": "deis",
    "creationTimestamp": null,
    "labels": {
      "heritage": "deis",
      "version": "2.0.0-beta"
    }
  },
  "spec": {
    "volumes": [
      {
        "name": "minio-user",
        "secret": {
          "secretName": "minio-user"
        }
      }
    ],
    "containers": [
      {
        "name": "deis-slugbuilder",
        "image": "quay.io/deisci/slugbuilder:v2-beta",
        "env": [
          {
            "name": "DEBUG",
            "value": "1"
          },
          {
            "name": "TAR_URL",
            "value": "http://10.3.0.197:9000/git/home/foo:git-6df41946/tar"
          },
          {
            "name": "put_url",
            "value": "http://10.3.0.197:9000/git/home/foo:git-6df41946/push"
          }
        ],
        "resources": {},
        "volumeMounts": [
          {
[...]

PR #940, opened 2016-08-02
feat(healthcheck): Add support to create healthcheck per proctype

kmala:
Requires workflow-cli#160

### DIFF
--- a/rootfs/api/models/config.py
+++ b/rootfs/api/models/config.py
@@ -46,7 +46,8 @@ class Config(UuidAuditedModel):
         success_threshold = int(self.values.get('HEALTHCHECK_SUCCESS_THRESHOLD', 1))
         failure_threshold = int(self.values.get('HEALTHCHECK_FAILURE_THRESHOLD', 3))
 
-        self.healthcheck['livenessProbe'] = {
+        self.healthcheck['web/cmd'] = {}
+        self.healthcheck['web/cmd']['livenessProbe'] = {
             'initialDelaySeconds': delay,
             'timeoutSeconds': timeout,
             'periodSeconds': period_seconds,
@@ -57,7 +58,7 @@ class Config(UuidAuditedModel):
             }
         }
 
-        self.healthcheck['readinessProbe'] = {
+        self.healthcheck['web/cmd']['readinessProbe'] = {
             'initialDelaySeconds': delay,
             'timeoutSeconds': timeout,
             'periodSeconds': period_seconds,
@@ -70,6 +71,12 @@ class Config(UuidAuditedModel):
 
         # Unset all the old values
         self.values = {k: v for k, v in self.values.items() if not k.startswith('HEALTHCHECK_')}
+
+    def get_healthcheck(self):
+        if('livenessProbe' in self.healthcheck.keys() or
+           'readinessProbe' in self.healthcheck.keys()):
+            return {'web/cmd': self.healthcheck}
+        return self.healthcheck
 
     def set_registry(self):
         # lower case all registry options for consistency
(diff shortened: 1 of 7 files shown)


rootfs/api/models/config.py
```diff
@@ -71,6 +72,12 @@ def _migrate_legacy_healthcheck(self):
         # Unset all the old values
         self.values = {k: v for k, v in self.values.items() if not k.startswith('HEALTHCHECK_')}
 
+    def get_healthcheck(self):
+        if 'livenessProbe' in self.healthcheck.keys() or 'readinessProbe' in self.healthcheck.keys():  # noqa
+            self.app.log(self.healthcheck.keys())
```
helgi: What is this log for?

rootfs/api/models/config.py
```diff
@@ -46,7 +46,8 @@ def _migrate_legacy_healthcheck(self):
         success_threshold = int(self.values.get('HEALTHCHECK_SUCCESS_THRESHOLD', 1))
         failure_threshold = int(self.values.get('HEALTHCHECK_FAILURE_THRESHOLD', 3))
 
-        self.healthcheck['livenessProbe'] = {
+        self.healthcheck[''] = {}
```
helgi: Empty is a little weird, but we are only doing that to have a "global" proc type (so to speak)? Can we use `__all` or something similar that our proc name verifier would otherwise disallow when a user passes it in?

rootfs/api/serializers.py
```diff
@@ -326,27 +326,30 @@ def validate_registry(self, data):
         return data
 
     def validate_healthcheck(self, data):
-        for key, value in data.items():
-            if value is None:
+        for procType, healthcheck in data.items():
+            if healthcheck is None:
                 continue
+            else:
```
helgi: Else is not needed here, drop that and bring everything in one level would be my recommendation

helgi:
I'm seeing many red (not covered by tests) line from codecov - Would want to get that fixed before we merge this

rootfs/api/models/config.py
kmala: how about `web/cmd` as @slack wan't it to be just for those two proctypes.https://github.com/deis/workflow/pull/424#discussion_r74125645

helgi: I could be fine with that, and also `_all` or something that makes it descriptive but unique and not usable via Procfile or submitting it through our system in general

rootfs/api/models/config.py
```diff
@@ -71,6 +72,11 @@ def _migrate_legacy_healthcheck(self):
         # Unset all the old values
         self.values = {k: v for k, v in self.values.items() if not k.startswith('HEALTHCHECK_')}
 
+    def get_healthcheck(self):
+        if 'livenessProbe' in self.healthcheck.keys() or 'readinessProbe' in self.healthcheck.keys():  # noqa
```
helgi: You can add `()` around this here and split across lines to get past `# noqa`

rootfs/api/models/config.py
```diff
@@ -71,6 +72,11 @@ def _migrate_legacy_healthcheck(self):
...
+            return {'web/cmd': self.healthcheck}
```
helgi: tests never reach this line

rootfs/api/models/config.py
```diff
@@ -105,6 +111,40 @@ def set_tags(self, previous_config):
 
         raise DeisException(message)
 
+    def set_healthcheck(self, previous_config):
+        data = getattr(previous_config, 'healthcheck', {}).copy()
+        new_data = getattr(self, 'healthcheck', {}).copy()
+        # update the config data for healthcheck if they are not
+        # present for per proctype
+        if 'livenessProbe' in data.keys() or 'readinessProbe' in data.keys():
```
helgi: tests never reach into this if statement

rootfs/api/models/config.py
```diff
@@ -105,6 +111,40 @@ def set_tags(self, previous_config):
...
+            prev_data = data.copy()
+            data = {}
+            data['web/cmd'] = prev_data
+        if 'livenessProbe' in new_data.keys() or 'readinessProbe' in new_data.keys():  # noqa
```
helgi: tests never reach into this if statement

rootfs/api/models/config.py
kmala: this code gets called only during the load from previous controller release data,i can't insert data as serializers will fail.
same for below.

helgi: Multiple releases would not trigger this?

kmala: the function gets called but the if doesn't pass as keys can't be probes unless it is read from the database.

rootfs/api/models/config.py
```diff
@@ -105,6 +112,40 @@ def set_tags(self, previous_config):
...
+            lat_data = new_data.copy()
+            new_data = {}
+            new_data['web/cmd'] = lat_data
```
helgi: this can be changed to a `new_data = {'web/cmd': new_data.copy()}` - Same with the above

rootfs/api/models/config.py
helgi: Guess I am not following why it can't be tested. Why will the serializers fail?

kmala: this code is present to provide backward compatibility for the users who already set using previous versions of controller and are updating the controller with existing database.

helgi: Ah that is what you mean by serializer validation, can't even object create things direclty anymore - Please add some comments to that effect and specifically that this can be removed in the next major version, since we've been adding `# TODO` like comments around the code base

PR #941, opened 2016-08-02
chore(requirements): move from django-cors-headers to django-cors-middleware

helgi:
django-cors-headers is inactive and django-cors-middleware is a fork of the former


helgi:
See https://pypi.python.org/pypi/django-cors-middleware and https://github.com/zestedesavoir/django-cors-middleware

PR #946, opened 2016-08-03
fix(scheduler): create application config (env secrets) outside of deploy / scale to reduce k8s thrasing

helgi:
Prior to this the application config was being created inside `_set_container` which gets called for every proc type, yet we share this single config across all proc types (currently)

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -389,7 +389,9 @@ class App(UuidAuditedModel):
 
     def _scale_pods(self, scale_types):
         release = self.release_set.latest()
-        envs = release.config.values
+        version = "v{}".format(release.version)
+        image = release.image
+        envs = self._build_env_vars(release.build.type, version, image, release.config.values)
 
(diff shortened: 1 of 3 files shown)


rootfs/api/models/app.py
```diff
@@ -775,3 +785,27 @@ def _scheduler_filter(self, **kwargs):
...
+        # mix in default environment information deis may require
+        default_env = {
+            'DEIS_APP': self.id,
+            'WORKFLOW_RELEASE': version
+        }
+
+        # Check if it is a slug builder image.
+        if build_type == 'buildpack':
+            # overwrite image so slugrunner image is used in the container
+            default_env['SLUG_URL'] = settings.SLUGRUNNER_IMAGE
```
kmala: this is not slugrunner_image....this should be the path of the app slug to be run

helgi: yeah, copy paste error for sure. Let me fix up

PR #947, opened 2016-08-03
Cleanup tests and start splitting them out

helgi:
Simplify how apps are created (reduce repeat logic), move to using more unique app names instead of hardcoding when possible (reduce clashes if ran in parallel) and split out the config test into smaller pieces


slack:
Looks like this reduced test coverage by 0.12%. RED BUILD!

PR #955, opened 2016-08-05
fix(middleware): move to 1.10 style middleware

helgi:
Waiting for https://github.com/zestedesavoir/django-cors-middleware/pull/25 to land as we are blocked on that for now


helgi:
Made the Certificate fields UTC only, otherwise there are bunch of warnings. It's a bit of an assumption but it's not very time sensitive information.

Also seems the way a property is now dealt with a bit differently so changed that as well

PR #961, opened 2016-08-09
fix(health):Add healthchecks only for routable apps or proctypes

kmala:


rootfs/scheduler/__init__.py
```diff
@@ -585,19 +585,20 @@ def _set_container(self, namespace, container_name, data, **kwargs):
...
-            if (
-                healthchecks.get('livenessProbe') is not None and
-                healthchecks['livenessProbe'].get('httpGet') is not None and
-                healthchecks['livenessProbe']['httpGet'].get('port') is None
-            ):
-                healthchecks['livenessProbe']['httpGet']['port'] = env['PORT']
-            data.update(healthchecks)
-        else:
-            self._default_readiness_probe(data, kwargs.get('build_type'), env.get('PORT', None))
+        if kwargs.get('routable', False):
```
helgi: What is this changing beside breaking up the if statement?

bacongobbler: from what I can read, routable *must* be enabled in order for a readiness probe to be set. In the old code, you could have both no healthchecks and routable disabled but it'd get a readiness probe.

kmala: yes.you shouldn't even get default readiness probe if its not routable

helgi: Ah yeah, I see now. What about Build Packs? They don't need port, it's only pull apps that need that

kmala: since the apps are not routable they shouldn't be having any readiness checks as it doesn't make any sense because readiness checks just makes sure that service endpoints are updated.
I will make another change with the health checks per proctype where an app can have a healthcheck even though it is not routable but we don't add a default healthcheck.

helgi: Being routable and being health checked are two different concerns, that's why I liked the default check. Tho of course makes much less sense if build pack apps are the only ones that can be checked for health (as a default) (=

bacongobbler:
any way we can get tests for this?

PR #965, opened 2016-08-09
feat(scheduler): add scheduler tests that do not go through normal api tests

helgi:
This expands the scheduler test suite and can run outside of the API setup. Brings is simple testing per resource type.

Also fixes a few mock and scheduler discrepencies that were discovered in the testing

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -31,11 +31,10 @@ class KubeHTTPException(KubeException):
         self.response = response
 
         msg = errmsg.format(*args)
-        msg = "failed to {}: {} {}\n{}".format(
+        msg = "failed to {}: {} {}".format(
             msg,
             response.status_code,
-            response.reason,
-            response.json()
+            response.reason
         )
         KubeException.__init__(self, msg, *args, **kwargs)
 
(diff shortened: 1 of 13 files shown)


helgi:
There are more tests to write here, just PRing this for some visual codecov.io response and to get visibility on the work

rootfs/scheduler/__init__.py
```diff
@@ -1841,6 +1839,10 @@ def _build_deployment_manifest(self, namespace, name, image, entrypoint, command
         return manifest
 
     def _scale_deployment(self, namespace, name, image, entrypoint, command, **kwargs):
+        """
+        A conveniance wrapper around Deployment update that does a little bit of introspection
```
arschles: s/conveniance/convenience

rootfs/scheduler/utils.py
```diff
@@ -1,4 +1,65 @@
...
+        'sailfish', 'sandwich', 'sculptor', 'seashore', 'seedcake', 'skylight', 'stickpin',
+        'tabletop', 'tailbone', 'teamwork', 'teaspoon', 'tinkerer', 'traverse', 'turbojet',
+        'umbrella', 'underdog', 'undertow', 'unicycle', 'universe', 'uptowner', 'utensils',
+        'vacation', 'vagabond', 'valkyrie', 'variable', 'villager', 'vineyard', 'vocalist',
+        'waggoner', 'waxworks', 'waterbed', 'wayfarer', 'whitecap', 'windmill', 'woodshed',
+        'yachting', 'yardbird', 'yardwork', 'yearbook', 'yearling', 'yeomanry', 'yodeling',
+        'zaniness', 'zeppelin', 'ziggurat', 'zillions', 'zirconia', 'zoologer', 'zucchini',
+    ]
+    return "{}-{}".format(
+        random.choice(adjectives), random.choice(nouns))
```
arschles: was this moved from somewhere else?

helgi: yes, from api.utils, it's a copy of it - trying to keep scheduler unaware of api

arschles:
Test code looks good, and I'm ok with falling short of the coverage target, since this goes toward that goal

PR #966, opened 2016-08-10
feat(model): Add new model to store settings

kmala:


rootfs/api/migrations/0012_auto_20160810_0315.py
```diff
@@ -0,0 +1,73 @@
+# -*- coding: utf-8 -*-
```
helgi: Should probably try to split this into 2 migrations, the stuff that's coming from all those other models and then yours.

Maybe make a separate PR for that, or i can

helgi:
Why is this a whole new model instead of under the Config model?

rootfs/api/migrations/0012_auto_20160810_0315.py
helgi: https://github.com/deis/controller/pull/967

kmala:
Currently App maps to release and release object is aggregation of config object and build object. So any change to the config or the build would trigger a new release and hence a new deploy of the app and we store all the configuration values in the config object.
But sometimes we may not need/want to create a release for few configurations changes like (enable or disable of maintenance mode or setting value for the deployments strategy or setting KUBERNETES_DEPLOYMENTS_REVISION_HISTORY_LIMIT(https://github.com/deis/controller/blob/master/rootfs/api/models/app.py#L473))
So i had created a new model appsettings which maps to App model and stores different configuration data which doesn't need to do a release.

rootfs/api/models/__init__.py
```diff
@@ -142,6 +143,12 @@ def _log_config_updated(**kwargs):
     config.app.log("config {} updated".format(config))
 
 
+def _log_app_settings_updated(**kwargs):
+    appSettings = kwargs['instance']
+    # log only to the controller; this event will be logged in the release summary
+    appSettings.app.log("application setings {} updated".format(appSettings))
```
bacongobbler: typo: settings

rootfs/api/models/app.py
```diff
@@ -822,3 +833,20 @@ def _build_env_vars(self, build_type, version, image, envs):
         # merge envs on top of default to make envs win
         default_env.update(envs)
         return default_env
+
+    def set_maintenance(self, maintenance):
```
helgi: Debating if this should live here or in the AppSettings model. What's the benefit of keeping it in `App`?

rootfs/scheduler/__init__.py
```diff
@@ -259,13 +261,14 @@ def _get_deploy_batches(self, steps, desired):
...
 
-    def _update_application_service(self, namespace, name, app_type, port, routable=False):
+    def _update_application_service(self, namespace, name, app_type, port, routable=False, maintenance=False):  # noqa
         """Update application service with all the various required information"""
         service = self.get_service(namespace, namespace).json()
         old_service = service.copy()  # in case anything fails for rollback
 
         try:
             # Update service information
+            service['metadata']['annotations']['router.deis.io/maintenance'] = str(maintenance)
```
helgi: You already are doing this with `set_maintenance` what's the benefit here?

kmala: if the user sets the appsettings first and then does a deploy set_maintenance doesn't get called..set_maintenance is called only during the save of the record and if the value is changed.

rootfs/api/models/app.py
kmala: not much..this can be moved.

helgi: Lets just move it, keep the models focused and tidy

rootfs/scheduler/__init__.py
helgi: Currently the `Service` is always created when the application is, so the resource is always around, that should mean the `Service` is ready to go before a deploy. That's how `Domains` work, so I'm just making sure things are all in sync here.

helgi:
Don't forget to add this model into the `load_db_to_k8s.py`

kmala:
doesn't `load_db_to_k8s.py` of the app take care of the setting? am i missing something?

helgi:
Look at https://github.com/deis/controller/blob/master/rootfs/api/management/commands/load_db_state_to_k8s.py#L50

rootfs/api/models/appsettings.py
```diff
@@ -0,0 +1,63 @@
...
+            if new_maintenance is None and prev_maintenance is not None:
+                setattr(self, 'maintenance', prev_maintenance)
+            elif prev_maintenance != new_maintenance:
+                self.set_maintenance(new_maintenance)
+                summary += "{} changed maintenance mode from {} to {}".format(self.owner, prev_maintenance, new_maintenance)  # noqa
+        except Exception as e:
+            self.delete()
+            raise DeisException(str(e)) from e
+
+        if not summary and previous_settings:
```
bacongobbler: why are we checking for a summary, but not doing anything with it? Shouldn't we just check if the current settings is != old settings?

kmala: i wanna log the summary but i think i missed that.

rootfs/api/tests/test_maintenance.py
```diff
@@ -0,0 +1,66 @@
+import requests_mock
```
bacongobbler: can you rename this file to `test_app_settings.py` since that is the object we're testing here?

rootfs/api/tests/test_maintenance.py
```diff
@@ -0,0 +1,66 @@
...
+from django.core.cache import cache
+from django.contrib.auth.models import User
+from rest_framework.authtoken.models import Token
+
+from api.models import App
+from api.tests import adapter, DeisTransactionTestCase
+
+
+@requests_mock.Mocker(real_http=True, adapter=adapter)
+class TestMaintenance(DeisTransactionTestCase):
```
bacongobbler: same here, can we change this to `TestAppSettings`

bacongobbler:
I see all the tests for maintenance mode, but could we please add unit tests for all the intricate edge cases in the AppSettings.save() function? For example, when the summary is empty and previous_settings is not none, we delete the current settings and raise AlreadyExists?

rootfs/scheduler/__init__.py
```diff
@@ -176,13 +177,14 @@ def _get_deploy_batches(self, steps, desired):
 
         return batches
 
-    def _update_application_service(self, namespace, name, app_type, port, routable=False):
+    def _update_application_service(self, namespace, name, app_type, port, routable=False, maintenance=False):  # noqa
```
bacongobbler: perhaps we should rewrite this to read for an `annotations` kwarg and attach that to the service metadata instead of making this an N-length method, where N is the number of supported annotations in the router.

rootfs/scheduler/__init__.py
```diff
@@ -89,6 +89,7 @@ def deploy(self, namespace, name, image, entrypoint, command, **kwargs):  # noqa
 
         app_type = kwargs.get('app_type')
         routable = kwargs.get('routable', False)
```
bacongobbler: same thought as below IRT an `annotations` kwarg.

rootfs/api/views.py
```diff
@@ -296,6 +296,11 @@ def restart(self, *args, **kwargs):
         return Response(pagination, status=status.HTTP_200_OK)
 
 
+class AppSettingsViewSet(AppResourceViewSet):
```
bacongobbler: we need to implement a `get_object` method or the only way to retrieve an AppSetting will be through the UUID. See the CertificateViewSet's `get_object` method for an example.

rootfs/scheduler/__init__.py
kmala: currently there is only one annotation and one label which is updated through controller.

helgi: Could use https://github.com/deis/controller/blob/master/rootfs/api/models/__init__.py#L52-L94 if we want to go as far as naming it after a particular component - That's what `Domains` currently uses

kmala:
The test are already present to test for AlreadyExists and  previous_settings is not None

rootfs/api/models/app.py
```diff
@@ -389,6 +394,7 @@ def scale(self, user, structure):  # noqa
 
     def _scale_pods(self, scale_types):
         release = self.release_set.latest()
+        appSettings = self.appsettings_set.latest()
```
bacongobbler: this use_underscores for the variable name rather than camelCase.

rootfs/api/models/app.py
```diff
@@ -23,6 +23,7 @@
 from api.models.release import Release
 from api.models.config import Config
 from api.models.domain import Domain
+from api.models.appsettings import AppSettings
```
bacongobbler: I almost feel as if this should be part of another package dir...

```
from api.models.app.settings import Settings
```

Any thoughts on the reasoning behind this package naming scheme? I mean it's directly tied to the app model...

bacongobbler: this is getting into bikeshed territory so disregard.

rootfs/scheduler/__init__.py
kmala: `routable` is added as label not as annotation

bacongobbler:
> The test are already present to test for AlreadyExists and previous_settings is not None

I'm sorry but I don't see any `assertRaises` calls for either `AlreadyExists` or `DeisException` in `test_maintenance.py`, nor any tests that the setting was deleted when said if block was encountered. Where are these unit tests?

rootfs/api/models/appsettings.py
bacongobbler: m0ar tests

rootfs/scheduler/__init__.py
bacongobbler: correct, but I'm referring to other annotations such as IP whitelisting in https://github.com/deis/controller/issues/240.

kmala:
I am checking for the response codes AlreadyExists(409)

bacongobbler:
Cool, I missed that. Sorry. I was looking for the error itself.

rootfs/api/models/appsettings.py
```diff
@@ -0,0 +1,65 @@
...
+            self.delete()
+            raise AlreadyExists("{} changed nothing".format(self.owner))
+        self.app.log('summary of app setting changes: {}'.format(summary), logging.DEBUG)
```
bacongobbler: nitpick but I think we should just log the summary in the app logs, similar to when we [dump the release summary](https://github.com/deis/controller/blob/master/rootfs/api/models/__init__.py#L135-L136).

rootfs/api/models/__init__.py
```diff
@@ -142,6 +143,12 @@ def _log_config_updated(**kwargs):
     config.app.log("config {} updated".format(config))
 
 
+def _log_app_settings_updated(**kwargs):
+    appSettings = kwargs['instance']
+    # log only to the controller; this event will be logged in the release summary
+    appSettings.app.log("application settings {} updated".format(appSettings))
```
bacongobbler: this is logged also to the application, which means the user will see this log as well as the summary. Do you mean to call `logger.info()` instead?

kmala: This just says that settings have been updated but not what all has changed. app.log will call logger.log as with level as info
If Debug mode is on then user will see the full details of what all has changed which is what i had done above.

bacongobbler:
if you wanted to move `routable` to appsettings,  now is the time.

kmala:
started making changes, will be changing controller sdk, workflow-cli and controller..that should be it right?

bacongobbler:
yep! That should be the only places required.

rootfs/api/migrations/0012_auto_20160810_1603.py
```diff
@@ -9,7 +9,7 @@
 class Migration(migrations.Migration):
```
helgi: Since you deleted the other one, may want to rename this to 0011

kmala: changed

rootfs/api/models/app.py
helgi: Agree but we can refactor later

rootfs/api/models/app.py
```diff
@@ -424,6 +430,7 @@ def _scale_pods(self, scale_types):
                 'app_type': scale_type,
                 'build_type': release.build.type,
                 'healthcheck': healthcheck,
+                'annotations': {'maintenance': app_settings.maintenance},
```
helgi: This is rather unclear since this is service only but all resource objects in k8s takes annotations

kmala: do you want it to be renamed to service_annotations?

helgi: Yeah think that'd be better given the scope of that `kwargs`

PR #968, opened 2016-08-10
ref(api): simplify routing summary logic

bacongobbler:


bacongobbler:
So the problem appears to be that `routable` still appears to be true from the client. When adding debug to compare the old config against the new config, I see:

```
Disabling routing for go... Error: Unknown Error (409): {"detail":"True True"}
```

So this won't fix the problem. However, it is still a nice simplification of the logic.

bacongobbler:
closing since this doesn't resolve anything.

PR #969, opened 2016-08-10
fix(Makefile): add deis namespace to make deploy

bacongobbler:


bacongobbler:
So the problem appears to be that `routable` still appears to be true from the client. When adding debug to compare the old config against the new config, I see:

```
Disabling routing for go... Error: Unknown Error (409): {"detail":"True True"}
```

PR #972, opened 2016-08-11
ref(Makefile,.travis.yml): run shellcheck in container

vdice:

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,13 +6,12 @@ branches:
     - master
 services:
   - postgresql
+  - docker
 sudo: required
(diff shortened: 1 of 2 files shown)


helgi:
Looks like this might add ~20 seconds on a run - Not sure if travis caches docker images?

vdice:
@helgi I don't believe travis caches docker images yet... There are some [workarounds](http://stackoverflow.com/questions/35305492/cache-docker-images-on-travis-ci) but sounds like it could be iffy in actually offering speed improvements...

PR #974, opened 2016-08-11
fix(scheduler): make app un-routable

### LINKED ISSUE #973: `deis routing:disable` doesn't amend the routable annotation
using https://github.com/deis/workflow-cli/pull/172, I'm noticing that the service is not being updated with the change:

```
><> deis routing:disable
Disabling routing for go... done

><> deis routing
Routing is disabled.
><> curl go.k8s.local
Powered by Deis
Release v3 on go-web-2930364115-uf94c
```

```
><> k --namespace=go get svc go -o yaml | grep routable
    router.deis.io/routable: "true"
```

bacongobbler:
If an app is not routable, then the scheduler should remove the "routable" annotation from the application's service.

This is blocked by a bug in the router: https://github.com/deis/router/issues/233

closes #973

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -267,6 +267,9 @@ class KubeHTTPClient(object):
             # Update service information
             if routable:
                 service['metadata']['labels']['router.deis.io/routable'] = 'true'
+            else:
+                # delete the annotation
+                service['metadata']['labels'].pop('router.deis.io/routable', None)
 
             # Set app type if there is not one available
             if 'type' not in service['spec']['selector']:


helgi:
Why is this using labels instead of annotations?

https://github.com/deis/controller/blob/master/rootfs/api/models/__init__.py#L66

annotations are the agreed upon way to set configuration, and not labels, between controller and router

Makes me wonder if this shouldn't live inside the config model (or AppSettings) where it manipulates things outside of the `scheduler` code, like Domains already do. THoughts?

/cc @krancour

helgi:
oh wow... ignore me... I didn't realise this was using labels! My bad! weird to have things in so many places doing very similar things...

PR #975, opened 2016-08-11
fix(scheduler): when user asks for limits beyond their allowance then error out faster

helgi:
This detect errors from the pods event stream by turning the image error handling code into something slightly more generic so we can expand it over time if need be

```
Events:
  FirstSeenLastSeenCountFromSubobjectPathTypeReasonMessage
  ------------------------------------------------------------
  4m1m2{default-scheduler }WarningFailedSchedulingpod (gaslit-joyrider-cmd-2657344266-tv41n) failed to fit in any node
fit failure on node (ip-172-20-0-109.us-west-2.compute.internal): Node didn't have enough resource: CPU, requested: 2000000, used: 520, capacity: 2000
```

I made the "failed deploy, going back to old release" contain more contextual information. This can be good but may also get to be busy... Could change it so we log the additional information but do not return it via the API

# Testing Instructions

1. Create a Deis Cluster
2. Register an app
4. `deis limits:set -c cmd=2000` (this should fail)

Look at the message that comes back - For now it will not look good since the CLI or SDK (or go?) are not dealing with newlines

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -127,7 +127,8 @@ class KubeHTTPClient(object):
 
                 raise KubeException(
                     'There was a problem while deploying {} of {}-{}. '
-                    'Going back to the previous release'.format(version, namespace, app_type)
+                    'Going back to the previous release. '
+                    "Additional information:\n{}".format(version, namespace, app_type, str(e))
                 ) from e
 
         # Make sure the application is routable and uses the correct port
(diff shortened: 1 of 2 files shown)


mboersma:
```bash
$ deis limits:set -c cmd=20000 -a inward-inventor
Applying limits... Error: Unknown Error (400): {"detail":"(app::deploy): There was a problem while deploying v3 of inward-inventor-cmd. Going back to the previous release. Additional information:\npod (inward-inventor-cmd-3948341660-3fasy) failed to fit in any node\nfit failure on node (gke-mboersma-default-pool-f041a048-xxs8): Insufficient CPU\nfit failure on node (gke-mboersma-default-pool-f041a048-crig): Insufficient CPU\nfit failure on node (gke-mboersma-default-pool-f041a048-mlza): Insufficient CPU\n"}
```
It's pretty verbose, but informative.

PR #985, opened 2016-08-16
ref(hooks): remove push hook model as the builder stopped using it

helgi:
ref https://github.com/deis/builder/pull/407


Joshua-Anderson:
Is it okay to break backwards compatibility with hooks? It might make sense just to make `/v2/hooks/push` always return a 200 and mark it as deprecated.

helgi:
The current approach only affects people that do not update the builder but do update the controller - That's only a developer workflow as it stands since we do not support that behaviour outside of dev.

That being said I could keep the route, but people could also just clean install dev :P

bacongobbler:
LGTM since this does not affect the end user at all. This is explicitly an internal API and is not supported outside of the builder/controller.

PR #988, opened 2016-08-17
fix(registry): tell a user they need PORT when using off-cluster native iaas registry

### LINKED ISSUE #986: Using Private Registry doesn't always force a $PORT to be set
https://github.com/deis/builder/issues/409

The current logic for enforcing `$PORT` is done in `registry:set` (on the Controller side in the COnfig model). If a user sets a private registry overall via helm then the Controller does not enforce the `$PORT` behaviour.

Desired behaviour is for both of them to work like `registry:set` when it comes to `$PORT` - This might mean raising inside the Release or App model on a deploy

helgi:
registry:set does the same kind of warning but since this is a whole platform setting then we need to warn on deploy as well

Fixes #986

Requires off-cluster native iaas cluster to test

```
deis create --no-remote
Creating Application... done, created august-sandwich
If you want to add a git remote for this app later, use `deis git:remote -a august-sandwich`

deis pull deis/example-go -a august-sandwich
Creating build... Error: Unknown Error (400): {"detail":"PORT needs to be set in the config when using a private registry"}

deis config:set PORT=80 -a august-sandwich
Creating config... done

=== august-sandwich Config
PORT      80

deis pull deis/example-go -a august-sandwich
```

### DIFF
--- a/rootfs/api/models/release.py
+++ b/rootfs/api/models/release.py
@@ -152,8 +152,10 @@ class Release(UuidAuditedModel):
             # application has registry auth - $PORT is required
             if (creds is not None) or (settings.REGISTRY_LOCATION != 'on-cluster'):
                 if envs.get('PORT', None) is None:
-                    self.app.log('Private registry detected but no $PORT defined. Defaulting to $PORT 5000', logging.WARNING)  # noqa
-                    return 5000
+                    raise DeisException(
+                        'PORT needs to be set in the config '
+                        'when using a private registry'
+                    )
 
                 # User provided PORT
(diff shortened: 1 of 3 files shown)


Joshua-Anderson:
Is there any way to add tests to this?

helgi:
Actually yes I will do that

helgi:
Tests added - I had to amend a few other tests as the way the registry setting override had been done didn't gel with the test suite. Moved it all to use `@override` decorator to contain it to the test function instead of a global override

PR #993, opened 2016-08-19
ref(apps): move AppSettings scheduler logic to App model for simplicity

helgi:


bacongobbler:
massive cleanup. Nice.

PR #996, opened 2016-08-20
fix(scheduler): check if RC exists during a deploy before trying to use Deployments

### LINKED ISSUE #995: Issue upgrading form RC to Deployments (2.3 -> 2.4)
Upgraded from 2.3 to 2.4, application was converted to deployments on start. Things got weird.

https://gist.github.com/slack/613a0e6086b8fe92398caff606d9e5dd

helgi:
This prevents RCs and Deployments fighting about pods when upgrading from v2.3 to v2.4 - An application is not migrated to Deployments until configuration / image / etc is udone

Fixes #995


slack:
Testing with image `quay.io/jhansen/controller:git-eead5a2`.

Gory details here: https://gist.github.com/slack/cd0a9ddc12c50e734b3520cb0b73907e

Upgrade left app as RC, next push converted to deployments.

PR #1010, opened 2016-08-24
feat(whitelist): Add support for IP whitelist

kmala:
ref https://github.com/deis/controller/issues/240


rootfs/api/models/app.py
```diff
@@ -430,7 +430,6 @@ def _scale_pods(self, scale_types):
                 'app_type': scale_type,
                 'build_type': release.build.type,
                 'healthcheck': healthcheck,
-                'service_annotations': {'maintenance': app_settings.maintenance},
```
helgi: Why the removal here?

rootfs/api/models/app.py
```diff
@@ -912,3 +912,18 @@ def _update_application_service(self, namespace, app_type, port, routable=False,
...
+        Add/ Delete addresses to application whitelist
+        """
+        service = self._fetch_service_config(self.id)
+        old_service = service.copy()  # in case anything fails for rollback
+
+        try:
+            addresses = ",".join(address for address in whitelist)
+            service['metadata']['annotations']['router.deis.io/whitelist'] = addresses
+            self._scheduler.update_service(self.id, self.id, data=service)
+        except KubeException as e:
```
helgi: tests never reach this - would be good to get some testing on that

rootfs/api/models/app.py
kmala: its unused...though i can cleanup a bit.

rootfs/api/serializers.py
```diff
@@ -474,3 +475,18 @@ class Meta:
         """Metadata options for a :class:`AppSettingsSerializer`."""
         model = models.AppSettings
         fields = '__all__'
+
+    def validate_whitelist(self, data):
```
helgi: Wonder if we should use http://www.django-rest-framework.org/api-guide/fields/#ipaddressfield like we do  here https://github.com/deis/controller/blob/master/rootfs/api/serializers.py#L441-L444

Would potentially save us from custom validations

rootfs/api/urls.py
```diff
@@ -60,6 +60,9 @@
     # application settings
     url(r"^apps/(?P<id>{})/settings/?".format(settings.APP_URL_REGEX),
         views.AppSettingsViewSet.as_view({'get': 'retrieve', 'post': 'create'})),
+    # application ip whitelist
+    url(r"^apps/(?P<id>{})/whitelist/?".format(settings.APP_URL_REGEX),
```
helgi: You did it as a new route for more flexibility around add / remove of each IP?

kmala: yes.

rootfs/api/serializers.py
kmala: i think this is just for ip address and fails for cidr notation

helgi: ah yeah hmm https://github.com/django/django/blob/12b4280444b58c94197255655e284e4103fe00a9/django/core/validators.py#L273-L295

Thoughts on https://pypi.python.org/pypi/django-netfields/0.7 ?

kmala: even they use the ipaddress module of the python which i am using to validate and i need to handle it the same way as each value in the array can be an ip or cidr so saving as string is the best way i can think of.

helgi: But it would also move us to use a different PG field from what I could gather.

Guess you want CIDR and IP to work with this?

kmala: yes.

helgi: Alright - I'm just trying to make sure we're not maintaining stuff we don't have to... If that package can't allow for a mixed use field than we have to do our own

rootfs/api/models/app.py
kmala: Added

PR #1011, opened 2016-08-25
ref(api): change image pull policy to "IfNotPresent"

bacongobbler:
When networking issues are present with quay.io or Dockerhub, deploys with `deis pull` images will
always fail because the image pull policy is set to "Always". Even if every node in the cluster has
the image, the deploy will still fail because dockerhub or quay.io is down.

### DIFF
--- a/rootfs/api/settings/production.py
+++ b/rootfs/api/settings/production.py
@@ -254,8 +254,8 @@ BUILDER_KEY = os.environ.get('DEIS_BUILDER_KEY', random_secret)
 
 # k8s image policies
 SLUGRUNNER_IMAGE = os.environ.get('SLUGRUNNER_IMAGE_NAME', 'quay.io/deisci/slugrunner:canary')  # noqa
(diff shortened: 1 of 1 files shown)


helgi:
Didn't we set this to Always and then have the production / release charts change it? Idea being that dev is always pulling

kmala:
we should be adding the env variable in the charts to set to `IfNotPresent` instead of here because the ci environment uses canary and not tags.
also why are SLUG_BUILDER_IMAGE_PULL_POLICY and docker builder present in controller, shouldn't it be slugrunner?

bacongobbler:
it's been there since the dawn of time. Nobody's changed it. https://github.com/deis/controller/blob/bff6d609a2cfaa8bd91044ef864c996a8d3b9843/rootfs/scheduler/__init__.py#L272-L275

helgi:
That 404's for me :|

As far as what you are trying to achieve - More stable e2e?

bacongobbler:
More stable platform in general. A user in zendesk has flaky DNS issues, but quay.io/deis/slugrunner:v2.2.0 was present on all machines so changing the image pull policy to IfNotPresent by default would allow the deploy to succeed.

PR #1014, opened 2016-08-26
ref(Makefile): use make test-unit-quick by default

bacongobbler:
This target runs all the unit tests but in parallel, speeding up test time on Travis.


helgi:
`test-unit-quick` does not produce a code cov file. There are some issues around django + coverage and the parallel notion. I never cracked it.

bacongobbler:
Ah if that's the case then that sucks :( closing for now then as intentional

helgi:
Feel free to try to figure out the combination of commands we need to make it all work! :D

PR #1016, opened 2016-08-26
ref(scheduler): split up the scheduler code into individual resources and to be modular

### LINKED ISSUE #875: _api in the scheduler needs to handle non-v1 URLs better
Right now in the code for Deployments and ReplicaSets we special case them based on path information but that is not really a solid solution since they may graduate from `extensions/v1beta1` to stable.

A better solution needs to be found for this, maybe feed off the template version (which also would suffer from needing to be updated but at last that's a central source for a give Resource)

helgi:
This adds a resources package which contains all Kubernetes resources as their own individual classes.
Each resource is self registering via the ResourceRegistry metaclass, this allows the base Resource class to know what resources are available and expose that up into KubeHTTPClient

Each resource can specify their own api prefix and version, which allows us to support resources moving between API endpoints (see HPA, a new resource).
To make it easy for resource to interact with one another each resource and the KubeHTTPClient can access other resources via `self.hpa.create()` or `self.pod.get()`, the short form (if a resource supports that) works, singular and plural also work via some intelligent mapping

Additionally `get()` and `gets()` have been merged together into one `get()` per Resource

`version()` was added to let Resources find out what Kubernetes version they are dealing with.

Closes #875

### DIFF
--- a/rootfs/scheduler/__init__.py
+++ b/rootfs/scheduler/__init__.py
@@ -1,52 +1,18 @@
 from collections import OrderedDict
-from datetime import datetime, timedelta
-import json
+from datetime import datetime
 import logging
-import operator
-import os
-import time
-from urllib.parse import urljoin
-import base64
-
-from django.conf import settings
-from docker.auth import auth as docker_auth
-from .states import PodState
 import requests
 from requests_toolbelt import user_agent
-from .utils import dict_merge
+import time
 
 from api import __version__ as deis_version
+from scheduler.exceptions import KubeException, KubeHTTPException   # noqa
+from scheduler.states import PodState
 
 
 logger = logging.getLogger(__name__)
-
-
-class KubeException(Exception):
-    def __init__(self, *args, **kwargs):
-        Exception.__init__(self, *args, **kwargs)
-
-
-class KubeHTTPException(KubeException):
-    def __init__(self, response, errmsg, *args, **kwargs):
-        self.response = response
-
-        msg = errmsg.format(*args)
-        msg = "failed to {}: {} {}".format(
-            msg,
-            response.status_code,
-            response.reason
-        )
-        KubeException.__init__(self, msg, *args, **kwargs)
-
-
-def unhealthy(status_code):
-    if not 200 <= status_code <= 299:
-        return True
-
-    return False
-
-
 session = None
+resource_mapping = OrderedDict()
 
 
 def get_session():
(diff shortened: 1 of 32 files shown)


helgi:
There was a successful e2e run but CI didn't update GH - https://ci.deis.io/job/workflow-test-pr/5904/console

rootfs/scheduler/__init__.py
```diff
@@ -65,35 +31,127 @@ def get_session():
...
+        data = response.json()
+        return float('{}.{}'.format(data['major'], data['minor']))
+
+    @staticmethod
+    def parse_date(date):
+        return datetime.strptime(date, KubeHTTPClient.DATETIME_FORMAT)
+
+    @staticmethod
+    def unhealthy(status_code):
+        if not 200 <= status_code <= 299:
```
mboersma: This function could be simplified:
```python
def unhealthy(status_code):
    return 200 <= status_code <= 299
```

helgi: Sure could, but it was just a copy of what we had there originally but brought into the class

rootfs/scheduler/mock.py
```diff
@@ -112,6 +113,54 @@ def get_type(key, pos=-1):
...
+
+        # check if the resource referenced actually exists
+        kind = hpa['spec']['scaleRef']['kind'].lower() + 's'  # make plural
+        name = hpa['spec']['scaleRef']['name'].lower()
+        deployment = None
+        for deploy in cache.get(kind, []):
+            item = cache.get(deploy)
+            if item['metadata']['name'] == name:
+                deployment = item
+
```
mboersma: Should there be a `break` here? Could save some unnecessary looping at least.

rootfs/scheduler/resources/__init__.py
```diff
@@ -0,0 +1,12 @@
+from scheduler.resources.__resource import Resource  # noqa
+
+# Load in all resources
+import pkgutil
+import importlib
+import os
+pkgpath = os.path.dirname(__file__)
+for _, name, _ in pkgutil.iter_modules([pkgpath]):
```
mboersma: I think this would read a little better without the `continue` and blank line:
```python
for _, name, _ in pkgutil.iter_modules([pkgpath]):
    if not name.startwswith('__'):
        importlib.import_module('.{}'.format(name), 'scheduler.resources')
```

rootfs/scheduler/resources/horizontalpodautoscaler.py
```diff
@@ -0,0 +1,154 @@
...
+        # fetch HPA details
+        hpa = self.hpa.get(namespace, name).json()
+
+        # FIXME all of the below can be replaced with hpa['status'][desiredReplicas']
+        # when https://github.com/kubernetes/kubernetes/issues/29739 is fixed
+        # until then we have to query things ourselves
+
+        # only wait 30 seconds / attempts - this is not optimal
+        # ideally it would use the resources wait commands but they vary
+        for _ in range(30):
```
mboersma: `30` (and some other hard-coded numbers) keeps showing up here. Maybe it should be a module-level variable called `ATTEMPTS` or `RETRIES`?

helgi: Legacy - I'm not really inclined to make that configurable just yet since it would involve configuration work we don't want to expose due to the granularity. However we can do it another day of course

rootfs/scheduler/resources/pod.py
```diff
@@ -0,0 +1,709 @@
...
+
+    def find_container(self, container_name, containers):
+        """
+        Locate a container by name in a list of containers
+        """
+        for container in containers:
+            if container['name'] == container_name:
+                return container
+
+        return None
```
mboersma: This line can be omitted.

rootfs/scheduler/resources/__init__.py
helgi: addressed

mboersma:
My comments are all minor and can be ignored. Nice refactor! :lgtm:

helgi:
e2e https://ci.deis.io/job/workflow-test-pr/5921/console

PR #1017, opened 2016-08-26
fix(models): add "_added" log function, lowercase class name

bacongobbler:
This change caused e2e to fail because it expects the class name to be lower case.


helgi:
e2e https://ci.deis.io/job/workflow-test-pr/5919/console

PR #1018, opened 2016-08-26
feat(autoscale): add autoscaling support to application on per proc type basis

helgi:
* [ ] docs
* [ ] e2e
* [x] cli
* [x] sdk


rootfs/api/models/appsettings.py
```diff
@@ -86,6 +90,47 @@ def update_whitelist(self, previous_settings):
...
 
+    def update_autoscale(self, previous_settings):
+        data = getattr(previous_settings, 'autoscale', {}).copy()
+        new = getattr(self, 'autoscale', {}).copy()
+        # If no previous settings then do nothing
+        if not previous_settings:
+            return
+
+        # if nothing changed copy the settings from previous
+        if new is None and data is not None:
```
kmala: since the default is `{}` i don't think this check passes.

helgi: True - pushing change

rootfs/api/models/appsettings.py
```diff
@@ -86,6 +90,46 @@ def update_whitelist(self, previous_settings):
...
+
+            # only apply new items
+            for proc, scale in new.items():
+                self.app.autoscale(proc, scale)
+
+            # if the autoscale information changed, log the dict diff
+            changes = []
+            old_autoscale = getattr(previous_settings, 'autoscale', {})
+            diff = dict_diff(self.autoscale, old_autoscale)
+            # try to be as succinct as possible
```
arschles: the following lines take a while to grok. any chance they can be split up so they're more understandable?

arschles:
g2g, except for https://github.com/deis/controller/pull/1018#discussion_r76661043. I wouldn't consider that blocking for this PR though

rootfs/api/models/appsettings.py
helgi: That's a copy pasta of our Config / Release summary (and other things) :| I had talked to @bacongobbler about doing a more unified way of doing changelog summaries (if possible) across multiple models. Will refactor the below during that.

rootfs/api/models/appsettings.py
```diff
@@ -86,6 +90,46 @@ def update_whitelist(self, previous_settings):
...
+            setattr(self, 'autoscale', data)
+        elif data != new:
+            for proc, scale in new.items():
+                if scale is None:
+                    # error if unsetting non-existing key
+                    if proc not in data:
+                        raise UnprocessableEntity('{} does not exist under {}'.format(proc, 'autoscale'))  # noqa
+                    del data[proc]
+                else:
+                    data[proc] = scale
```
kmala: this depends a lot on how the cli sends the data....can user edit the hpa configuration and if so what does the cli ux looks like?

helgi: `deis autoscale:set cmd --min=3 --max=5 --cpu-percent=60 -a test-app`

helgi: So they can update / edit an existing rule by applying a full rule again with the same proc type

kmala: i am interested about edit....if he wants to edit max to 10 after he set it using the above command, what should he do?

helgi: Runs the exact same command as above but changes the max

kmala: so user has to do a set of set of everything even if he has to edit on parameter. wouldn't it be nice if user can be able to edit just the one which he want's to change rather copy everything again?

helgi: It would be but that also immensely complicates things since none of those values are optional on first set

kmala: yes...but i think handling them is better asking user to enter the whole thing again...right now there are only 3 parameters but in future it might grow into more right?
can we override create method https://docs.djangoproject.com/en/1.10/ref/models/instances/#django.db.models.Model to copy previous into current if its get called before the serializers but am not sure if it works.

helgi: Inclined to say we solve that problem when we get to it unless we have a clear cut way towards that right now. @slack ?

helgi: @kmala how do healthchecks behave in this same situation? They are far more complicated tbh but still

kmala: it has only one mandatory argument(port/command) and remaining are optional...even there to edit you have to enter the mandatory argument again

helgi: going by the `kubectl autoscale` help (I have to review HPA again to double check if it is identical) then `max` is the only required element, as such, but if we made that even optional to allow editing only min or only cpu % then we'd have to do something smart on the Controller side and have the CLI understand that as well

slack:
I don't have a strong opinion. The "override/set single value" UX is nice as long as the implementation doesn't become unruly.

Seems that there are two options, driven mostly by "what happens when you **first** create this thing without some values":

What does k8s do if the values are not set? I assume it takes some default value? If we could look up those defaults and store them off on the app/proc-type after creating the HPA, that would probably be preferable to a masked set of values. That could confuse the operator.

The other option would be for Workflow to hold HPA defaults for unspecified values. Benefit there is that you would get the same thing every time, no matter the k8s version. I think there is a strong argument there as well.

slack:
And to be fair. I'm okay with the "give me all of the values every time" even though it is annoying... I think there is going to be a bit of time before we declare autoscaling stable...

rootfs/api/models/app.py
```diff
@@ -925,3 +925,26 @@ def whitelist(self, whitelist):
...
+        try:
+            # get the target for autoscaler, in this case Deployment
+            self._scheduler.hpa.get(self.id, name)
+            if autoscale is None:
+                self._scheduler.hpa.delete(self.id, name)
+            else:
+                self._scheduler.hpa.update(
+                    self.id, name, proc_type, target, **autoscale
+                )
+        except KubeHTTPException as e:
```
kmala: shouldn't we re-throw/raise the exception if its not 404.

helgi: Yeah and looking at it I should split this into 2 separate things

helgi:
https://ci.deis.io/job/workflow-test-pr/6082/console finished fine even if it didn't get reported back here

helgi:
e2e went green https://ci.deis.io/job/workflow-test-pr/6212/console

PR #1019, opened 2016-08-27
ref(scheduler): add HTTP functions in KubeHTTPClient

### LINKED ISSUE #1008: gracefully handle cases when the k8s api server refuses connections
As seen in internal support calls with GKE, it is entirely possible for the controller/database pod to be operating normally while the k8s API server refuses to accept incoming connections. We need to handle cases during this time so that we can recover gracefully from these situations without breaking application deploys.

bacongobbler:
These functions properly handle connection errors and clean up some of the child/parent class
relations when making a request.

closes #1008

### DIFF
--- a/rootfs/scheduler/tests/test_kubehttpclient.py
+++ b/rootfs/scheduler/tests/test_kubehttpclient.py
@@ -1,0 +1,106 @@
+"""
+Unit tests for the Deis scheduler module.
+
+Run the tests with "./manage.py test scheduler"
+"""
+import json
+import requests
+import requests_mock
+from unittest import mock
+
+from django.conf import settings
+from django.test import TestCase
+
+import scheduler
+from scheduler import exceptions
+
+
+def mock_session():
+    return requests.Session()
+
+
+def connection_refused_matcher(request):
+    raise requests.ConnectionError("connection refused")
+
+
+@mock.patch('scheduler.get_session', mock_session)
+class KubeHTTPClientTest(TestCase):
+    """Tests kubernetes HTTP client calls"""
+
+    def setUp(self):
+        self.adapter = requests_mock.Adapter()
+        self.path = '/foo'
+        self.url = settings.SCHEDULER_URL + self.path
+        # use the real scheduler client.
+        self.scheduler = scheduler.KubeHTTPClient(settings.SCHEDULER_URL)
+        self.scheduler.session.mount(self.url, self.adapter)
+
+    def test_head(self):
+        """
+        Test that calling .http_head() uses the client session to make a HEAD request.
+        """
+        self.adapter.register_uri('HEAD', self.url)
+        response = self.scheduler.http_head(self.path)
+        assert response is not None
+        self.assertTrue(self.adapter.called)
+        self.assertEqual(self.adapter.call_count, 1)
+        # ensure that connection errors get raised as a KubeException
+        self.adapter.add_matcher(connection_refused_matcher)
+        with self.assertRaises(exceptions.KubeException):
+            self.scheduler.http_head(self.path)
+
+    def test_get(self):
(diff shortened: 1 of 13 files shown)


helgi:
Awesome, wonder if we should handle the `ConnectTimeout` and `ReadTimeout` exceptions `requests` has? There is also a `SSLError` one but that's perhaps a step too far.

Btw this can go into 2.5 since it was delayed.

rootfs/scheduler/tests/test_kubehttpclient.py
```diff
@@ -0,0 +1,106 @@
...
+        """
+        self.adapter.register_uri('GET', self.url)
+        response = self.scheduler.http_get(self.path)
+        assert response is not None
+        self.assertTrue(self.adapter.called)
+        self.assertEqual(self.adapter.call_count, 1)
+        # ensure that connection errors get raised as a KubeException
+        self.adapter.add_matcher(connection_refused_matcher)
+        with self.assertRaises(exceptions.KubeException):
+            self.scheduler.http_head(self.path)
```
helgi: Should this be `http_head` when it is in `test_get`?

rootfs/scheduler/tests/test_kubehttpclient.py
```diff
@@ -0,0 +1,106 @@
...
+        """
+        self.adapter.register_uri('POST', self.url)
+        response = self.scheduler.http_post(self.path, data=json.dumps({'hello': 'world'}))
+        assert response is not None
+        self.assertTrue(self.adapter.called)
+        self.assertEqual(self.adapter.call_count, 1)
+        # ensure that connection errors get raised as a KubeException
+        self.adapter.add_matcher(connection_refused_matcher)
+        with self.assertRaises(exceptions.KubeException):
+            self.scheduler.http_head(self.path)
```
helgi: Should this be `http_head` when it is in `test_post`?

rootfs/scheduler/tests/test_kubehttpclient.py
```diff
@@ -0,0 +1,106 @@
...
+        """
+        self.adapter.register_uri('PUT', self.url)
+        response = self.scheduler.http_put(self.path, data=json.dumps({'hello': 'world'}))
+        assert response is not None
+        self.assertTrue(self.adapter.called)
+        self.assertEqual(self.adapter.call_count, 1)
+        # ensure that connection errors get raised as a KubeException
+        self.adapter.add_matcher(connection_refused_matcher)
+        with self.assertRaises(exceptions.KubeException):
+            self.scheduler.http_head(self.path)
```
helgi: Should this be `http_head` when it is in `test_put`?

rootfs/scheduler/tests/test_kubehttpclient.py
```diff
@@ -0,0 +1,106 @@
...
+        """
+        self.adapter.register_uri('DELETE', self.url)
+        response = self.scheduler.http_delete(self.path)
+        assert response is not None
+        self.assertTrue(self.adapter.called)
+        self.assertEqual(self.adapter.call_count, 1)
+        # ensure that connection errors get raised as a KubeException
+        self.adapter.add_matcher(connection_refused_matcher)
+        with self.assertRaises(exceptions.KubeException):
+            self.scheduler.http_head(self.path)
```
helgi: Should this be `http_head` when it is in `test_delete`?

rootfs/scheduler/tests/test_kubehttpclient.py
bacongobbler: good catch. will fix

bacongobbler: all fixed.

PR #1025, opened 2016-08-30
fix(imagepullpolicy): Use correct environment variable for image pull policy

kmala:


rootfs/api/settings/production.py
```diff
@@ -254,8 +254,8 @@
 
 # k8s image policies
 SLUGRUNNER_IMAGE = os.environ.get('SLUGRUNNER_IMAGE_NAME', 'quay.io/deisci/slugrunner:canary')  # noqa
-SLUG_BUILDER_IMAGE_PULL_POLICY = os.environ.get('SLUG_BUILDER_IMAGE_PULL_POLICY', "Always")  # noqa
-DOCKER_BUILDER_IMAGE_PULL_POLICY = os.environ.get('DOCKER_BUILDER_IMAGE_PULL_POLICY', "Always")  # noqa
+SLUG_RUNNER_IMAGE_PULL_POLICY = os.environ.get('SLUG_RUNNER_IMAGE_PULL_POLICY', "IfNotPresent")  # noqa
+IMAGE_PULL_POLICY = os.environ.get('IMAGE_PULL_POLICY', "Always")  # noqa
```
helgi: Should we change this to `IfNoPresent` by default?

kmala: this is the image pull policy of the user pods which uses docker file or deis pull. so i am not sure if should strict policy.

helgi: Always pulling can mean slower deploys and could be prone to DNS issues - it does have some security around it (has more to do with some admission controllers and other internal k8s things)

PR #1026, opened 2016-08-30
feat(scheduler): add the ability to set KUBERNETES_POD_TERMINATION_GRACE_PERIOD_SECONDS per application

### LINKED ISSUE #807: Make KUBERNETES_POD_TERMINATION_GRACE_PERIOD_SECONDS configurable per app
We already do this for `DEIS_DEPLOY_BATCHES`, where you can do globally and then per app

Might require a change in name but probably not

helgi:
This makes a few internal methods now fetch the timeout from the pod instead of depending on outside info, previously it was fetching the info from global django.settings object.

Makes `pod_termination_grace_period_seconds` a required param for a few scheduler methods, like we already have with `app_type` etc etc

Close #807

### DIFF
--- a/rootfs/scheduler/resources/pod.py
+++ b/rootfs/scheduler/resources/pod.py
@@ -110,8 +110,8 @@ class Pod(Resource):
         # apply tags as needed to restrict pod to particular node(s)
         spec['nodeSelector'] = kwargs.get('tags', {})
 
-        # How long until a pod is forcefully terminated
-        spec['terminationGracePeriodSeconds'] = settings.KUBERNETES_POD_TERMINATION_GRACE_PERIOD_SECONDS  # noqa
+        # How long until a pod is forcefully terminated. 30 is kubernetes default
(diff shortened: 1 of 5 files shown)


rootfs/scheduler/resources/pod.py
```diff
@@ -384,14 +384,20 @@ def _set_image_secret(self, data, namespace, **kwargs):
         data['imagePullSecrets'] = [{'name': secret_name}]
 
     def delete(self, namespace, name):
+        # get timeout info from pod
+        pod = self.pod.get(namespace, name).json()
```
kmala: is there a chance of this giving 404 or do we want to handle that case in the function where it is called?

helgi: I thought about it and at the time I wanted the function it bubbles up to to handle it since `delete()` would cause a `404` if you call that on a non-existent resource

PR #1027, opened 2016-08-30
fix(app): rollback all process types to previous version when one (or  more) process type fails a deploy

### LINKED ISSUE #1013: Automatic rollback with k8s deploys doesn't roll back other process types
When a automatic rollback happens with k8s deployments on, it will not roll back other process types suck as worker.

Related to #1012

helgi:
Previously only the failed process type would roll itself back, resulting in a scenario where worker may be on v6 and web on v5

This moves away from using the built in rollback functionality in Deployments and rather deploys the previous release again to get the same effect.
Rollback in Deployments was taking it to the last good known deployment, in this case that'd be the latest for some types... knowing the revision of the last deploy would be hard as well.

The way it works when deploying the old release is an identical replicaset (and thus identical template pod hash is generated) so things will very much so look like a native rollback.
If the Controller has done any changes to how it constructs the pod manifest then this could generate a totally new ReplicaSet but that is also fine as it will be booting the previous release from DB

Fixes #1013

### Test Plan

1. `deis create --no-remote test-rollback`
2. `deis pull deis/example-go -a test-rollback`
3. `deis config:set foo=bar -a test-rollback`
4. `deis pull deis/example-foo -a test-rollback`

There should be an error message along the lines of `(app::deploy): There was a problem deploying v6. Rolling back process types to release v5.\nThere was a problem while deploying v6 of test-rollback-cmd. Additional information:\nError: image deis/example-foo:latest not found`

Running `deis logs -a test-rollback` should show some of that, including `There was a problem deploying v6. Rolling back process types to release v5` twice. That's okay and is due to https://github.com/deis/logger/issues/109

The first one is when the internal rollback starts and the second one is when that message is combined with the information returned back to the user

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -470,7 +470,7 @@ class App(UuidAuditedModel):
             self.log(err, logging.ERROR)
             raise ServiceUnavailable(err) from e
 
-    def deploy(self, release, force_deploy=False):
+    def deploy(self, release, force_deploy=False, rollback_on_failure=True):  # noqa
         """
         Deploy a new release to this application
 
@@ -569,8 +569,22 @@ class App(UuidAuditedModel):
                 ) for scale_type, kwargs in deploys.items()
             ]
 
-            async_run(tasks)
+            try:
+                async_run(tasks)
+            except KubeException as e:
+                if rollback_on_failure:
+                    err = 'There was a problem deploying {}. Rolling back process types to release {}.'.format(version, "v{}".format(release.previous().version))  # noqa
+                    # This goes in the log before the rollback starts
+                    self.log(err, logging.ERROR)
+                    # revert all process types to old release
+                    self.deploy(release.previous(), force_deploy=True, rollback_on_failure=False)
+                    # let it bubble up
+                    raise DeisException('{}\n{}'.format(err, str(e))) from e
+
+                # otherwise just re-raise
+                raise
         except Exception as e:
+            # This gets shown to the end user
             err = '(app::deploy): {}'.format(e)
             self.log(err, logging.ERROR)
             raise ServiceUnavailable(err) from e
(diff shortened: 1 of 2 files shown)


rootfs/api/models/app.py
```diff
@@ -555,8 +555,18 @@ def deploy(self, release, force_deploy=False):
...
 
-            async_run(tasks)
+            try:
+                async_run(tasks)
+            except KubeException as e:
+                err = 'There was a problem deploying {}. Rolling back process types to release {}.'.format(version, "v{}".format(release.previous().version))  # noqa
+                # This goes in the log before the rollback starts
+                self.log(err, logging.ERROR)
+                # revert all process types to old release
+                self.deploy(release.previous(), force_deploy=True)
```
kmala: if there is an api exception/unavailability doesn't this call deploy in recursion?

helgi: It will recurse all the way until it finds a good deploy, and the last one would be `no build` which we bail out on at the top, or you hit the 100 recursion limit (or so) in python.

bacongobbler: yep. this code could technically roll back all the way back when the controller is refusing connections like with https://github.com/deis/controller/pull/1019 and raise a NotExists error when it's at v1 and `previous()` raises that NotExists.

helgi: I think it is worth letting it go as far as possible back when it is doing a rollback but another thought would be to have a function arg that says "stop on failure", so we'd let it rollback _one_ level and that's it

kmala: we should just rollback to the previous release as that is what user expects and that has to be successful release otherwise we are doing deploy wrong.

helgi: Then I will implement my idea as that only goes to the previous release by only setting `stop_on_failure=True` when we hit the rollback

bacongobbler:
Haven't tested this, but code LGTM.

Issue #485, opened 2016-03-09
Port error after #473


aledbf:
Before #473 [this](https://github.com/deis/workflow/blob/master/rootfs/scheduler/__init__.py#L236) was `$port`. New apps set 8080 when the value for `PORT` is 5000

helgi:
@aledbf did you run into a bug? Can you give me steps to reproduce. 

The fact it is hardcoded as `8080` there shouldn't matter since we pull the port from the docker image later on in the flow. see https://github.com/deis/workflow/blob/master/rootfs/scheduler/__init__.py#L334-L341

helgi:
Haven't heard anything more about this. Things seem to work as intended despite what is set as default

@aledbf Please reopen if you think it is still a problem

PR #1035, opened 2016-09-02
fix(logs): app logs endpoint was returning binary string instead of a normal string

### LINKED ISSUE #1033: Logs API returns weird string
When GETting the logs, I'm seeing a long string with an unusual encoding

```
b'some long string with escaped new lines (\n)'
```

Would it be possible instead to return text/plain logs? I'm not sure how to decode the current format in JavaScript for https://github.com/olalonde/deisdash

helgi:
When python-requests got data from logger it was in binary string format, converting that to normal string was sufficient

Docs change https://github.com/deis/workflow/pull/485

Fixes #1033

### DIFF
--- a/rootfs/api/__init__.py
+++ b/rootfs/api/__init__.py
@@ -2,4 +2,4 @@
 The **api** Django app presents a RESTful web API for interacting with the **deis** system.
 """
 
-__version__ = '2.2.0'
+__version__ = '2.3.0'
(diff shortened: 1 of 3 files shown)


helgi:
Bumping the API Version as the SDK (and CLI, since beta) have been doing something like this https://github.com/deis/controller-sdk-go/blob/master/apps/apps.go#L110 

```
	// We need to trim a few characters off the front and end of the string
	return string(body[2 : len(body)-1]), reqErr
```

Older clients are not going to be very compatible with the new [correct] output

PR #1045, opened 2016-09-08
fix(whitelist): Handle empty whitelist from user gracefully

### LINKED ISSUE #1044: cannot remove whitelist
While testing v2.5.0:

```
><> deis whitelist:add 0.0.0.0/0
Adding 0.0.0.0/0 to go whitelist...
done
><> deis whitelist:list
=== go Whitelisted Addresses
0.0.0.0/0
><> curl g0.k8s.local
curl: (6) Could not resolve host: g0.k8s.local
><> curl go.k8s.local
Powered by Deis
Release v2 on go-web-2414165083-hyx25
><> deis whitelist
=== go Whitelisted Addresses
0.0.0.0/0
><> deis whitelist:remove 0.0.0.0/0
Removing 0.0.0.0/0 from go whitelist...
Error: Unknown Error (409): {"detail":"bacongobbler changed nothing"}
```

Looks like the release ledger summary needs to be updated in the controller.

kmala:
fixes #1044


rootfs/api/models/app.py
```diff
@@ -970,8 +970,13 @@ def whitelist(self, whitelist):
         service = self._fetch_service_config(self.id)
 
         try:
-            addresses = ",".join(address for address in whitelist)
-            service['metadata']['annotations']['router.deis.io/whitelist'] = addresses
+            if whitelist:
+                addresses = ",".join(address for address in whitelist)
+                service['metadata']['annotations']['router.deis.io/whitelist'] = addresses
+            elif 'router.deis.io/whitelist' in service['metadata']['annotations']:
+                service['metadata']['annotations'].pop('router.deis.io/whitelist', None)
```
bacongobbler: don't we want to pop, then update it?

kmala: pop should fix https://github.com/deis/router/issues/246 as i don't want to have the value empty.

rootfs/api/models/appsettings.py
```diff
@@ -19,7 +19,9 @@ class AppSettings(UuidAuditedModel):
     app = models.ForeignKey('App', on_delete=models.CASCADE)
     maintenance = models.NullBooleanField(default=None)
     routable = models.NullBooleanField(default=None)
-    whitelist = ArrayField(models.CharField(max_length=50), default=[])
+    # the default values is None to differentiate from user sending an empty whitelist
+    # and user just updating other fields meaning the values needs to be copied from prev release
+    whitelist = ArrayField(models.CharField(max_length=50), default=None)
```
bacongobbler: Won't this break the client since it expects an array?

https://github.com/deis/workflow-cli/blob/master/cmd/whitelist.go#L24-L26
https://github.com/deis/controller-sdk-go/blob/master/api/appsettings.go#L25

kmala: this is just the default value, the value saved in database will be empty array....i am doing it update_whitelist function which is called before saving to database.

bacongobbler: Ah I see what this does now. Same problem as with routable and maintenance mode... Need three values to differentiate between values unchanged (`None`), set and delete.

PR #1048, opened 2016-09-08
fix(tls): Update the value properly to work with morph

kmala:
load and fetch service methods use morph package(https://github.com/metagriffin/morph) which flatten the dictionary.
we are adding ssl.enforce as another key and using morph to flatten the dict https://github.com/deis/controller/blob/master/rootfs/api/models/__init__.py#L92
```
morph.flatten({'domains': 'earthy-quilting', 'maintenance': 'false', 'ssl': {'enforce': 'False'}, 'ssl.enforce': 'True'})
{'ssl.enforce': 'False', 'domains': 'earthy-quilting', 'maintenance': 'false'}
```

### DIFF
--- a/rootfs/api/models/tls.py
+++ b/rootfs/api/models/tls.py
@@ -41,11 +41,13 @@ class TLS(UuidAuditedModel):
         config = self._load_service_config(app, 'router')
 
         # See if the ssl.enforce annotation is available
-        if 'ssl.enforce' not in config:
-            config['ssl.enforce'] = 'false'
+        if 'ssl' not in config:
+            config['ssl'] = {}
+        if 'enforce' not in config['ssl']:
+            config['ssl']['enforce'] = 'false'
 
         # convert from bool to string
-        config['ssl.enforce'] = str(https_enforced)
+        config['ssl']['enforce'] = str(https_enforced)
 
         self._save_service_config(app, 'router', config)
 


vdice:
tested manually; looks good!
```
$ deis create app
Creating Application... done, created app
Git remote deis successfully created for app app.
$ deis tls -a app
=== app TLS
HTTPS Enforced: not set
$ deis tls:enable -a app
Enabling https-only requests for app... done
$ k get svc --namespace=app app -o json | jq -r '.metadata.annotations'
{
  "router.deis.io/domains": "app",
  "router.deis.io/maintenance": "false",
  "router.deis.io/ssl.enforce": "True"
}
$ curl -sL -w "%{http_code}\\n" app.104.198.47.116.nip.io -o /dev/null
301
$ deis tls:disable -a app
Disabling https-only requests for app... done
$ k get svc --namespace=app app -o json | jq -r '.metadata.annotations'
{
  "router.deis.io/domains": "app",
  "router.deis.io/maintenance": "false",
  "router.deis.io/ssl.enforce": "False"
}
$ curl -sL -w "%{http_code}\\n" app.104.198.47.116.nip.io -o /dev/null
200
```

bacongobbler:
Curious why this would Work on my Machine :tm:. Regardless if this is the fix then sounds good!

PR #1049, opened 2016-09-09
feat(Makefile): containerize dev workflow

### LINKED ISSUE #18: Proposal: containerize the development environment
This is to lower the barrier to entry for hacking on Deis while also standardizing the development environment for all contributors.

Joshua-Anderson:
Fixes #18 

Right now you have to manually start and stop postgres, but hopefully that can be improved so we don't have to in the future.

~~Does anyone know how to always run the target `make stop-postgres` after `make test-unit`?~~


vdice:
Nice!  Very excited for this.  

>Does anyone know how to always run the target make stop-postgres after make test-unit?
@Joshua-Anderson first thought would be to have separate task for the actual unit test command(s), say, `make do-test-unit` and then the actual `test-unit` target would just be a 'meta' target and look something like:
```
test-unit: docker-build-dev start-postgres do-test-unit stop-postgres
```

helgi:
Why is the `Dockerfile` duplicated from `rootfs` ?

Joshua-Anderson:
@helgi The main difference in the two is that this one has dev-dependencies installed and doesn't expose any ports. Any ideas on how to combine the two?

Joshua-Anderson:
> first thought would be to have a separate target for the actual unit test command(s), say, make do-test-unit and then the actual test-unit target would just be a 'meta'

@vdice but if the unit tests fail, postgres wouldn't be stopped, would it?

helgi:
> @helgi The main difference in the two is that this one has dev-dependencies installed and doesn't expose any ports. Any ideas on how to combine the two?

No idea tbh - Having 2 `Dockerfiles` is going to be confusing as they look near identical. I'm not sure I'm seeing the benefit of containerizing the dev env here, not off the bat. It solves the "I'm running one of the make targets" problem but while developing on the Controller at least I commonly run `python rootfs/manage.py test etc etc` to have granularity on a particular test module or test method

Joshua-Anderson:
This solves the problem of needing to have the right python version installed, all the dependencies, and python configured to talk to a postgres database you need to install and run. This sort of stuff can be a big hurdle for community contributors and people like me who don't want to spend 30 minutes getting my computer properly configured. Now a developer can just run `make test-unit` and if they have docker installed they can run the unit tests.

You're right that it doesn't solve the problem of running a particular test target. However, you can copy-paste the docker command from the makefile and edit it, or run the tests locally, since most core developers like you already have the environment setup.

helgi:
Again, not a fan of this. However if this is to be done I want a clearer distinction between the `Dockerfiles`. Would be ideal if it was possible to combine together with the current `sha` image at least and then at in the dev tooling only. Not sure how you'd do that without magic? `sed` replace on the base image at the top? Not sure

bacongobbler:
I'm not a fan of two Dockerfiles either. We should repurpose the Dockerfile in rootfs.

If we are forced to go the two Dockerfile path, then perhaps a `Dockerfile.dev` file in `rootfs` that has a FROM clause, building on top of quay.io/deis/controller:canary would be ideal. That way we're not having to maintain two Dockerfiles that are exact copies of eachother.

vdice:
> but if the unit tests fail, postgres wouldn't be stopped, would it?

@Joshua-Anderson good point; although it would involve a bit of duplicated `deispg` container teardown logic, extending the suggestion above, `do-test-unit` might look like:
```
start-postgres:
  docker run ...
stop-postgres:
  docker stop deispg && docker rm deispg
do-test-unit: start-postgres
  docker run <unit test command> || (docker stop deispg && docker rm deispg)
test-unit: do-test-unit stop-postgres
```

Joshua-Anderson:
I updated the test-unit targets to automatically start/stop the database. I moved the main rootfs Dockerfile to be the root dockerfile and renamed the development dockerfile to `Dockerfile.test`, which @helgi suggested to avert confusion.

@helgi and I talked and we're thinking the longer term solution is to update docker-python-dev and use that as a base for the test image.

Joshua-Anderson:
I tested `test-unit-quick` with and without the container and the both were within 5 secs of each other in a 3:30 test run.

Makefile
```diff
@@ -59,18 +64,20 @@ test: test-style test-check test-unit test-functional
...
+test-style: docker-build-dev
+	docker run --rm ${DEV_IMAGE} flake8 --show-pep8 --show-source
 	${SHELLCHECK_PREFIX} $(SHELL_SCRIPTS)
 
-test-unit:
-	cd rootfs \
-		&& coverage run manage.py test --settings=api.settings.testing --noinput registry api scheduler.tests \
-		&& coverage report -m
+test-unit: start-postgres do-test-unit stop-postgres
+do-test-unit: docker-build-dev
```
vdice: I'd propose we remove the `docker-build-dev` dependency here (as, if it fails, the `deispg` stop logic won't be executed) and place it into the above `test-unit` target, to look like:
```
test-unit: docker-build-dev start-postgres do-test-unit stop-postgres
do-test-unit: 
...
```
(and same for the `-quick` variants below)

Makefile
```diff
@@ -42,13 +46,14 @@ commit-hook:
...
-	@echo "    export PGHOST=`docker-machine ip $$(docker-machine active) 2>/dev/null || echo 127.0.0.1`"
-	@echo "    export PGPORT=5432"
-	@echo "    export PGUSER=postgres"
+start-postgres:
+	docker run -d --name deispg postgres:9.4
+	# wait for the database to come up
+	sleep 3
+
+stop-postgres:
+	docker stop deispg
```
vdice: a minor DRY improvement: might consider defining a function in this Makefile executing these commands:
```
define stop-pg
  	docker stop deispg && docker rm deispg
endef
```
and then every target that needs to run this logic can then do so:
```
stop-postgres:
  $(call stop-pg)
...
do-unit-test:
  docker  run ... || $(call stop-pg)
```

Joshua-Anderson:
@vdice Fixed!

vdice:
I'm not a regular controller contributor or maintainer, but as regards the Makefile/Dockerfile code here, LGTM, on top of the huge ease-of-use gains both for devs/contributors and ci.  Tested manually via running `make test-unit`.

vdice:
One more thought @Joshua-Anderson -- should we perhaps update the `.travis.yml` accordingly to now run the containerized test target(s) and get its blessing?  (We currently do not run unit tests in the Jenkins build jobs, though with this effort we are very close if not ready to revisit)

Dockerfile
```diff
@@ -7,24 +7,31 @@ RUN adduser --system \
...
-        python3 \
-        sudo && \
-    ln -s /usr/bin/python3 /usr/bin/python && \
+				libpq5 \
+				python3 \
+				sudo && \
+		ln -s /usr/bin/python3 /usr/bin/python && \
 	curl -sSL https://bootstrap.pypa.io/get-pip.py | python - pip==8.1.2 && \
 	mkdir -p /configs && chown -R deis:deis /configs && \
+	apt-get clean && \
```
helgi: Do we want to cleanup 2x

Joshua-Anderson: Yes, because you have to cleanup in the same layer you add the files. Deleting files in a different would make the image larger.

helgi: fair enough

Joshua-Anderson:
I got this almost all working, but it's having codecov issues. If anyone is interested in going the last mile 😄 all deis maintainers have permission to push to this PR branch (thanks to a new GH feature), so you can update it.

helgi:
@Joshua-Anderson what's wrong with the codecov? I see it got posted in the travis build but also didn't see it post to GH.

Joshua-Anderson:
@helgi even though I specifically selected the file no test coverage reports were found.

helgi:
@Joshua-Anderson curious... the CLI system is having problems talking to codecov so that was my first instinct (that it was identical problems)

helgi:
Leaving this here http://docs.codecov.io/docs/testing-with-docker if anyone wants to take a crack at this

bacongobbler:
Shall we punt this to another release or is someone planning on picking this back up this week?

PR #1051, opened 2016-09-09
ref(Dockerfile): ditch the pip cache for a slightly smaller image

helgi:

### DIFF
--- a/rootfs/Dockerfile
+++ b/rootfs/Dockerfile
@@ -10,19 +10,20 @@ RUN adduser --system \
 COPY requirements.txt /app/requirements.txt
 
 RUN buildDeps='gcc git libffi-dev libpq-dev python3-dev'; \
-	apt-get update && \
-	apt-get install -y --no-install-recommends \
+    apt-get update && \
+    apt-get install -y --no-install-recommends \
         $buildDeps \
         libpq5 \
         python3 \
         sudo && \
     ln -s /usr/bin/python3 /usr/bin/python && \
-	curl -sSL https://bootstrap.pypa.io/get-pip.py | python - pip==8.1.2 && \
-	mkdir -p /configs && chown -R deis:deis /configs && \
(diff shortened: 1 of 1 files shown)


rootfs/Dockerfile
```diff
@@ -10,19 +10,20 @@ RUN adduser --system \
...
     ln -s /usr/bin/python3 /usr/bin/python && \
-	curl -sSL https://bootstrap.pypa.io/get-pip.py | python - pip==8.1.2 && \
-	mkdir -p /configs && chown -R deis:deis /configs && \
-	pip install --disable-pip-version-check --no-cache-dir -r /app/requirements.txt && \
-	apt-get purge -y --auto-remove $buildDeps && \
-	apt-get clean && \
-	rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/* /usr/share/man /usr/share/doc
+    curl -sSL https://bootstrap.pypa.io/get-pip.py | python - pip==8.1.2 && \
+    mkdir -p /configs && chown -R deis:deis /configs && \
+    pip install --disable-pip-version-check --no-cache-dir -r /app/requirements.txt && \
```
Joshua-Anderson: Doesn't `--no-cache-dir` already do this?

helgi: You'd think but if you drop onto the controller container and `dh -h` (or similar) that dir, it has data. Can you try it on yours just to confirm?

helgi:
Also considering removing pip itself

Joshua-Anderson:
Removing pip makes sense. Any reason not to?

helgi:
Mostly if we are extending main to dev would be a reason to keep it but that's about it!

Joshua-Anderson:
LGTM, but we should probably investigate why this is neccessary (maybe open an issue?). Could it be that `--no-cache` only partially disables caching?

PR #1053, opened 2016-09-12
ref(scheduler): move image pull policy settings from scheduler to App model and remove SLUG_RUNNER_IMAGE_PULL_POLICY

helgi:
Removes the last of django.settings from scheduler main line code (not tests)

IMAGE_PULL_POLICY is now the only thing used and can be set per app, SLUG_RUNNER_IMAGE_PULL_POLICY has been removed

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -417,6 +417,9 @@ class App(UuidAuditedModel):
         # get application level pod termination grace period
         pod_termination_grace_period_seconds = release.config.values.get('KUBERNETES_POD_TERMINATION_GRACE_PERIOD_SECONDS', settings.KUBERNETES_POD_TERMINATION_GRACE_PERIOD_SECONDS)  # noqa
 
+        # set the image pull policy that is associated with the application container
+        image_pull_policy = release.config.values.get('IMAGE_PULL_POLICY', settings.IMAGE_PULL_POLICY)  # noqa
+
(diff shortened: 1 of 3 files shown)


helgi:
I changed things around a bit so that image pull policy per app is configurable via `IMAGE_PULL_POLICY` using `config:set` but I did keep `SLUG_RUNNER_IMAGE_PULL_POLICY` for defaults.

Given sometimes people may want to do `IfNotPresent` for slug runner only but `Always` on `pull` then the separation may make sense but I'd also argue that people could configure it just as well with the per app settings and a singular `IMAGE_PULL_POLICY` globally.

@slack (and others interested): Thoughts on the above?

helgi:
@kmala thoughts on ditching the `SLUG` env var?

kmala:
if we remove SLUG_RUNNER_IMAGE_PULL_POLICY then user can't globally specify different policies for buildpack and docker apps right?

helgi:
Yeah, that'd be the case - They'd be able to do 1 type of image policy globally but then configure them also per app (which this PR adds)

kmala:
i am fine with removing it.

slack:
Collapsing `SLUG_RUNNER_IMAGE_PULL_POLICY` to `IMAGE_PULL_POLICY` makes sense.

The referenced issue was about the `BUILDER_` variants and those should stay independent.

PR #1056, opened 2016-09-12
chore(requirements) Update to docker-py 1.10.2

helgi:
https://github.com/docker/docker-py/blob/master/docs/change_log.md#1100

@kmala do you want to add anything into this PR that uses https://github.com/docker/docker-py/pull/1133/files ?

### DIFF
--- a/rootfs/requirements.txt
+++ b/rootfs/requirements.txt
@@ -4,7 +4,7 @@ Django==1.10.1
 django-cors-middleware==1.3.1
 django-guardian==1.4.6
 djangorestframework==3.4.6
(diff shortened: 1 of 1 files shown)


kmala:
i would like to remove the `docker login` https://github.com/deis/controller/blob/master/rootfs/registry/dockerclient.py#L35 and use custom auth for both pull and push  so that we don't save the auth information on the nodes but that can be done as part of separate PR if needed.

helgi:
@kmala okay - I guess that means the other functions we use to support that - Hopefully with good enough error handling as well

PR #1060, opened 2016-09-13
fix(whitelist): remove the whitelist from annoations if its empty

kmala:
This removes the unnecessary warnings from router logs
```
WARNING: Field "router.deis.io/whitelist" value "" does not satisfy constraint /^((([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/([0-9]|[1-2][0-9]|3[0-2]))?(\s*,\s*)?)+$/ -- skipping this field and using default value "[]".
```

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -494,7 +494,10 @@ class App(UuidAuditedModel):
             raise DeisException('No build associated with this release')
 
         app_settings = self.appsettings_set.latest()
-        addresses = ",".join(address for address in app_settings.whitelist)
+        if app_settings.whitelist:
+            addresses = ",".join(address for address in app_settings.whitelist)
+        else:
+            addresses = None
         service_annotations = {'maintenance': app_settings.maintenance, 'whitelist': addresses}
 
         # use create to make sure minimum resources are created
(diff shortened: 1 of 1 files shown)


rootfs/api/models/app.py
```diff
@@ -494,7 +494,7 @@ def deploy(self, release, force_deploy=False, rollback_on_failure=True):  # noqa
             raise DeisException('No build associated with this release')
 
         app_settings = self.appsettings_set.latest()
-        addresses = ",".join(address for address in app_settings.whitelist)
+        addresses = ",".join(address for address in app_settings.whitelist) if app_settings.whitelist else None  # noqa
```
mboersma: I think it would be more Pythonic to spell this one out instead of make it a one-liner that needs `# noqa`:
```python
if app_settings.whitelist:
    addresses = ",".join(address for address in app_settings.whitelist)
else:
    addresses = None
```

kmala: Changed.

PR #1068, opened 2016-09-15
feat(api): add swagger support at the root of the API

### LINKED ISSUE #811: Specify the API using swagger
This issue is for investigating swagger server stub code generators for Python & Django Rest Framework, and then if there is a suitable one that won't require a large refactor of this codebase, doing the following:

1. Writing a swagger specification for the controller's full API
2. Generating server stubs and refactoring this codebase into the stubs
3. Generate the Go client bindings in [deis/builder](https://github.com/deis/builder) and [deis/workflow-cli](https://github.com/deis/workflow-cli) (each to be done in a separate issue in the respective repository)
  1. The changes in deis/workflow-cli may move to a new repo if https://github.com/deis/workflow-cli/issues/18 is completed first

helgi:
Content Type for it is `application/openapi+json` or `?format=openapi`

Closes #811

Very much so work in progress still

A few off the cuff considerations:
- Auth the endpoint? (use API key). Hooks get hidden from view when using normal tokens, which defeats the point
- throttle access to that endpoint
- save swagger on each release? add into the docs site somehow potentially

### DIFF
--- a/rootfs/api/settings/production.py
+++ b/rootfs/api/settings/production.py
@@ -101,6 +101,7 @@ INSTALLED_APPS = (
     'jsonfield',
     'rest_framework',
     'rest_framework.authtoken',
+    'rest_framework_swagger',
     # Deis apps
     'api'
 )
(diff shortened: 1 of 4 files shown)


helgi:
When https://github.com/tomchristie/django-rest-framework/pull/4525 (DRF 3.5) is out then I'll revisit most of this

mboersma:
We are also considering using https://github.com/kubernetes-incubator/client-python since it seems to be swagger-generated and may come close to matching our hand-crafted, artisanal k8s client functionality.

mboersma:
I'm closing this PR since it's old and rusty, but @helgi please re-open if this code gets resurrected. Thanks again!

PR #1076, opened 2016-09-16
fix(api): cast DEPLOY_BATCHES and DEPLOY_TIMEOUT to int

bacongobbler:
Otherwise it will be a string in the scheduler code and break it.

closes deis/builder#429

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -996,16 +996,16 @@ class App(UuidAuditedModel):
         config = release.config
 
         # see if the app config has deploy batch preference, otherwise use global
-        batches = config.values.get('DEIS_DEPLOY_BATCHES', settings.DEIS_DEPLOY_BATCHES)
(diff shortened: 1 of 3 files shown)


helgi:
https://github.com/deis/controller/blob/master/rootfs/scheduler/__init__.py#L374 is where some of that was being cast to int thus far - I like yours better

bacongobbler:
blocked by #1077

helgi:
@bacongobbler the `or 0` part worries me - no timeout then? not grace period?

bacongobbler:
I'm going to fix that up, actually. It's because production settings just fetch the envvar but don't supply a default value for these integers, therefore they become the NoneType. Going to shift it around once https://github.com/deis/controller/pull/1078 is merged.

helgi:
Sounds good - I'm sure there are defaults being set through the scheduler / app model that probably could just be shifted all the way up to the settings file

PR #1078, opened 2016-09-16
fix(api): check if release.build is NoneType

### LINKED ISSUE #1077: unit tests failing
The change introduced in https://github.com/deis/controller/pull/1061 has started to cause unit test failures:

```
...................E....................................................................................................................................................................................................................................................................

ERROR: test_gather_app_settings (api.tests.test_app.AppTest)

  File "/home/travis/virtualenv/python3.5.0/src/request-mock/requests_mock/mocker.py", line 180, in inner

    return func(*args, **kwargs)

  File "/opt/python/3.5.0/lib/python3.5/unittest/mock.py", line 1157, in patched

    return func(*args, **keywargs)

  File "/home/travis/build/deis/controller/rootfs/api/tests/test_app.py", line 589, in test_gather_app_settings

    3)

  File "/home/travis/build/deis/controller/rootfs/api/models/app.py", line 994, in _gather_app_settings

    envs = self._build_env_vars(release)

  File "/home/travis/build/deis/controller/rootfs/api/models/app.py", line 797, in _build_env_vars

    if release.build.type == 'buildpack':

AttributeError: 'NoneType' object has no attribute 'type'
```

bacongobbler:
closes #1077

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -23,9 +23,9 @@ from api import __version__ as deis_version
 from api.models import UuidAuditedModel, AlreadyExists, DeisException, ServiceUnavailable
 
 from api.utils import generate_app_name, async_run
-from api.models.release import Release
 from api.models.config import Config
 from api.models.domain import Domain
+from api.models.release import Release
 from api.models.tls import TLS
 from api.models.appsettings import AppSettings
 
@@ -786,6 +786,8 @@ class App(UuidAuditedModel):
         Build a dict of env vars, setting default vars based on app type
         and then combining with the user set ones
         """
+        if release.build is None:
+            raise DeisException('No build associated with this release to run this command')
 
         # mix in default environment information deis may require
         default_env = {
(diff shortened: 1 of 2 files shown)


helgi:
Just had a look, the reason why this fails is because the code assumes (incorrectly I suppose) that it's being called from within `run`/`scale`/`deploy` and all of those already do a `release.build` check and raise an exception. `_gather_app_settings` should not be even returning anything without `release.build` being set

https://github.com/deis/controller/pull/1076 made it so the release object does not have `Build` and is calling `_gather_app_settings` directly, whichi s why it got triggered.

I think you can overall change all the above checks to just this at the top of the function

```
        if release.build is None:
            raise DeisException('No build associated with this release to run this command')
```

bacongobbler:
This also consistently fails because I'm on my laptop and don't have python3 installed so I'm just abusing CI as my tester.

> _gather_app_settings should not be even returning anything without release.build being set

Sure, I'll make that change.

PR #1080, opened 2016-09-16
feat(api): inject release metadata into application

### LINKED ISSUE #1057: Proposal: Add Release Data to Pod Environment
It would be great if DEIS could add some environment variables with release information to the pods environment.

Currently only `WORKFLOW_RELEASE=v123` is set, but for things like release and error tracking the commit hash would be very useful.

Heroku Labs has a similar feature called [Dyno Metadata](https://devcenter.heroku.com/articles/dyno-metadata).

We could have a few additional `WORKFLOW_*` environment vars:

* `WORKFLOW_RELEASE=v88` (already exists)
* `WORKFLOW_RELEASE_CREATED_AT=2016-09-12T00:08:22Z`
* `WORKFLOW_RELEASE_DESCRIPTION="gitlab-runner deployed 5c2d30d"`
* `WORKFLOW_RELEASE_HASH=5c2d30d6719413ce7d5f5fc7a66d8c349dce162b`

The WORKFLOW_RELEASE_HASH could use the GIT commit SHA for Buildpack or Dockerfile builds and the Docker Image ID when pulling images.

The important thing about `WORKFLOW_RELEASE_HASH` is that it should be able to be determined outside of DEIS based on the pushed data, e.g. in a CI system one might have these stages:

* Push git ref to DEIS Workflow
* Create a release with same git ref in error monitoring tool (eg. Sentry)
* Push source code / source maps to Sentry

Currently to achieve the same one would have to either grab the release version number from the git push logs to pass it on to the other stages or query deis for the last release version, which is prone to race conditions.

If the ambiguos value for `WORKFLOW_RELEASE_HASH` is a problem, this variable could be omitted for Docker image deploys and instead replaced with `WORKFLOW_RELEASE_COMMIT` or `SOURCE_VERSION` (like during build) for Buildpack and Dockerfile deploys, because an equivalent value can easily be baked into a Docker image.

This is somewhat related to deis/builder#418 which introduced the `SOURCE_VERSION` environment variable at build time.

bacongobbler:
closes #1057 

rebased off of #1078 for the _build_env_vars tests, see e88c52f for the new commit.

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -792,7 +792,10 @@ class App(UuidAuditedModel):
         # mix in default environment information deis may require
         default_env = {
             'DEIS_APP': self.id,
-            'WORKFLOW_RELEASE': 'v{}'.format(release.version)
+            'WORKFLOW_RELEASE': 'v{}'.format(release.version),
+            'WORKFLOW_RELEASE_SUMMARY': release.summary,
+            'WORKFLOW_RELEASE_CREATED_AT': str(release.created.strftime(
+                settings.DEIS_DATETIME_FORMAT))
         }
 
         # Check if it is a slug builder image.
@@ -802,6 +805,9 @@ class App(UuidAuditedModel):
             default_env['BUILDER_STORAGE'] = settings.APP_STORAGE
             default_env['DEIS_MINIO_SERVICE_HOST'] = settings.MINIO_HOST
             default_env['DEIS_MINIO_SERVICE_PORT'] = settings.MINIO_PORT
+
+        if release.build.sha:
+            default_env['SOURCE_VERSION'] = release.build.sha
 
         # fetch application port and inject into ENV vars as needed
         port = release.get_port()
--- a/rootfs/api/tests/test_app.py
+++ b/rootfs/api/tests/test_app.py
@@ -582,16 +582,21 @@ class AppTest(DeisTestCase):
         url = "/v2/apps/{app.id}/builds".format(**locals())
         response = self.client.post(url, data)
         self.assertEqual(response.status_code, 201, response.data)
+        time_created = app.release_set.latest().created
         self.assertEqual(
             app._build_env_vars(app.release_set.latest()),
             {
                 'DEIS_APP': app.id,
                 'WORKFLOW_RELEASE': 'v2',
-                'PORT': 5000
+                'PORT': 5000,
+                'WORKFLOW_RELEASE_SUMMARY': 'autotest deployed autotest/example',
+                'WORKFLOW_RELEASE_CREATED_AT': str(time_created.strftime(
+                    settings.DEIS_DATETIME_FORMAT))
             })
         data['sha'] = 'abc1234'
         response = self.client.post(url, data)
         self.assertEqual(response.status_code, 201, response.data)
+        time_created = app.release_set.latest().created
         self.assertEqual(
             app._build_env_vars(app.release_set.latest()),
             {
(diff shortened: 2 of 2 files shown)


helgi:
Is this what the ticket that was created asked for? Thought there was a workflow prefix and all?

bacongobbler:
Given deis/builder#418 and Heroku both use SOURCE_VERSION it makes sense to use that convention. I can add the other environment variables here, one sec

helgi:
Ah is that the only one that was not prefixed? Sounds good

rootfs/api/models/app.py
```diff
@@ -800,11 +802,11 @@ def _build_env_vars(self, release):
             default_env['BUILDER_STORAGE'] = settings.APP_STORAGE
             default_env['DEIS_MINIO_SERVICE_HOST'] = settings.MINIO_HOST
             default_env['DEIS_MINIO_SERVICE_PORT'] = settings.MINIO_PORT
+            default_env['SOURCE_VERSION'] = release.build.sha
 
         # fetch application port and inject into ENV vars as needed
         port = release.get_port()
-        if port:
-            default_env['PORT'] = port
+        default_env['PORT'] = port
```
kmala: any specific reason for this change or just a clean up

bacongobbler: this is the rebased commit from #1078, but essentially this is cleanup from dead code. `get_port` used to return a NoneType if the port didn't exist; now it just raises a DeisException or a numerical value so we don't need this check any more

rootfs/api/models/app.py
```diff
@@ -800,11 +804,11 @@ def _build_env_vars(self, release):
             default_env['BUILDER_STORAGE'] = settings.APP_STORAGE
             default_env['DEIS_MINIO_SERVICE_HOST'] = settings.MINIO_HOST
             default_env['DEIS_MINIO_SERVICE_PORT'] = settings.MINIO_PORT
+            default_env['SOURCE_VERSION'] = release.build.sha
```
kmala: should we check for sha and add as environment as it is available even for dockerfile apps

bacongobbler: Sure, that's not a bad idea.

bacongobbler: FYI this is done now

ultimateboy:
Seems like we should add these new vars to the test here: https://github.com/deis/controller/blob/master/rootfs/api/tests/test_app.py#L585

bacongobbler:
They have been added! See the changes to test_app.py in this PR. :)

rootfs/api/models/app.py
```diff
@@ -792,7 +792,10 @@ def _build_env_vars(self, release):
         # mix in default environment information deis may require
         default_env = {
             'DEIS_APP': self.id,
-            'WORKFLOW_RELEASE': 'v{}'.format(release.version)
+            'WORKFLOW_RELEASE': 'v{}'.format(release.version),
+            'WORKFLOW_RELEASE_SUMMARY': release.summary,
+            'WORKFLOW_RELEASE_CREATED_AT': str(release.created.strftime(
+                settings.DEIS_DATETIME_FORMAT))
```
helgi: Does this need to go through `strftime`? Just curious since we set the format already for DRF so I'd expect it to translate it for us when we do `str()` on it, but maybe not?

helgi:
Did we consider the ambiguousness that was brought up in #1057 and what was the final decision? I see we only cover the `SOURCE_VERSION` story when `sha` is available

rootfs/api/models/app.py
bacongobbler: I think it does because we do it elsewhere in the codebase. DRF just handles it in the serializer, I think.

bacongobbler:
> Did we consider the ambiguousness that was brought up in #1057 and what was the final decision? I see we only cover the SOURCE_VERSION story when sha is available

With this PR we now cover WORKFLOW_RELEASE_SUMMARY, WORKFLOW_RELEASE_CREATED_AT and SOURCE_VERSION which is WORKFLOW_RELEASE_HASH. I didn't bother with WORKFLOW_RELEASE_HASH because SOURCE_VERSION already has that data.

felixbuenemann:
@bacongobbler Thanks for implementing this!

So for docker image deploys I would have to set SOURCE_VERSION when building the image, right?

bacongobbler:
Yep!

bacongobbler:
Wait no. If you're using Dockerfile or buildpack deploys the environment variable will be injected into your environment.

felixbuenemann:
@bacongobbler Docker image != Dockerfile =)

bacongobbler:
Okay then yes if you're using `deis pull` you'll have to inject that into the environment yourself because we don't have any git metadata to derive the sha from in `deis pull`. The only information we have is the image and the release summary.

PR #1088, opened 2016-09-25
ref(Dockerfile): use base 0.3.4, smaller packages and better cleanup

helgi:
Use `python3-pip` instead of using non-package management approach as this provides us with the proper pip version anyway, and better cleanup

Old vs New image size:

```
REPOSITORY                TAG                 IMAGE ID             SIZE
quay.io/deis/controller   v2.5.1              8e926c780d4d         214.4 MB
quay.io/helgi/controller   canary              a1017a79f8a2       204.8 MB
```

### DIFF
--- a/rootfs/Dockerfile
+++ b/rootfs/Dockerfile
@@ -1,4 +1,4 @@
-FROM quay.io/deis/base:0.3.1
+FROM quay.io/deis/base:v0.3.4
 
 RUN adduser --system \
 	--shell /bin/bash \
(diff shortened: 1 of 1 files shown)


rootfs/Dockerfile
```diff
@@ -9,21 +9,39 @@ RUN adduser --system \
...
+        /var/lib/apt/lists/* \
+        /var/log/* \
+        /var/cache/debconf/* \
+        /usr/share/common-licenses* \
+        /etc/systemd \
+        /lib/lsb \
+        /lib/udev \
+        /usr/lib/x86_64-linux-gnu/gconv/IBM* \
+        /usr/lib/x86_64-linux-gnu/gconv/EBC* && \
+    mkdir -p /usr/share/man/man{1..8}
```
bacongobbler: can you wrap this with `bash -c`? See https://github.com/deis/docker-base/pull/15

helgi: Can you or @mboersma explain the rationale? We'd been using `mkdir` directly for a release or two already without issues I thought?

bacongobbler: Dockerfile commands run under /bin/sh. `mkdir foo{1..8}` is bash syntax. Notice the difference between these two commands:

```
$ docker run ubuntu /bin/sh -c 'mkdir -p /usr/share/man/man{1..8} && ls /usr/share/man'
cs
da
de
es
fi
fr
hu
id
it
ja
ko
man1
man3
man5
man7
man8
man{1..8}
nl
pl
pt
pt_BR
ru
sl
sv
tr
zh_CN
zh_TW
$ docker run ubuntu /bin/bash -c 'mkdir -p /usr/share/man/man{1..8} && ls /usr/share/man'
cs
da
de
es
fi
fr
hu
id
it
ja
ko
man1
man2
man3
man4
man5
man6
man7
man8
nl
pl
pt
pt_BR
ru
sl
sv
tr
zh_CN
zh_TW
```

helgi: Hmmm - that's fair. annoying tho since it made it all the way up to ubuntu-slim :| guess it wasn't really tested the first time around :(

bacongobbler: It was originally tested, but it was broken in https://github.com/deis/docker-base/pull/13. Before then we were wrapping it with `bash -c`: https://github.com/deis/docker-base/blob/cb24562c729d9ad87abd1d74f0786cd0f925eb6f/rootfs/Dockerfile#L16

helgi: Didn't even see that change in that PR... trippy

jchauncey:
this caused the controller to crashloop

helgi:
@jchauncey thanks, already looking into it - doesn't happen on my persona cluster so I'm having to look at other places right now

rootfs/Dockerfile
helgi: This is not redundant - like lines above you can see `/usr/share/man` being included in the `rm -rf` statement

rootfs/Dockerfile
```diff
@@ -9,21 +9,39 @@ RUN adduser --system \
 
 COPY requirements.txt /app/requirements.txt
 
-RUN buildDeps='gcc git libffi-dev libpq-dev python3-dev'; \
+RUN buildDeps='gcc git libffi-dev libpq-dev python3-dev python3-pip python3-wheel python3-setuptools'; \
     apt-get update && \
     apt-get install -y --no-install-recommends \
```
helgi: Can you point me to where it is warning you off that?

mboersma: I don't have logs handy, so just ignore me. It's a pervasive issue that isn't strictly in scope for this PR, so let's fix it later.

helgi: I asked because I haven't found it in the e2e building phase but I have seen it in chef and such before. Depends if the `env` in ubuntu-slim isn't doing its job

PR #1090, opened 2016-09-27
ref(registry): remove simpleflock and depend on Docker 1.9 minimum

helgi:
Concurrent pull issue was resolved in 1.9 (https://github.com/docker/docker/issues/9718) and Kubernetes is validated against that as well

I can update https://deis.com/docs/workflow/installing-workflow/system-requirements/ with the required info if we want to go with this change

### DIFF
--- a/rootfs/registry/dockerclient.py
+++ b/rootfs/registry/dockerclient.py
@@ -7,7 +7,6 @@ import os
 import backoff
 from django.conf import settings
 from rest_framework.exceptions import PermissionDenied
-from simpleflock import SimpleFlock
 
 import docker
 import docker.constants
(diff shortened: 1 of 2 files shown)


bacongobbler:
optimally we could do a check using the socket to determine the docker version and check against that. Not required for this PR, just a thought :)

helgi:
@bacongobbler I did consider that but wanted reactions first since we've been so careful about this particular thing in the past

helgi:
https://docker-py.readthedocs.io/en/latest/api/#version is available but then we also need to have a version comparison of some sort, maybe `distutils.version.StrictVersion`

PR #1091, opened 2016-09-28
chore(CHANGELOG): remove changelog

bacongobbler:
We now publish changelogs on github and no longer update the local changelog.


helgi:
Just noticed that some of the Changelog items are not on the actual tags / releases, mostly pre-2.0 versions

Should we update those?

bacongobbler:
I think we can do that as part of deis/workflow#464, yes. I'll add a comment as a TODO.

helgi:
Suspect we need to do it with all repos at this point

PR #1096, opened 2016-09-30
feat(errors): give more context on what model is involved when 404 Not Found is involved

### LINKED ISSUE #893: Return custom message for app not found 404
For context, see deis/workflow-cli#77

This would allow the cli to give a specific error message if an application isn't found, such as `Error: Application foo not found` rather than just the generic `Error: Not Found` that is currently logged.

helgi:
Right now if you run a `deis config:list` outside of a deis repo you will still see a `Not Found` instead of a good message but that is due to how SDK / CLI handles errors.

I forced it to a random status code and get this back

```
deis config:list
Error: Unknown Error (418): {"detail":"Config was not found."}
```

Note that the app is missing but `Config` is the info since it happens within the config view, which is the only thing I get to plug into - Additionally there is no info about _what_ is missing if we were actually showing that App is missing. DAMN YOU DRF...

closes #893

### DIFF
--- a/rootfs/api/exceptions.py
+++ b/rootfs/api/exceptions.py
@@ -1,3 +1,4 @@
+from django.http import Http404
 import logging
 from rest_framework.compat import set_rollback
 from rest_framework.exceptions import APIException, status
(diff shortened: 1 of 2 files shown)


helgi:
I changed things to just return the original django Http404 since getting the real model that was erroring our is kinda hard unless we provide our own `get_object_or_404`, totally doable but it felt like an overkill.

Downside is `No App matches the given query.` is now the return data instead of `App not found.` or similar.

Thoughts?

bacongobbler:
> Downside is No App matches the given query. is now the return data instead of App not found. or similar.
> 
> Thoughts?

I think `No App matches the given query` is more than acceptable.

Would you mind adding unit tests for this?

helgi:
@bacongobbler absolutely - I just wanted to gauge people's reaction first. Added the tests.

PR #1098, opened 2016-10-02
fix(key): ssh key name has to be unique now - migration included that appends a number

### LINKED ISSUE #1023: Internal server error when removing multiple keys of the same name
Reported by @lzerma in deis/builder#416

screenshot 2016-08-29 15 53 47

Stacktrace:

```
ERROR:root:Uncaught Exception
  File "/usr/local/lib/python3.5/dist-packages/rest_framework/views.py", line 471, in dispatch
  File "/usr/local/lib/python3.5/dist-packages/rest_framework/mixins.py", line 86, in destroy
    instance = self.get_object()
  File "/usr/local/lib/python3.5/dist-packages/rest_framework/generics.py", line 97, in get_object
    obj = get_object_or_404(queryset, **filter_kwargs)
  File "/usr/local/lib/python3.5/dist-packages/rest_framework/generics.py", line 20, in get_object_or_404
    return _get_object_or_404(queryset, *filter_args, **filter_kwargs)
  File "/usr/local/lib/python3.5/dist-packages/django/shortcuts.py", line 155, in get_object_or_404
    return queryset.get(*args, **kwargs)
  File "/usr/local/lib/python3.5/dist-packages/django/db/models/query.py", line 391, in get
    (self.model._meta.object_name, num)
api.models.key.MultipleObjectsReturned: get() returned more than one Key -- it returned 7!
10.244.2.7 "DELETE /v2/keys/travis_ci HTTP/1.1" 500 25 "Deis Client vv2.4.0"
```

Looks like it shouldn't be too difficult to change delete to request an array of keys and delete all of them.

helgi:
This only fixes the API side of things and not SDK / CLI - Those are independent of one another as the API has always taken in a `name` (as `id`)

#1023 should be split into additional tickets to see if we want to make the UX improvements or just simply require people to modify the comment portion of a key to make it unique for Deis (IMO that's not amazing UX). Thoughts on that?

Fixes #1023


Joshua-Anderson:
The CLI could allow a user to optionally specify a name and override the comment in the key?

helgi:
@Joshua-Anderson Yeah, that would be one way to do it and possibly the least invasive. 

If you look at one of my suggestions in #1023 then I proposed the appending `-2` (and then keep incrementing) and propose that to the user with the ability to specify an alternative name. I liked the idea of the automation of that aspect

Issue #1062, opened 2016-09-14
Deploying from outside the app's repo with deis pull -a xxx destroys custom process types


felipejfc:
I think that this is caused by deis not detecting the app's Procfile, right? But is deleting the custom process types really the desired behavior here? Shouldn't it be better to read the Procfile from inside the image being deployed for example? Or just leaving the processes there, not deleting them if Procfile is not found but continue adding them whenever there is a Procfile within the path...

helgi:
Was the application a `git push` app originally (with a Procfile) and then you are doing a `deis pull` on top of that? Or was it always a `deis pull` app?

felipejfc:
It has mixed deploy types, the first one was a git push, later our Jenkins made a Deis Pull

kmala:
@felipejfc you can specify a procfile with custom proctypes during a `deis pull `  either through argument and keeping it in the directory of the pull.
```
deis pull --help
Creates a new build of an application. Imports an <image> and deploys it to Deis
as a new release. If a Procfile is present in the current directory, it will be used
as the default process types for this application.

Usage: deis builds:create <image> [options]

Arguments:
  <image>
    A fully-qualified docker image, either from Docker Hub (e.g. deis/example-go:latest)
    or from an in-house registry (e.g. myregistry.example.com:5000/example-go:latest).
    This image must include the tag.

Options:
  -a --app=<app>
    The uniquely identifiable name for the application.
  -p --procfile=<procfile>
    A YAML string used to supply a Procfile to the application.
```
Reading from inside the image is not going to be accepted as that would playing with the image which not many/any users want.
If you want to always add or just leave the proctypes as it is, how can one delete existing proctypes.

felipejfc:
@kmala, maybe with `deis ps:delete proctype`... I agree that getting from inside the Docker image is not a good option but deleting proctypes stealthily this way can cause confusion... It happened with a dev on my team this week.

helgi:
@felipejfc Think the big issue here is mixing deployment types - Not something we'd recommend if you have multiple process types. 

What happened in your scenario is how Deis has always worked. For `deis pull` you have to provide the `yaml` string of the Procfile to keep the behaviour you are mentioning above.

felipejfc:
Maybe just a warning that deploying will delete proctypes would be sufficient...

bacongobbler:
[...]

PR #1099, opened 2016-10-04
fix(deploy): add global settings DEIS_DEPLOY_PROCFILE_MISSING_REMOVE and DEIS_DEPLOY_REJECT_IF_PROCFILE_MISSING

helgi:
Add 2 new settings, only available globally

`DEIS_DEPLOY_REJECT_IF_PROCFILE_MISSING` rejects a deploy if the previous build had a `Procfile` but the current deploy is missing it for some reason. A `409` is thrown

`DEIS_DEPLOY_PROCFILE_MISSING_REMOVE` when set to `false` will allow an empty `Procfile` to go through without remove missing process types if the previous build had a `Procfile`

ref #1062

### DIFF
--- a/rootfs/api/models/build.py
+++ b/rootfs/api/models/build.py
@@ -2,7 +2,8 @@ from django.conf import settings
 from django.db import models
 from jsonfield import JSONField
 
-from api.models import UuidAuditedModel, DeisException
+from api.models import UuidAuditedModel
+from api.exceptions import DeisException, Conflict
 
 import logging
 logger = logging.getLogger(__name__)
@@ -78,15 +79,54 @@ class Build(UuidAuditedModel):
             raise DeisException(str(e)) from e
 
     def save(self, **kwargs):
-        removed = {}
         previous_release = self.app.release_set.filter(failed=False).latest()
-        if previous_release.build is not None:
+
+        if (
+            settings.DEIS_DEPLOY_REJECT_IF_PROCFILE_MISSING is True and
+            # previous release had a Procfile and the current one does not
+            (
+                previous_release.build is not None and
+                len(previous_release.build.procfile) > 0 and
+                len(self.procfile) == 0
+            )
+        ):
+            # Reject deployment
+            raise Conflict(
+                'Last deployment had a Procfile but is missing in this deploy. '
+                'For a successful deployment provide a Procfile.'
+            )
+
+        # See if processes are permitted to be removed
+        remove_procs = (
+            # If set to True then contents of Procfile does not affect the outcome
+            settings.DEIS_DEPLOY_PROCFILE_MISSING_REMOVE is True or
+            # previous release had a Procfile and the current one does as well
+            (
+                previous_release.build is not None and
+                len(previous_release.build.procfile) > 0 and
+                len(self.procfile) > 0
+            )
+        )
+
+        # spin down any proc type removed between the last procfile and the newest one
+        if remove_procs and previous_release.build is not None:
+            removed = {}
             for proc in previous_release.build.procfile:
                 if proc not in self.procfile:
                     # Scale proc type down to 0
                     removed[proc] = 0
 
             self.app.scale(self.owner, removed)
+
+        # make sure the latest build has procfile if the intent is to
+        # allow empty Procfile without removals
+        if (
+            settings.DEIS_DEPLOY_PROCFILE_MISSING_REMOVE is False and
+            previous_release.build is not None and
+            len(previous_release.build.procfile) > 0 and
+            len(self.procfile) == 0
+        ):
+            self.procfile = previous_release.build.procfile
 
         return super(Build, self).save(**kwargs)
 
--- a/rootfs/api/settings/production.py
+++ b/rootfs/api/settings/production.py
@@ -1,7 +1,7 @@
 """
 Django settings for the Deis project.
 """
-
+from distutils.util import strtobool
 import os.path
 import tempfile
 
@@ -255,6 +255,29 @@ BUILDER_KEY = os.environ.get('DEIS_BUILDER_KEY', random_secret)
 # k8s image policies
 SLUGRUNNER_IMAGE = os.environ.get('SLUGRUNNER_IMAGE_NAME', 'quay.io/deisci/slugrunner:canary')  # noqa
 IMAGE_PULL_POLICY = os.environ.get('IMAGE_PULL_POLICY', "IfNotPresent")  # noqa
+
+# True, true, yes, y and more evaluate to True
+# False, false, no, n and more evaluate to False
+# https://docs.python.org/3/distutils/apiref.html?highlight=distutils.util#distutils.util.strtobool
+# see the above for all available options
+#
+# If a user deploys one build with a Procfile but then forgets to in the next one
+# then let that go through without scaling the missing process types down
+#
+# If the user has a Procfile in both deploys then processes are scaled up / down as per usual
+#
+# By default the process types are scaled down unless this setting is turned on
+DEIS_DEPLOY_PROCFILE_MISSING_REMOVE = bool(strtobool(os.environ.get('DEIS_DEPLOY_PROCFILE_MISSING_REMOVE', 'true')))  # noqa
+
+# True, true, yes, y and more evaluate to True
+# False, false, no, n and more evaluate to False
+# https://docs.python.org/3/distutils/apiref.html?highlight=distutils.util#distutils.util.strtobool
+# see the above for all available options
+#
+# If a previous deploy had a Procfile but then the following deploy has no Procfile then it will
+# result in a 406 - Not Acceptable
+# Has priority over DEIS_DEPLOY_PROCFILE_MISSING_REMOVE
+DEIS_DEPLOY_REJECT_IF_PROCFILE_MISSING = bool(strtobool(os.environ.get('DEIS_DEPLOY_REJECT_IF_PROCFILE_MISSING', 'false')))  # noqa
 
 # Define a global default on how many pods to bring up and then
 # take down sequentially during a deploy
--- a/rootfs/api/tests/test_build.py
+++ b/rootfs/api/tests/test_build.py
@@ -8,6 +8,7 @@ import json
 from django.contrib.auth.models import User
 from django.core.cache import cache
 from django.conf import settings
+from django.test.utils import override_settings
 from unittest import mock
 from rest_framework.authtoken.models import Token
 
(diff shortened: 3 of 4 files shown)


rootfs/api/models/build.py
```diff
@@ -78,9 +78,14 @@ def create(self, user, *args, **kwargs):
             raise DeisException(str(e)) from e
 
     def save(self, **kwargs):
-        removed = {}
         previous_release = self.app.release_set.filter(failed=False).latest()
-        if previous_release.build is not None:
+        config = previous_release.config
+        # See if processes are permitted to be removed
+        remove_procs = bool(config.values.get('PROCFILE_REMOVE_PROCS_ON_DEPLOY', settings.PROCFILE_REMOVE_PROCS_ON_DEPLOY))  # noqa
+        print(remove_procs)
```
kmala: i think this was just for testing.

helgi: yes, indeed. removing

rootfs/api/tests/test_build.py
```diff
@@ -235,6 +236,144 @@ def test_build_default_containers(self, mock_requests):
...
+        self.assertEqual(container['type'], 'worker')
+        self.assertEqual(container['release'], 'v2')
+        # pod name is auto generated so use regex
+        self.assertRegex(container['name'], app_id + '-worker-[0-9]{8,10}-[a-z0-9]{5}')
+
+        # do another deploy for this time forget Procfile
+        url = "/v2/apps/{app_id}/builds".format(**locals())
+        body = {'image': 'autotest/example'}
+        response = self.client.post(url, body)
+        self.assertEqual(response.status_code, 201, response.data)
```
heynemann: shouldn't this be an error? I don't quite understand how it will not change my targets, yet still succeed. Am I missing something?

helgi: This is the example of default Deis, which removes processes if you do not provide any `Procfile` information. 

Look at the next test case, that's the one where the "remove processes even if Procfile is missing or has changes" - This implies that you (or your users) have to specifically scale down processes yourself if you do not want it anymore.

It could be extended to only work that way if a `Procfile` is missing vs if you provide a `Procfile` that may remove *some* of the proc types but not all.

heynemann: I'm sorry, I meant to comment this test case. Shouldn't this return a `406 - Not Acceptable` response? What actually happened in my app if it didn't scale? Did it update any units?

helgi: No, a 201 is returned because the setting is not rejecting the changes, it is saying that it won't remove any process type despite the lack of procfile or the procfile removing any process types - it is up to the end user to spin processes down by hand

helgi: To clarify, no process types get removed but all other changes are applied. New image, new config, etc - instead of rejecting the change then instead it allows it through without spinning process types down from underneath you

Are you rather looking for a reject deploy model?

heynemann: I think this is a step forward, but I was expecting that if the flag is up, the API would reject the push as it would be an "imcomplete" push if it goes through, right?
From what I grasped, the images will be there, configuration as well, but there won't be any processes running the new version. Is that correct?

helgi: They will all be there, including the processes - since we are not rejecting the deploy and we are specifically not spinning down any process types of the flag is set.

I could make it reject instead of letting changes through - could also make that a different flag to make it very configurable

helgi: @felipejfc Could you weigh in here? I believe you want to the changes to go through instead of being rejected? Or are you happy with having the changes being rejected (as a setting an Deis admin can set) which in turn forces people to provide the `Procfile`

felipejfc:
Hi @helgi, I think I agree with @heynemann... either we accept everything or we reject everything :)
However we will still be able to intentionally delete process types specifying a procfile if we want, right? I mean, with some flag like -y...

rootfs/api/tests/test_build.py
felipejfc: Hi @helgi, I think I agree with @heynemann... either we accept everything or we reject everything :)
However we will still be able to intentionally delete process types specifying a procfile if we want, right? I mean, with some flag like -y...

rootfs/api/tests/test_build.py
```diff
@@ -235,6 +236,144 @@ def test_build_default_containers(self, mock_requests):
...
+        response = self.client.get(url)
+        self.assertEqual(response.status_code, 200, response.data)
+        self.assertEqual(response.json()['structure'], {'web': 0, 'worker': 0})
+
+    @override_settings(PROCFILE_REMOVE_PROCS_ON_DEPLOY="")
+    def test_build_no_remove_process(self, mock_requests):
+        """
+        Specifically test PROCFILE_REMOVE_PROCS_ON_DEPLOY being turned off
+        and a user posting a new build without a Procfile that was previously
+        applid
```
bacongobbler: typo: applied

rootfs/api/tests/test_build.py
bacongobbler: Nope. This is a server-side feature flag. Enabled (by default) is the current behaviour. If you read the tests below, disabling it means if you do a `deis pull myimage` with no Procfile in the current directory, the process types are retained. There is no way to remove process types; it becomes append-only.

felipejfc: I think it would be preferrable if we could still delete process types if we wanted to. What do you guys think @helgi @bacongobbler?
Still, the way it is now already solves the problem of deleting process types by mistake, thanks for that 👍

bacongobbler: Sure. At this point it's probably best if you'd take the PR torch and make a patch as you see fit. You probably know the requirements of your org's workflow better than we do. ;)

bacongobbler: I don't think I'd accept a PR if it calls for a client flag that only works when the feature flag is flipped on, though. The entire intent of this PR is to make it work for this specific workflow, but keep the rest of the platform running the same.

felipejfc: > I could try to look at doing this in such a way where the flag only matters if you previously had a Procfile but none is submitted now. 

@helgi, what would it look like then?

Lets say I have this app (deploy was made with deis pull and a Procfile specified):

```
=== xapp Processes
--- cmd:
xapp-cmd-1286488693-43lgu up (v92)
--- worker:
xapp-worker-1872629609-70hu1 up (v92)
```

Later I make a deis pull without specifying a Procfile, then, deis would reject the pull with something like:

```
This deploy is not allowed because it would delete process types and you have the PROCFILE_REMOVE_PROCS_ON_DEPLOY flag set to false.

Processes before:
--- cmd:
xapp-cmd-1286488693-43lgu up (v92)
--- worker:
xapp-worker-1872629609-70hu1 up (v92)

Processes after:
--- cmd:
xapp-cmd-1286488693-43lgu up (v93)

If you really want to make the change, use the -x flag (x can be -f or -y, don't know...)

```

What do you think @helgi? @heynemann any thoughts here?

helgi: @felipejfc No, there wouldn't be a "Before / After" view - We are not planning on adding client flags that only pertain to particular server side features. At least I am not convinced that is what we want to do.

The way I described would work for your particular use case was that if you submit a build with a Procfile and then submit another build without it then Deis would let that change through without removing any process types

However if you do the first build with a Procfile that contains `cmd`, `worker` and `redis` and then someone else submits a Procfile with `cmd` and `worker` then `redis` gets deleted, which is the normal Deis behaviour.

Basically that `server flag` for your scenario would be forgiving only when people forget the `Procfile` - Does that make sense?? - It's getting a little confusing since we have 2 similar but distinctly different outcomes being discussed in this thread ;-)

Just FYI the usage scenario @heynemann has is to fully *reject* if you go from Procfile to no Procfile.

helgi: @heynemann I understand that you like it - However I still am not convinced about adding flags around this, plus it's a lot of back and forth between the CLI and the API.

I'm working on an updated PR to support both your individual cases without any client involvement.

felipejfc: @helgi ok then, can you tell me what happens in this case then: I make the first build that contains cmd, worker and api, later I make a build without a procfile, you said that:

> Deis would let that change through without removing any process types

Will all process types be updated to the new image or just cmd?

helgi: All of them will be update as the only thing it prevents is the removal of proc types.

The tests illustrate that - I will be pushing an updated version soon

felipejfc: sounds good to me then, thanks @helgi

heynemann: I think 409 Conflict is the most appropriate one, due to it's description:
```
The request could not be completed due to a conflict with the current state of the resource. This code is only allowed in situations where it is expected that the user might be able to resolve the conflict and resubmit the request. The response body SHOULD include enough information for the user to recognize the source of the conflict.
```
What this means is that in the response the user should be able to identify that no procfile was specified and be able to decide what to do next. Sounds like what we need.

heynemann: Looks great!!! :)

helgi:
All that is left then is to figure out if this should also be per app setting - I removed that notion in the last code push since Deis Operators may not wish to let their devs change that but then again you may want to have a different global setting than per app... Thoughts on making it per app as well before reviews happen?

heynemann:
We wouldn't use a per app setting, so from our perspective it does not make sense.

helgi:
@bacongobbler, @mboersma or @kmala thoughts?

rootfs/api/models/build.py
```diff
@@ -78,9 +79,37 @@ def create(self, user, *args, **kwargs):
...
+            # Reject deployment
+            raise Conflict(
+                'Last deployment had a Procfile but is missing in this deploy. '
+                'For a successful deployment provide a Procfile.'
+            )
+
+        # See if processes are permitted to be removed
+        remove_procs = (
+            # If set to True then contents of Procfile does not affect the outcome
+            settings.DEIS_DEPLOY_PROCFILE_MISSING_REMOVE is True or
```
helgi: @kmala those 2 settings do 2 different things. One is all about rejecting if you are missing a `Procfile` and the other is about letting a deploy through without scaling down any procs types if you miss the `Procfile`

I'm not sure I am following your reasoning around both of those settings being set to false - Is that something you tried doing?

I could do the reject only option or we could copy the procfile info between builds if the fear is the Build object now has empty `Procfile` field (a valid concern to have).

rootfs/api/settings/production.py
```diff
@@ -256,6 +256,29 @@
...
+# https://docs.python.org/3/distutils/apiref.html?highlight=distutils.util#distutils.util.strtobool
+# see the above for all available options
+#
+# If a user deploys one build with a Procfile but then forgets to in the next one
+# then let that go through without scaling the missing process types down
+#
+# If the user has a Procfile in both deploys then processes are scaled up / down as per usual
+#
+# By default the process types are scaled down unless this setting is turned on
+DEIS_DEPLOY_PROCFILE_MISSING_REMOVE = strtobool(os.environ.get('DEIS_DEPLOY_PROCFILE_MISSING_REMOVE', 'true'))  # noqa
```
bacongobbler: `strtobool` appears to return a number instead of a boolean. Is this what we'd like?

```
><> python3
>>> bool('true')
True
>>> bool('false')
True
>>> from distutils.util import strtobool
>>> strtobool('false')
0
>>> strtobool('true')
1
```

Perhaps `bool(strtobool(...))` would be preferred here?

rootfs/api/settings/production.py
```diff
@@ -256,6 +256,29 @@
...
+
+# True, true, yes, y and more evaluate to True
+# False, false, no, n and more evaluate to False
+# https://docs.python.org/3/distutils/apiref.html?highlight=distutils.util#distutils.util.strtobool
+# see the above for all available options
+#
+# If a previous deploy had a Procfile but then the following deploy has no Procfile then it will
+# result in a 406 - Not Acceptable
+# Has priority over DEIS_DEPLOY_PROCFILE_MISSING_REMOVE
+DEIS_DEPLOY_REJECT_IF_PROCFILE_MISSING = strtobool(os.environ.get('DEIS_DEPLOY_REJECT_IF_PROCFILE_MISSING', 'false'))  # noqa
```
bacongobbler: same here IRT the `strtobool` comment

rootfs/api/models/build.py
kmala: yes.....app structure will different than build.procfile and it causes issues with deploys and scale

rootfs/api/settings/production.py
helgi: Good point - I had not looked at the return value - I only made sure it worked in an `if` statement as per normal. `0` and `1` work fine in those cases but I'm going to make it explicit as per your suggestion. Thanks!

rootfs/api/models/build.py
helgi: @kmala in that case `app.structure` will be correct (see the tests, I verify `structure` has not changed, unless you mean something else) whereas the latest `build.procfile` will empty.

2 options then, ditch "remove if missing" setting or copy procfile info from the previous build when that setting is enabled and the Procfile info is missing. That'd let things work for scale / deploy.

kmala: yes app.structure  won't change but build.procfile will be empty which means we can't scale as the proctype validates against it and can't get the command to run for that proctype.
I want to remove this setting and just one setting for the build to pass or not when a procfile is missing.

helgi: Okay fair - going to wait for others to weigh in. 

I want to keep both setting since I can fix your concern with one line of code.

kmala: yes that can be fixed but it will definitely make things corrupt and i am not sure why do we want the build to succeed in that scenario.

helgi: @kmala because it's something the user wants, to allow services to push a new build (new image via Jenkins for example) without having to reference the `Procfile` - I see your point tho. See below for improvements.

In any case, I fixed the code and expanded the tests (see api.tests.test_build.BuildTest.test_build_no_remove_process) to prove that running a scale at the end. Prior to my fix that test failed as `build.procfile` was empty. Now it is not and the tests pass.

What are you worried will get corrupt at this point?

helgi: @kmala just to be clear I am fine with removing the second flag, I just want to make sure that's actually what we want to do since it feels like your worries are more tech related than the feature itself being useful (or not useful) - That's why I fixed your last concern to see how people feel about it now :)

helgi:
Can I get another round of reviews on this? Otherwise, I'll bump it to the 2.8 milestone, which is not  a big deal.

bacongobbler:
sorry @helgi that I wasn't able to get back to this. I'll assign myself as a reviewer so we can definitely get this in for v2.8 if we feel it's good to :shipit:.

bacongobbler:
@kmala would you mind taking a look at this?

kmala:
I am fine with this changes but i think we should document on the use of these setting.

bacongobbler:
indeed. @helgi would you please refrain from merging this until some documentation is in deis/workflow, mmkay? :)

helgi:
@bacongobbler @kmala Yeah, for sure - Was waiting for reviews so I wouldn't have to change that around since this was highly discussed.

@bacongobbler you added the "needs tests" - Which are you referring to? e2e specifically or more tests in the Controller (already wrote some)

helgi:
Doc added https://github.com/deis/workflow/pull/551

PR #1103, opened 2016-10-04
fix(urls): terminate url regex with $ so that there is no cascading attempts at URLs

helgi:
When using -a=foo for an app (see https://github.com/deis/workflow-cli/issues/22) the app name will be sent as =foo but since
=foo does not match the URL app name regex it will simply keep trying to match to the next URL in the hierarchy that can take POST which happens to match to app create URL

Anchoring all URLs to match to the end via $ prevents those cascading behaviours


ultimateboy:
I notice that 3 routes did not get the addition of `$`. Is this intentional?

helgi:
Nope - updated

PR #1107, opened 2016-10-07
feat(charts): Add helm charts for controller

kmala:


helgi:
for my own curiosity - do the templates need to be in a folder called `charts` to be packaged up correctly?

kmala:
no...i just kept them in the charts folder just to signify that the folder contains charts and nothing related to code.

kmala:
Also it will be useful for all the repos to have charts in a common folder so that we can write script to ignore e2e if the changes are just related to chart.

helgi:
Yeah, that's fair - Was more curious why that wasn't directly in `charts` instead of `charts/controller` but I hear ya. Sounds good to me.

kmala:
the reason why inside controller instead of directly charts  is because helm requires the folder name to be same as the name mentioned in the chart.yaml. here the ci will just cd to charts, chane the versions and do a `helm package <component-name>` and publish that chart

helgi:
Ok great - that's what my initial question was about :)

Another nit question - Thoughts on dropping `controller-` from the filenames? Is that too excessive / would cause confusion in general when debugging and such?

kmala:
i can remove the prefix but the name in the manifest is what is used for debugging mostly as the k8s objects are with that name and it still has deis prefix which i have plan to remove.

helgi:
Yeah, filename only - I remember the `deis-` removal part, already happened on the files which is nice :) Just got curious if it made sense now that things are even more contained than in the classic version.

Leaving it up to you if you keep the prefix or remove, was just a curiosity of mine.

PR #1110, opened 2016-10-08
docs(README) Refer to postgres setup in Makefile

seanknox:
Remove instructions for installing PostgreSQL via Homebrew; also check that Docker is installed before executing `make postgres`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,16 +50,11 @@ helmc install workflow-dev
 
 ### Postgresql
 
-Postgresql can be installed via `homebrew`:
-
+The application and tests use PostgreSQL. To start a local instance via Docker, run `make postgres` and set the following in your shell:
 ```
-brew install postgresql
-```
-
-Or via your package manager. For example, on Debian Jessie:
-
-```
-apt-get install postgresql libpq-dev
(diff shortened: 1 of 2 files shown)


deis-admin:

README.md
```diff
@@ -50,16 +50,11 @@ helmc install workflow-dev
...
-```
-brew install postgresql
-```
-
-Or via your package manager. For example, on Debian Jessie:
-
+The application and tests use PostgreSQL. To start a local instance via Docker, run `make postgres` and set the following in your shell:
 ```
-apt-get install postgresql libpq-dev
+  export PGHOST=`docker-machine ip $$(docker-machine active) 2>/dev/null || echo 127.0.0.1`
```
ultimateboy: Looks like you copied this directly from the Makefile. In that case, the `$$(docker-machine active)` gets correctly expanded to the active docker-machine's IP. However in the case of this addition to the README, there should only be a single `$`, otherwise this errors.

kmala:
@seanknox Can you rebase your PR so that we can merge it?

seanknox:
@kmala done and thanks!

PR #1116, opened 2016-10-19
fix(registration): allow admin user registration when the mode is admin_only

kmala:
This can be a first step in providing different options on what the default registration mode can be.
Now the user can install workflow with `admin_only`  as default and register himself as the admin user without having to change the manifest after install.


helgi:
Seems reasonable enough since it verifies that the no user exists first

PR #1118, opened 2016-10-23
feat(limits-cmd): accept new limits:set value type

zinuzoid:
Accept new cli format limits:set web=500m/2000m
Value is map to requests/limits parameter in Kubernetes resource management.
See https://github.com/deis/workflow/issues/562
e2e https://github.com/deis/workflow-e2e/pull/335

### DIFF
--- a/rootfs/api/serializers.py
+++ b/rootfs/api/serializers.py
@@ -19,8 +19,10 @@ from api import models
 # proc type name is alphanumeric
 # https://docs-v2.readthedocs.io/en/latest/using-workflow/process-types-and-the-procfile/#declaring-process-types
 PROCTYPE_MATCH = re.compile(r'^(?P<type>[a-z0-9]+)$')
-MEMLIMIT_MATCH = re.compile(r'^(?P<mem>[0-9]+(MB|KB|GB|[BKMG]))$', re.IGNORECASE)
-CPUSHARE_MATCH = re.compile(r'^(?P<cpu>[-+]?[0-9]*\.?[0-9]+[m]{0,1})$')
+MEMLIMIT_MATCH = re.compile(
+    r'^(?P<mem>(([0-9]+(MB|KB|GB|[BKMG])|0)(/([0-9]+(MB|KB|GB|[BKMG])))?))$', re.IGNORECASE)
+CPUSHARE_MATCH = re.compile(
+    r'^(?P<cpu>(([-+]?[0-9]*\.?[0-9]+[m]?)(/([-+]?[0-9]*\.?[0-9]+[m]?))?))$')
 TAGVAL_MATCH = re.compile(r'^(?:[a-zA-Z\d][-\.\w]{0,61})?[a-zA-Z\d]$')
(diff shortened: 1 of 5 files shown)


deis-admin:

helgi:
Could the CLI/SDK not split this up and pass it in differently so the regex doesn't have to support a different format?

zinuzoid:
@helgi not sure if you mean that in CLI we'll split `/` and send to controller by array?

helgi:
@zinuzoid yeah - pretty much. Mostly wondering out loud if that made more sense than to complicate the validation regex

zinuzoid:
@helgi agreed, that would be a cleaner solution. but I have some question

- should we still keep backward compatibility with older CLI (keep accept string format value + new array format) or just don't accept it anymore. (btw currently it'll show warning message about version mismatch)
- how should we save to database, currently we have 1 field for this value. should I again join it with some character eg. "/", or update schema?

helgi:
I'd say BC should be kept, scheme updated as well. 

My ideal scenario is to not have `deis limits` set both based on order. `/` as a separate where requests are one side and limits are other seems prone to error (CLI and API). Optional flag could work best (IMHO but I'm willing to hear the downsides of that) but given how we have cpu/mem flags already it is already a bit awkward

zinuzoid:
I would prefer optional flag also. But somehow current CLI accept multiple appType in limits command. eg. `deis limits:set web=2 db=3 --cpu` possibly we could do `deis limits:set web=2 db=3 --requests --cpu` in this case if user wants to set both requests/limits they would have to make at least 2 releases.
One problem that we discussed in https://github.com/deis/workflow/issues/562 is Kubernetes itself will default requests value if we don't explicitly provide. Let's say if user set only limits, then should we pass to Kubernetes as request=0/limit=X or just limit=X(which will result in request=X/limit=X internally, maybe we'll keep this way for BC)

In case we might want to go with this `deis limits:set web=2 db=3 --requests --cpu` approach, we can split command to `deis requests:set web=2 db=3 --cpu`.

3 choices so far to solve this.

**1. leave as it is**
`deis limits:set web=2/3 db=3/5 --cpu`

**2. add another flag to limits:set + update database schema**
`deis limits:set web=2 db=3 --requests --cpu`

**3. split to new command requests:set + update database schema**
`deis requests:set web=2 db=3 --cpu`

cc @kmala @bacongobbler if you have some input.

kmala:
@helgi i know this is not the best UX but the way we have our model around limits, i would think this is probably the best way we can go without worrying about breaking things.

kmala:
Tested manually and this looks good,
```
 deis limits:set web=10M/100M
Applying limits... done

=== syrupy-yeomanry Limits

--- Memory
web     10M/100M

--- CPU
Unlimited

deis limits:set web=100M
Applying limits... done

=== syrupy-yeomanry Limits

--- Memory
web     100M

--- CPU
Unlimited
```

kmala:
@zinuzoid can you rebase you PR and so we can merge it

kmala:
Jenkins, Add to whitelist.

zinuzoid:
done

PR #1121, opened 2016-10-26
Ensure DEIS_DEBUG==true for debug output

seanknox:
Currently debug output is turned on if `DEIS_DEBUG` is set to anything (including `DEIS_DEBUG=FALSE`). This requires it be explicitly `true`, like the rest of the code base.

### DIFF
--- a/rootfs/bin/boot
+++ b/rootfs/bin/boot
@@ -7,7 +7,7 @@
 set -eo pipefail
 
 # set debug based on envvar
-[[ $DEIS_DEBUG ]] && set -x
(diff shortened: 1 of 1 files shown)


deis-admin:

kmala:
can you change the commit as per http://docs.deis.io/en/latest/contributing/standards/#commit-style-guide so that this is picked by the Changelog

seanknox:
@kmala fixed

kmala:
Jenkins, Add to whitelist.

PR #1129, opened 2016-11-03
chore(requirements): update DRF to 3.5.2

mboersma:
See http://www.django-rest-framework.org/topics/release-notes/#352

### DIFF
--- a/rootfs/requirements.txt
+++ b/rootfs/requirements.txt
@@ -3,7 +3,7 @@ backoff==1.3.1
 Django==1.10.3
 django-cors-middleware==1.3.1
 django-guardian==1.4.6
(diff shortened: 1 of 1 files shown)


bacongobbler:
Looks like a legitimate bug? Thank goodness for unit tests

mboersma:
> Looks like a legitimate bug?

It does, but `make test-unit` passes for me locally with DRF 3.5.2. Odd...

mboersma:
It passed Travis CI after I clicked the "retry" button. That's a bit worrisome, since the unit tests should be deterministic. I also haven't been able to reproduce the failure locally running the tests several times.

PR #1132, opened 2016-11-04
fix(proctype): Change the regex used for validating proctypes

kmala:
fixes https://github.com/deis/controller/issues/1124

Not adding the regex to support underscore because kubernetes doesn't support it in the names and it is better to not add it than make changes internally from `_` to `-` which is might be prone to errors.

But if it is really important to have support for the `_` then i am open to adding it.

### DIFF
--- a/rootfs/api/serializers.py
+++ b/rootfs/api/serializers.py
@@ -18,7 +18,7 @@ from api import models
 
 # proc type name is alphanumeric
 # https://docs-v2.readthedocs.io/en/latest/using-workflow/process-types-and-the-procfile/#declaring-process-types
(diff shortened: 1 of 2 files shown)


bacongobbler:
Works for me!

bacongobbler:
@kmala would you mind changing the reflex to not allow dashes as the first character? I think that's one edge case we shouldn't support.

bacongobbler:
e.g. `-web` or `-` should not be acceptable values

kmala:
@bacongobbler i had changed it to allow only alphanumeric separated by hypens.

bacongobbler:
Great! Would you mind adding tests for the following process types such that they will fail?

```
-
-web
web-
```

kmala:
Added tests

PR #1146, opened 2016-11-22
fix(api): transfer all downstream resources along with app

### LINKED ISSUE #1126: deis apps:transfer does not transfer downstream resources
As seen in https://github.com/deis/deis/issues/5019#issuecomment-214624914, any downstream resource like releases aren't transferred when calling `deis apps:transfer`, causing the release list to be empty with no way to re-deploy or restart the application.

We should be mindful that multiple users can modify an app thanks to `deis perms`, so we should cascade only to resources owned by the user we are transferring from.

bacongobbler:
This gives the target user complete ownership over the releases, configuration and
other resources required when transferring an app's ownership.

closes #1126 
closes deis/deis#5019

### DIFF
--- a/rootfs/api/fixtures/tests.json
+++ b/rootfs/api/fixtures/tests.json
@@ -52,5 +52,23 @@
         "email": "autotest3@deis.io",
         "date_joined": "2013-05-10T16:08:09.357Z"
     }
+},
+{
+    "pk": 10,
+    "model": "auth.user",
+    "fields": {
+        "username": "autotest4",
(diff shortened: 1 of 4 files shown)


vdice:
LGTM -- wondering if there is any related scenario that we want to add to e2e (if not already)?  or is unit coverage sufficient?

bacongobbler:
Unit tests should be sufficient. If we add e2e tests we're just covering the same ground twice. :)

PR #1147, opened 2016-11-22
fix(api): ensure a 409 is raised when cancelling a user with downstream objects

bacongobbler:
This ensures a 409 error is raised with the proper error to ensure the user knows why they cannot cancel the account. Beforehand it would show a 500 error because `django.db.models.deletion.ProtectedError` would be raised.

### DIFF
--- a/rootfs/api/views.py
+++ b/rootfs/api/views.py
@@ -8,6 +8,7 @@ from django.shortcuts import get_object_or_404
 from guardian.shortcuts import assign_perm, get_objects_for_user, \
     get_users_with_perms, remove_perm
 from django.views.generic import View
+from django.db.models.deletion import ProtectedError
 from rest_framework import mixins, renderers, status
 from rest_framework.exceptions import PermissionDenied, NotFound, AuthenticationFailed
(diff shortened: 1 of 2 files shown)


rootfs/api/views.py
```diff
@@ -86,11 +87,14 @@ def destroy(self, request, **kwargs):
 
         # A user can not be removed without apps changing ownership first
         if len(models.App.objects.filter(owner=target_obj)) > 0:
-            msg = '{} still has applications assigned. Delete or transfer ownership'.format(str(target_obj))  # noqa
+            msg = '{} still has applications assigned. Please delete or transfer ownership'.format(str(target_obj))  # noqa
```
kmala: this would mean the controller sdk also needs to be changed.
https://github.com/deis/controller-sdk-go/blob/master/errors.go#L32
https://github.com/deis/controller-sdk-go/blob/master/errors.go#L209

bacongobbler: Okay if it's too much trouble I'll just remove this commit for now. It was more of a QoL change. I'll make a separate PR for this if I feel motivated :)

PR #1149, opened 2016-11-23
Add basic support of resource quotas for applications

rvadim:
If set KUBERNETES_NAMESPACE_DEFAULT_QUOTA_SPEC to something like:
`{"spec":{"hard":{"cpu":"3","pods":"10","secrets":"10"}}}`
```json
{
    "spec": {
        "hard": {
            "cpu": "3",
            "pods": "10",
            "secrets": "10"
        }
    }
}
```
For all application namespaces will be added resource quota:
```bash
$ kubectl --namespace test20 get quota
NAME           AGE
test20-quota   35m
$ kubectl --namespace test20 get quota -o yaml
apiVersion: v1
items:
- apiVersion: v1
  kind: ResourceQuota
  metadata:
    labels:
      app: test20
      heritage: deis
    name: test20-quota
    namespace: test20
  spec:
    hard:
      cpu: "3"
      pods: "10"
      secrets: "10"
  status:
    hard:
      cpu: "3"
      pods: "10"
      secrets: "10"
    used:
      cpu: "0"
      pods: "0"
      secrets: "1"
kind: List
metadata: {}
```
We need this PR to restrict users from creation applications with a lot of resource usage. In our staging there are near 200+ applications and its count rise every month. This PR can save ops time and protect k8s nodes from overflow.

Documentation PR https://github.com/deis/workflow/pull/613

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -215,7 +215,8 @@ class App(UuidAuditedModel):
                 quota_name = '{}-quota'.format(namespace)
                 self.log(settings.KUBERNETES_NAMESPACE_DEFAULT_QUOTA_SPEC)
                 quota_spec = json.loads(settings.KUBERNETES_NAMESPACE_DEFAULT_QUOTA_SPEC)
-                self.log('creating Quota {} for namespace {}'.format(quota_name, namespace), level=logging.DEBUG)
+                self.log('creating Quota {} for namespace {}'.format(quota_name, namespace),
+                         level=logging.DEBUG)
                 try:
                     self._scheduler.quota.get(namespace, quota_name)
                 except KubeException:
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -203,6 +203,7 @@ class App(UuidAuditedModel):
         # create required minimum resources in k8s for the application
         namespace = self.id
         service = self.id
+        quota_name = '{}-quota'.format(self.id)
         try:
             self.log('creating Namespace {} and services'.format(namespace), level=logging.DEBUG)
             # Create essential resources
@@ -212,8 +213,6 @@ class App(UuidAuditedModel):
                 self._scheduler.ns.create(namespace)
 
             if settings.KUBERNETES_NAMESPACE_DEFAULT_QUOTA_SPEC != '':
-                quota_name = '{}-quota'.format(namespace)
-                self.log(settings.KUBERNETES_NAMESPACE_DEFAULT_QUOTA_SPEC)
                 quota_spec = json.loads(settings.KUBERNETES_NAMESPACE_DEFAULT_QUOTA_SPEC)
                 self.log('creating Quota {} for namespace {}'.format(quota_name, namespace),
                          level=logging.DEBUG)
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -211,6 +211,16 @@ class App(UuidAuditedModel):
             except KubeException:
                 self._scheduler.ns.create(namespace)
 
+            if settings.KUBERNETES_NAMESPACE_DEFAULT_QUOTA_SPEC != '':
+                quota_name = '{}-quota'.format(namespace)
+                self.log(settings.KUBERNETES_NAMESPACE_DEFAULT_QUOTA_SPEC)
+                quota_spec = json.loads(settings.KUBERNETES_NAMESPACE_DEFAULT_QUOTA_SPEC)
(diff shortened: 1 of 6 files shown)


deis-admin:

bacongobbler:
So this is a global quota across all applications? This looks good as a first pass, but I'm curious if there's a way we can curate individual quotas for each application, but only allow administrators access to the endpoint. It'd probably also need some schema validation so we're not injecting invalid quota specs into the cluster.

rvadim:
> So this is a global quota across all applications?
No, this is per application quota

> It'd probably also need some schema validation so we're not injecting invalid quota specs into the cluster.
Ok, im try to write some negative test cases for invalid schema and handle this scenarios

kmala:
Jenkins, Add to whitelist.

rootfs/api/models/app.py
```diff
@@ -211,6 +211,17 @@ def create(self, *args, **kwargs):  # noqa
             except KubeException:
                 self._scheduler.ns.create(namespace)
 
+            if settings.KUBERNETES_NAMESPACE_DEFAULT_QUOTA_SPEC != '':
```
bacongobbler: What I meant to say is that you cannot tune the default quota spec for an individual application. This is a global quota setting that applies across all applications. :)

rootfs/api/models/app.py
```diff
@@ -211,6 +211,17 @@ def create(self, *args, **kwargs):  # noqa
...
+                quota_name = '{}-quota'.format(namespace)
+                self.log(settings.KUBERNETES_NAMESPACE_DEFAULT_QUOTA_SPEC)
+                quota_spec = json.loads(settings.KUBERNETES_NAMESPACE_DEFAULT_QUOTA_SPEC)
+                self.log('creating Quota {} for namespace {}'.format(quota_name, namespace),
+                         level=logging.DEBUG)
+                try:
+                    self._scheduler.quota.get(namespace, quota_name)
```
bacongobbler: If an administrator were to update the default quota, the change won't get applied here without first removing the old quota. Can we make a change here to update the quota if it differs from the expected spec??

rvadim: It's a bad idea, because an administrator may already changes quota through kubernetes interface for some apps and restart of controller rollback this changes.  But yes, there is problem.

mboersma:
There's a merge commit in this PR. Could you rebase it against current master?

rootfs/api/models/app.py
rvadim: The use case is the following:
- Admin sets default quota
- Users create applications (default quota is applied for each application)
- One of the users need more resources for particular application (my-shiny-app)
- Admin changes quota for my-shiny-app namespace with help of kubectl
With hundreds of users and applications it is impossible to run production installation without default quota - it is guaranteed someone will ruin the cluster by using all available memory and CPU.

rvadim:
Rebase done.

rootfs/api/models/app.py
rvadim: @bacongobbler im need your help here =)

bacongobbler: ping @kmala @mboersma any thoughts on this?

kmala: I accept that having default quota is good and i am for it but adding a feature/enhancement which is asking two different tools(deis and kubectl) to manage it is not something most people like. But this can be a first step and may be create an issue to tackle in the update scenario afterwards.

deis-admin:

rootfs/api/models/app.py
rvadim: @bacongobbler @kmala Any additional steps from me?

rvadim:
@mboersma whats next? I can add any changes. Im ready to improve, fix, etc. We need this PR to migrate to deis 2.0. We can contribute more after migration.

rootfs/api/models/app.py
bacongobbler: My only problem with this comes not with the use case, but with changing the default quota later. For example:

 - Admin sets default quota
 - Users create applications
 - Admin adds more nodes to the cluster
 - Admin wants to increase the default quota on all existing apps

Right now this would have to be handled outside of Workflow. I don't necessarily enjoy that. Do you have any thoughts on how we could roll that out eventually?

I think the feature is good enough to ship at the moment, just concerned about rolling out updates in the future and what that would look like.

kmala:
I would think the per app resource quota can be done by adding a new variable to the appsettings model which would require the changes to cli also.But this Pr looks good to me.

mboersma:
LGTM3. Thanks @rvadim!

rvadim:
Great, thank you!  :beers:

PR #1157, opened 2016-12-07
feat(api): validate a certificate's private key

### LINKED ISSUE #1155: Validate certificate on certs:add
In the current implementation when you add an invalid certificate, deis-controller adds it and supplies it to the deis-router. Nginx tries to reload and fails with the following error:
```
nginx: [emerg] SSL_CTX_use_PrivateKey_file("/opt/router/ssl/foo.key") failed (SSL: error:0B080074:x509 certificate routines:X509_check_private_key:key values mismatch)
```
This has the following consequences:
1. Nginx can no longer reload. If you add a new application, it will always give 404. New nginx config will be generated, but nginx can never load it because of certificate issue. At least existing applications will run.
2. Should the router pod die, Kube will try to restart it and will always fail because nginx cannot start with incorrect certificate. This makes all applications inaccessible. Furthermore, this makes `deis` subdomain inaccessible as well, which makes it impossible to remove incorrect certificate via `deis certs:remove`.

Thus the controller should validate an uploaded certificate when the `deis certs:add` command is performed and give user a feedback if the certificate is incorrect.

bacongobbler:
closes #1155

### DIFF
--- a/rootfs/api/models/certificate.py
+++ b/rootfs/api/models/certificate.py
@@ -66,9 +66,16 @@ def get_subj_alt_name(peer_cert):
 
 def validate_certificate(value):
     try:
-        crypto.load_certificate(crypto.FILETYPE_PEM, value)
+        return crypto.load_certificate(crypto.FILETYPE_PEM, value)
     except crypto.Error as e:
         raise ValidationError('Could not load certificate: {}'.format(e))
+
+
+def validate_private_key(value):
+    try:
+        return crypto.load_privatekey(crypto.FILETYPE_PEM, value)
+    except crypto.Error as e:
+        raise ValidationError('Could not load private key: {}'.format(e))
 
 
 class Certificate(AuditedModel):
@@ -79,7 +86,7 @@ class Certificate(AuditedModel):
     name = models.CharField(max_length=253, unique=True, validators=[validate_label])
     # there is no upper limit on the size of an x.509 certificate
     certificate = models.TextField(validators=[validate_certificate])
-    key = models.TextField()
+    key = models.TextField(validators=[validate_private_key])
     # X.509 certificates allow any string of information as the common name.
     common_name = models.TextField(editable=False, unique=False, null=True)
     # A list of DNS records if certificate has SubjectAltName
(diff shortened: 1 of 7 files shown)


bacongobbler:
@ineu feel free to give this one a try. It should fix your issue.

ineu:
Thanks @bacongobbler, going to try. Could you please point me to some documentation on how to update a single component? And also is there some kind of automatic build or do I need to build a Docker image from the repo?

bacongobbler:
You can just use the image CI built and uploaded to quay.io. After installing v2.9.0, try

```
kubectl --namespace=deis edit deployment deis-controller
```

Change the image to `quay.io/deisci/controller:git-5b890a2` and Bob's your uncle.

rootfs/api/tests/test_certificate.py
```diff
@@ -176,6 +177,31 @@ def test_load_invalid_cert(self):
...
             )
 
+    def test_create_invalid_key(self):
+        """Upload a private key that can't be loaded by pyopenssl"""
+        response = self.client.post(
+            self.url,
+            {
+                'name': 'random-test-cert',
+                'certificate': self.cert,
+                'key': 'I am Groot.'
```
mboersma: 😺

rootfs/api/models/certificate.py
```diff
@@ -106,14 +113,23 @@ def domains(self):
     def __str__(self):
         return self.name
 
-    def _get_certificate(self):
+    def _load_certificate(self):
         try:
```
mboersma: Wouldn't this be a bit more DRY if it just called `validate_private_key(self.certificate)` from above and re-raised `ValidationError` as `SuspiciousOperation`?

bacongobbler: It would be, absolutely. I can refactor this.

ineu:
Tried it:
```
> $ deis certs:add foo bundle.crt bundle.crt
Adding SSL endpoint... Error: Unknown Error (400): {"key":["Could not load private key: []"]}
```
So if you upload something other but the correct private key it throws an error, as expected.

Though it's not the problem I've faced in #1155. I think to completely cover the cases when nginx would print an error (and fail to start) following usecases should be taken into account:

1. User uploads incorrect private key: Fixed by this PR
2. User uploads private key not matching the certificate (my case from #1155)
3. User joins cert and intermediate cert in the wrong order.

bacongobbler:
Thank you for providing your use cases. It wasn't clear in the original ticket.

If you feel like hacking on some tests/fixes for this in another PR that'd be great. I'll keep your ticket open in the meantime.

PR #1158, opened 2016-12-08
fix(apiserver): Add an option to skip ssl verification when interacti…

kmala:


charts/controller/templates/controller-deployment.yaml
```diff
@@ -58,6 +58,8 @@ spec:
             # NOTE(bacongobbler): use deis/registry_proxy to work around Docker --insecure-registry requirements
             - name: "DEIS_REGISTRY_SERVICE_HOST"
               value: "localhost"
+            - name: "K8S_API_SECURE"
```
slack: Not blocking, but is something like `K8S_API_VERIFY_TLS` `true/false`?

PR #1161, opened 2016-12-08
chore(Makefile): remove pyvenv script

bacongobbler:
pyvenv has been officially deprecated in favour of `python3 -m venv` as of the python 3.6 announcement made this morning.

This command also works in previous python versions. I have manually tested this with python 3.5.2 without issues. The `@if [ ! -d venv ];` check is no longer necessary as the venv module identifies if it already exists and just exits zero.


mboersma:
Futureproofing++. I also tested this locally.

PR #1162, opened 2016-12-08
[WIP] feat(rootfs): add channels and serve with daphne

mboersma:
This is the start of a `deis run bash` prototype I did a month ago using Django channels. It passes all e2e tests locally, so I'm mostly just curious to see what happens to it in CI.

I'm unconvinced this overall approach would work for bi-directional PTY sessions, and we have several other ideas about how to implement it. So please don't waste time reviewing this because it may never see the light of day.


sheerun:
Why don't you use grpc?

sheerun:
or whatever kubectl uses

mboersma:
> Why don't you use grpc?

Django channels has direct support for WebSockets, and the existing API is all HTTP/REST. So I think this is more consistent and much less code to write.

This is just the beginning of a proposed change and I haven't updated it in a while, BTW, but I'm hoping to tie things off soon and see if other contributors can help finish it up.

felixbuenemann:
@mboersma Is there a follow-up issue that discusses the implementation of this feature?

PR #1168, opened 2016-12-09
feat(api): add deploy hooks

### LINKED ISSUE #1153: implement webhooks
Continuation of https://github.com/deis/deis/issues/1486. While `kubectl get events` works for the power user that wants to see cluster events, one high-level important events that administrators may want to perform certain chat-ops or other webhook-based events would be when an application is deployed. The initial plan is to implement a "deploy hook" which is a generic HTTP POST sender to a set of configured URLs when an application release is deployed. This will be implemented as a cluster-wide setting, where administrators can add webhook URLs to receive these payloads.

bacongobbler:
closes #1153.


Joshua-Anderson:
There doesn't seem to be any way to authenticate the hooks, is that going to be a future feature?

bacongobbler:
From the trello ticket which I'm working from:

> Operators should be able to configure webhooks URLS that would receive release events from Workflow. Let's start cluster-wide to start, rather than per-application.

There were no requirements to authenticate the hooks. Can you describe in more detail how that would be sussed out? As it stands right now these hooks are injected by the administrator through the chart so it's generally accepted to be trusted URLs, but if there's a way to add that sorta functionality I can quickly hack it up in here.

Joshua-Anderson:
As a server that's receiving the webhook, it's important to be able to verify that the hook actually came from deis. Otherwise $EVIL_USER could send a fake post request to your server and mimic a deis deployment.

One method is using a [secret to create a HMAC digest of body, like github](https://developer.github.com/v3/repos/hooks/#example).

Another method is signing the web hooks with a private key, then verify it using the public key, which can be retrieved from the controller, [like travis-ci does](https://docs.travis-ci.com/user/notifications/#Verifying-Webhook-requests).

bacongobbler:
I think I'll do the dumb approach of a secret passphrase shared by both the controller and the deploy hook. Thanks for the feedback, @Joshua-Anderson :)

bacongobbler:
done, now just need to update the documentation to and write a blog post to demonstrate the HMAC-SHA1 signing process.

bacongobbler:
Also realized I should handle the use case when one of the deploy hook URLs are not responding correctly

bacongobbler:
That use case is now being handled as of 47cabd8.

bacongobbler:
@vdice mentioned to me it might be better to roll the secret key into a secret rather than an envvar. I'll fix that right up!

PR #804, opened 2016-06-08
fix(boot): change group ownership of docker socket

bacongobbler:
# Summary of Changes

on some providers, the docker socket is owned by the root group. We want it to be owned by docker
so the controller can connect to the socket.

# Issue(s) that this PR Closes

closes #801 

# Associated [End To End Test](https://github.com/deis/workflow-e2e) PR(s)

N/A since this occurs only on vagrant

# Associated [Documentation](https://github.com/deis/workflow) PR(s)

N/A

# Associated [Design Document](http://docs.deis.io/en/latest/contributing/design-documents)(s)

N/A

# Testing Instructions

1. Create a Deis Cluster on Vagrant using kube-up.sh
2. Register an app
3. Deploy an app using `deis pull`

Results:

1. The app called "abcd" should be deployed.

# Pull Request Hygiene TODOs

Please make sure the below checklist is complete.

- [x] Your pull request is concise and to the point (make another PR for refactoring nearby code)
- [x] Your commits are squashed into logical units of work
- [x] Your commits follow the [commit style guidelines](http://docs.deis.io/en/latest/contributing/standards/#commit-style-guide)

:cherry_blossom: Thank you! :cherry_blossom:

PR #1171, opened 2016-12-11
fix(boot): Don't change group ownership of docker socket

### LINKED ISSUE #1170: Controller changes GID of Docker socket
The startup script for the controller changes the gid of the host's docker.sock to match the gid of the "deis" group inside the controller, which causes members of the "docker" group on the host to loose access to the socket (for example on CoreOS the default "core" user is part of the "docker" group).

It is really bad practice to have a container modify permissions of files it doesn't own inside the host filesystem, so this should be fixed.

This change was made to fix issue #804 where the docker.sock was owned by root on some systems.

I propose a different approach, which would cause the "deis" user to be added to the group with the same gid as the docker socket, if such a group exists inside the container and otherwise add a new group "docker" with the same gid as the docker socket. This could mean that the "deis" user is added to the "root" group (gid 0) on the host on systems affected by #804.

I'll open a PR shortly with the proposed fix (**update:** see #1171).

felixbuenemann:
This is a better fix for issue #804 and fixes #1170:

* The gid of the docker socket is no longer changed to that of the deis group (gid 107).
* If an existing group in the container has the same gid as the socket, we add the deis user to that group (unless of course it is the "deis" group).
* If no such group exists, we create a new "docker" group and add the user to that (like before @bacongobbler's hack in b27c816e9ef21a989c68db23c0c98c02f726cf7b was added, which effectively disabled this code).

### DIFF
--- a/rootfs/bin/boot
+++ b/rootfs/bin/boot
@@ -16,12 +16,16 @@ python --version
 mkdir -p /app/data/logs
 chmod -R 777 /app/data/logs
 
-# HACK(bacongobbler): explicitly add the docker socket to the deis group
-chgrp deis /var/run/docker.sock
-
-# allow deis user group permission to Docker socket
-groupadd -g "$(stat -c "%g" /var/run/docker.sock)" docker || true
-usermod -a -G docker deis || true
+# modify deis user groups to grant access to Docker socket
+DOCKER_SOCKET_GID=$(stat -c "%g" /var/run/docker.sock)
+DOCKER_SOCKET_GROUP=$(getent group "$DOCKER_SOCKET_GID" | cut -d : -f 1 || :)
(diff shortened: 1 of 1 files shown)


deis-admin:

felixbuenemann:
As noted in #1170 this could cause the "deis" user to become part of the "root" group, if the socket is own by that group. We could extend the script to echo a warning in that case.

felixbuenemann:
I manually checked that the right thing happens if the socket on the host is currently owner by either "docker", "deis" or "root" by using `kubectl exec -n deis <deis-controller-pod> curl --unix-socket /var/run/docker.sock http:/containers/json` to list the running containers through the mounted socket.

kmala:
Jenkins, Add to whitelist.

PR #1182, opened 2016-12-27
Handle cases with namespace quotas limits sets

rvadim:
# Overview
This PR handle 2 specific cases:
- User did not set quota for applications or set it incorrectly
- User tries to scale the application when limits are already exceeded (overuse)

# Problem
Аdmin set default quota for namespaces (memory request and limit to1024M).
User tries to deploy application my-shiny-app (w/o setting limits for application): git push deis master
Deis controller return success after build and launch, but actually no pods are started because kubernates rejects pods creation.
Kubernates controller manager add event message to namespace with description:
```
Error creating: pods "my-shiny-app-web-1758015041-" is forbidden: failed quota: my-shiny-app-quota: must specify limits.memory,requests.memory
```
If quota is enabled in a namespace for compute resources like cpu and memory, users must specify requests or limits for those values; otherwise, the quota system may reject pod creation. 
Source: http://kubernetes.io/docs/admin/resourcequota/

Similar description when user set limits, but due to overuse next happened:
```
Error creating: pods "my-shiny-app-web-3136383174-" is forbidden: exceeded quota: my-shiny-app-quota, requested: requests.memory=600Mi, used: requests.memory=400Mi, limited: requests.memory=800Mi
```

So, user or CI/CD script make sure, that all are correct. But in fact, some actions needed from user.

# Solution
Explicit error return code with descriptive message for this situation.

### DIFF
--- a/rootfs/scheduler/resources/deployment.py
+++ b/rootfs/scheduler/resources/deployment.py
@@ -237,6 +237,12 @@ class Deployment(Resource):
             self.log(namespace, 'Deploy operation for Deployment {} in has expired. Rolling back to last good known release'.format(name), level='DEBUG')  # noqa
             return False, True
 
+        try:
+            self._check_for_failed_events(namespace, labels=labels)
+        except KubeException as e:
+            self.log(namespace, e)
+            return False, True
+
         return True, False
 
     def are_replicas_ready(self, namespace, name):
--- a/rootfs/scheduler/resources/deployment.py
+++ b/rootfs/scheduler/resources/deployment.py
@@ -326,6 +326,9 @@ class Deployment(Resource):
         timeout = len(batches) * deploy_timeout
         self.log(namespace, 'This deployments overall timeout is {}s - batch timout is {}s and there are {} batches to deploy with a total of {} pods'.format(timeout, deploy_timeout, len(batches), replicas))  # noqa
 
+        # check for failed events(when quota exceeded for example)
+        self._check_for_failed_events(namespace, labels=labels)
+
         waited = 0
         while waited < timeout:
             ready, availablePods = self.are_replicas_ready(namespace, name)
@@ -351,6 +354,39 @@ class Deployment(Resource):
         ready, _ = self.are_replicas_ready(namespace, name)
         if not ready:
             self.pod._handle_not_ready_pods(namespace, labels)
+
+    def _check_for_failed_events(self, namespace, labels):
+        """
+        Request for new ReplicaSet of Deployment and search for failed events involved by that RS
+        Raises: KubeException when RS have events with FailedCreate reason
+        """
+        response = self.rs.get(namespace, labels=labels)
+        data = response.json()
+        fields = {
+            'involvedObject.kind': 'ReplicaSet',
+            'involvedObject.name': data['items'][0]['metadata']['name'],
+            'involvedObject.namespace': namespace,
+            'involvedObject.uid': data['items'][0]['metadata']['uid'],
+        }
+        events_list = self.ns.events(namespace, fields=fields).json()
+        events = events_list.get('items', [])
+        if events is not None and len(events) != 0:
+            for event in events:
+                if event['reason'] == 'FailedCreate':
+                    log = self._get_formatted_messages(events)
+                    self.log(namespace, log)
+                    raise KubeException(log)
+
+    @staticmethod
+    def _get_formatted_messages(events):
+        """
+        Format each event by string and join all events to one string
+        """
+        message_format = 'Message:{message}, lastTimestamp:{lastTimestamp}, reason: {reason}, count: {count}'  # noqa
+        output = []
+        for event in events:
+            output.append(message_format.format(**event))
+        return '\n'.join(output)
 
     def _get_deploy_steps(self, batches, tags):
         # if there is no batch information available default to available nodes for app
--- a/rootfs/scheduler/resources/pod.py
+++ b/rootfs/scheduler/resources/pod.py
@@ -732,7 +732,7 @@ class Pod(Resource):
             self.log(namespace, log)
             raise KubeException(log)
         else:
-            raise RuntimeError('Unable to start pods, no events in namespace found.')  # noqa
+            return None
 
     def _get_namespace_events(self, namespace, event_type='Warning'):
         """
--- a/rootfs/scheduler/resources/pod.py
+++ b/rootfs/scheduler/resources/pod.py
@@ -719,42 +719,6 @@ class Pod(Resource):
                     message = "\n".join([x.strip() for x in event['message'].split("\n")])
                     raise KubeException(message)
 
-        # Handle cases when no one pods is started (quota check fialed)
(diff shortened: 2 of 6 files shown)


deis-admin:

kmala:
Jenkins, Add to whitelist.

bacongobbler:
Hey @rvadim, sorry for letting this one fall through the cracks.

If you wouldn't mind amending your commit such that it follows the [commit style guide](https://deis.com/docs/workflow/contributing/submitting-a-pull-request/#commit-style), that would be awesome! That way it'll get picked up by `deisrel` when we generate the release changelog. Thanks!

bacongobbler:
the code itself looks great, though. 👍

rootfs/scheduler/resources/pod.py
```diff
@@ -719,7 +719,41 @@ def _handle_not_ready_pods(self, namespace, labels):
...
+        # http://kubernetes.io/docs/admin/resourcequota/
+        # or when scale process trigger overuse case for some pods:
+        # For example: used 3/3 pods, but user scale to 4
+        events = self._get_namespace_events(namespace)
+        if len(events) != 0:
+            log = self._get_formatted_messages(events)
+            self.log(namespace, log)
+            raise KubeException(log)
+        else:
+            raise RuntimeError('Unable to start pods, no events in namespace found.')  # noqa
```
bacongobbler: actually, this code block will always return an error which is not what is intended here. It'll either raise a KubeException when there are events, or it'll raise a RuntimeError if there were no events. Wouldn't we want it to `return None` if there were no bad events?

rootfs/scheduler/resources/pod.py
```diff
@@ -719,7 +719,41 @@ def _handle_not_ready_pods(self, namespace, labels):
...
+        # memory, users must specify requests or limits for those values;
+        # otherwise, the quota system may reject pod creation.
+        # http://kubernetes.io/docs/admin/resourcequota/
+        # or when scale process trigger overuse case for some pods:
+        # For example: used 3/3 pods, but user scale to 4
+        events = self._get_namespace_events(namespace)
+        if len(events) != 0:
+            log = self._get_formatted_messages(events)
+            self.log(namespace, log)
+            raise KubeException(log)
```
bacongobbler: we probably don't want to raise an error when any normal event in the namespace occurs. Can you check specifically for events that fail a quota check instead?

rvadim: > Can you check specifically for events that fail a quota check instead?
It is my first plan, but it is very hard to detect quota check events, because events about quota not linked with any pods.

vdice:
@rvadim thank you for the contribution!  My only comment would be to request unit tests be added around these additions to make codecov happy.

rvadim:
- @bacongobbler Commit message changed
- @vdice  Test added.
Please notify me, if need some more changes.

Thank you.

rootfs/scheduler/resources/pod.py
bacongobbler: I still think we need to tackle this. WDYT @mboersma @vdice?

vdice: I'm in agreement with @bacongobbler and would push for finding a way to check events specific to quota check fails, as mentioned.

rvadim: As described in kubernetes controller:
https://github.com/kubernetes/kubernetes/blob/f18a921a0308c4ecd2e7e8b7e30e39a8216a3447/pkg/controller/deployment/util/deployment_util.go#L612

To get new RS need to get template by deployment and match RS to deployment template. And i have no idea, how to get last RS for specified deployment in deis controller. So, we can't detect object which cause error and what error.

And I actually don't understand why we need to return None when required pods != actual pods during operation.

I can add regexp to detect patterns with "exceeded quota", but this message may changes in next releases.

bacongobbler: > And i have no idea, how to get last RS for specified deployment in deis controller.

[The unit tests](https://github.com/deis/controller/blob/890a26318802fedf1ce82c42ad2401af57671af4/rootfs/scheduler/tests/test_deployments.py#L188) demonstrate how one can get a replicaset deployed by the controller.

> And I actually don't understand why we need to return None when required pods != actual pods during operation.

I'm sorry, I don't understand what you mean by this. Is there something in the code that needs clarifying, and if so where? The only added condition in this PR where I see we return `None` is when no events have occurred in the namespace on line 735.

> I can add regexp to detect patterns with "exceeded quota", but this message may changes in next releases.

If you're checking the replicaset itself, you can probably just check it for a condition called ["ReplicaFailure"](https://github.com/kubernetes/client-go/blob/8fa0506b2636f5a0324d0cd7783efacc5edb2464/pkg/apis/extensions/v1beta1/types.go#L890-L893) instead, which is a catch-all for quotas, limits, security policies, node selectors etc. That should probably live in [`deployments.are_replicas_ready()`](https://github.com/deis/controller/blob/890a26318802fedf1ce82c42ad2401af57671af4/rootfs/scheduler/resources/deployment.py#L242), though.

rvadim: @bacongobbler Thank you. It's really works. I rewrite some checks and test it on our environment, works well.

rvadim:
I can't reproduce failures, on my local env all tests passed. So, can anyone restart CI or help with reproducing?

vdice:
@rvadim it looks like the Jenkins and Travis CI runs both passed; it's actually codecov complaining -- it appears there is still code added here that isn't yet covered under test.  Could you take another look when convenient?

rootfs/scheduler/resources/deployment.py
```diff
@@ -237,6 +237,12 @@ def in_progress(self, namespace, name, timeout, batches, replicas, tags):
             self.log(namespace, 'Deploy operation for Deployment {} in has expired. Rolling back to last good known release'.format(name), level='DEBUG')  # noqa
             return False, True
 
+        try:
+            self._check_for_failed_events(namespace, labels=labels)
```
bacongobbler: how come we need to check for failed events when determining that the deploy is in progress as well as when the replicas are ready?

vdice:
@rvadim don't worry about the codecov failure.  Meanwhile, it appears there's just one more comment needing to be addressed: https://github.com/deis/controller/pull/1182#discussion_r98486535

rootfs/scheduler/resources/deployment.py
rvadim: If we not checked failures in RS here, use will be get in progress status of build, during default of custom deploy timeout, fo example:
```
deis scale cmd=4
error: Quota exceeded
deis scale cmd=3
Another build in progress
deis scale cmd=3
Another build in progress

vdice:
LGTM -- I'll let @bacongobbler address any remaining questions, etc. (https://github.com/deis/controller/pull/1182#discussion_r98486535)

bacongobbler:
Everything looks good to me now. :)

rvadim:
Thank you!  :tada:

PR #1195, opened 2017-01-04
chore(requirements): update Django to 1.10.5

mboersma:
See https://docs.djangoproject.com/en/1.10/releases/1.10.5/


mboersma:
This PR actually passed, but VersionEye didn't know about the Django update yet and flagged it as "No License," which has since been reconciled.

PR #1196, opened 2017-01-11
Get timezone from environment variable

ChillarAnand:
closes https://github.com/deis/deis/issues/5113

### DIFF
--- a/charts/controller/templates/controller-deployment.yaml
+++ b/charts/controller/templates/controller-deployment.yaml
@@ -75,6 +75,8 @@ spec:
                   key: image
             - name: "IMAGE_PULL_POLICY"
               value: "{{ .Values.app_pull_policy }}"
+            - name: "TZ"
+              value: "{{ default 'UTC' .Values.time_zone }}"
 {{- if (.Values.deploy_hook_urls) }}
             - name: DEIS_DEPLOY_HOOK_URLS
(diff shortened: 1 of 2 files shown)


deis-admin:

bacongobbler:
If you wouldn't mind amending your commit such that it follows the [commit style guide](https://deis.com/docs/workflow/contributing/submitting-a-pull-request/#commit-style), that would be awesome! That way it'll get picked up by `deisrel` when we generate the release changelog. Thanks!

ChillarAnand:
Amended commit. Thanks @bacongobbler

mboersma:
Thanks @ChillarAnand! Do we think this change is sufficient, or should we also add this to [the deployment](https://github.com/deis/controller/blob/master/charts/controller/templates/controller-deployment.yaml#L55) and the `values.yaml`:

```yaml
- name: "TZ"
  value: "{{ .Values.time_zone }}"
```

bacongobbler:
Good point @mboersma. @ChillarAnand would you mind making that change so users can modify the timezone from the chart?

the last line probably needs to be

```
value: "{{ default "UTC" .Values.time_zone }}"
```

So that existing charts that don't set the time zone will still work.

ChillarAnand:
Updated. Thanks @mboersma @bacongobbler

PR #1201, opened 2017-01-12
fix(models/app): recreate proc types on switch from Dockerfile to buildpack

mboersma:
Or vice versa.

Closes deis/workflow#658.

~~This seems to fix things, but I need to write some unit tests to cover these cases, possibly write some e2e tests that cover this as well, and reassure myself that this doesn't create any regressions.~~ *Edit*: finally got back to this PR and after some refactoring, this change fixes this edge case.

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -449,13 +449,28 @@ class App(UuidAuditedModel):
         # use create to make sure minimum resources are created
         self.create()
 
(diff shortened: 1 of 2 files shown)


mboersma:
Improved and ready for review now. If you want to test this manually (not required), use [example-dockerfile-python](https://github.com/deis/example-dockerfile-python) and alternate between branches with `git push deis master` with `git push deis dockerfile`.

PR #1205, opened 2017-01-18
fix(Makefile): build docker image with --no-cache by default

mboersma:
We've been bitten in CI by Docker layers inadvertently being cached. See https://github.com/deis/postgres/pull/180#issuecomment-273562141 for an example.

Developers can build like this for faster iterative builds:
```
DOCKER_BUILD_FLAGS= make docker-build
```


mboersma:
Should I remove the `--rm` flag since it already defaults to `true`?

bacongobbler:
sgtm

mboersma:
If this approach is acceptable, I'll make similar changes to the other projects where this is relevant.

PR #1230, opened 2017-01-20
Update README for correct python version.

robholland:


mboersma:
Thanks for doing this! Could you amend the commit message to conform with our [style guide](https://deis.com/docs/workflow/contributing/submitting-a-pull-request/#commit-style)? That way the CHANGELOG generator will include it properly. Something like:
```
docs(README): specify correct python version
```

robholland:
Updated.

PR #1235, opened 2017-01-25
chore(requirements): update requests to 2.13.0

mboersma:
From the requests library release notes:

Features
- Only load the idna library when we've determined we need it. This will save some memory for users.

Miscellaneous
- Updated bundled urllib3 to 1.20.
- Updated bundled idna to 2.2.


mboersma:
This actually breaks a regex-based unit test, so I'll work that out and then resubmit it for review.

mboersma:
My rewriting of `import` statements somehow had caused the unit test failure, so I reverted that. Works fine now: ready for review again.

PR #1236, opened 2017-01-27
fix(pods.py): use init container to download slug

mboersma:
Replaces #1179.


bacongobbler:
it looks like this has gone green but is out of date and is still marked as WIP. Is this still true or is it ready for review? If so would you mind rebasing this?

mboersma:
It's still WIP overall unfortunately as it requires other related changes. I'll retarget it for the next release.

PR #1239, opened 2017-01-31
fix(charts): reference registry-proxy on 127.0.0.1

bacongobbler:
On some providers, "localhost" is not set up on the host network. explicitly calling the loopback
address fixes this.

See https://github.com/deis/workflow/issues/678


vdice:
Tested by editing controller and builder deployments with this change.  Deploying a Dockerfile-based app (and using local registry) now works.

PR #1243, opened 2017-02-09
feat(ingress) Experimental Native Ingress

kris-nova:
Work for experimental native ingress. Run the documentation in https://github.com/deis/workflow/pull/732 to see how to use the new feature.

This PR handles the changes necessary to the controller to start managing Kubernetes ingresses

Requires deis/workflow#732
Requires deis/builder#495
Requires deis/router#316

### DIFF
--- a/charts/controller/templates/controller-deployment.yaml
+++ b/charts/controller/templates/controller-deployment.yaml
@@ -58,6 +58,11 @@ spec:
             # NOTE(bacongobbler): use deis/registry_proxy to work around Docker --insecure-registry requirements
             - name: "DEIS_REGISTRY_SERVICE_HOST"
               value: "127.0.0.1"
+            # Environmental variable value for $EXPERIMENTAL_NATIVE_INGRESS
+            - name: "EXPERIMENTAL_NATIVE_INGRESS"
+              value: "{{ .Values.global.experimental_native_ingress }}"
+            - name: "EXPERIMENTAL_NATIVE_INGRESS_HOSTNAME"
+              value: "{{ .Values.platform_domain }}"
             - name: "K8S_API_VERIFY_TLS"
               value: "{{ .Values.k8s_api_verify_tls }}"
             - name: "DEIS_REGISTRY_SERVICE_PORT"
--- a/charts/controller/values.yaml
+++ b/charts/controller/values.yaml
@@ -14,6 +14,10 @@ deploy_hook_urls: ""
 registration_mode: "admin_only"
 # Option to disable ssl verification to connect to k8s api server
 k8s_api_verify_tls: "true"
+# The public resolvable hostname to build your cluster with.
+#
+# This will be the hostname that is used to build endpoints such as "deis.$HOSTNAME"
+platform_domain: ""
 
 global:
   # Set the storage backend
@@ -45,3 +49,9 @@ global:
   host_port: 5555
   # Prefix for the imagepull secret created when using private registry
   secret_prefix: "private-registry"
+  # Experimental feature to toggle using kubernetes ingress instead of the Deis router.
+  #
+  # Valid values are:
+  # - true: The deis controller will now create Kubernetes ingress rules for each app, and ingress rules will automatically be created for the controller itself.
+  # - false: The default mode, and the default behavior of Deis workflow.
+  experimental_native_ingress: false
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -191,6 +191,7 @@ class App(UuidAuditedModel):
 
         # create required minimum resources in k8s for the application
         namespace = self.id
+        ingress = self.id
         service = self.id
         quota_name = '{}-quota'.format(self.id)
         try:
@@ -200,7 +201,6 @@ class App(UuidAuditedModel):
                 self._scheduler.ns.get(namespace)
             except KubeException:
                 self._scheduler.ns.create(namespace)
-
             if settings.KUBERNETES_NAMESPACE_DEFAULT_QUOTA_SPEC != '':
                 quota_spec = json.loads(settings.KUBERNETES_NAMESPACE_DEFAULT_QUOTA_SPEC)
                 self.log('creating Quota {} for namespace {}'.format(quota_name, namespace),
@@ -224,6 +224,18 @@ class App(UuidAuditedModel):
 
             raise ServiceUnavailable('Kubernetes resources could not be created') from e
 
+        try:
+            # In order to create an ingress, we must first have a namespace.
+            if settings.EXPERIMENTAL_NATIVE_INGRESS:
+                try:
+                    self._scheduler.ingress.get(ingress)
+                except KubeException:
+                    self.log("creating Ingress {}".format(namespace), level=logging.INFO)
+                    self._scheduler.ingress.create(ingress,
+                                                   namespace,
+                                                   settings.EXPERIMENTAL_NATIVE_INGRESS_HOSTNAME)
+        except KubeException as e:
+            raise ServiceUnavailable('Could not create Ingress in Kubernetes') from e
         try:
             self.appsettings_set.latest()
         except AppSettings.DoesNotExist:
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -227,6 +227,8 @@ class App(UuidAuditedModel):
         try:
             # In order to create an ingress, we must first have a namespace.
             if settings.EXPERIMENTAL_NATIVE_INGRESS:
+                if ingress == "":
+                    raise ServiceUnavailable('Empty hostname')
                 try:
                     self._scheduler.ingress.get(ingress)
                 except KubeException:
--- a/rootfs/api/settings/production.py
+++ b/rootfs/api/settings/production.py
@@ -260,6 +260,10 @@ random_secret = 'CHANGEME_sapm$s%upvsw5l_zuy_&29rkywd^78ff(qi*#@&*^'
 SECRET_KEY = os.environ.get('DEIS_SECRET_KEY', random_secret)
 BUILDER_KEY = os.environ.get('DEIS_BUILDER_KEY', random_secret)
 
+# experimental native ingress
+EXPERIMENTAL_NATIVE_INGRESS = bool(os.environ.get('EXPERIMENTAL_NATIVE_INGRESS', 0))
+EXPERIMENTAL_NATIVE_INGRESS_HOSTNAME = os.environ.get('EXPERIMENTAL_NATIVE_INGRESS_HOSTNAME', '')
+
 # k8s image policies
 SLUGRUNNER_IMAGE = os.environ.get('SLUGRUNNER_IMAGE_NAME', 'quay.io/deisci/slugrunner:canary')  # noqa
 IMAGE_PULL_POLICY = os.environ.get('IMAGE_PULL_POLICY', "IfNotPresent")  # noqa
--- a/rootfs/scheduler/resources/ingress.py
+++ b/rootfs/scheduler/resources/ingress.py
@@ -1,0 +1,66 @@
+from scheduler.exceptions import KubeHTTPException
+from scheduler.resources import Resource
+
+
+class Ingress(Resource):
+    short_name = 'ingress'
+
+    def get(self, name=None, **kwargs):
+        """
+        Fetch a single Ingress or a list of Ingresses
+        """
+        if name is not None:
+            url = "/apis/extensions/v1beta1/namespaces/%s/ingresses/%s" % (name, name)
+            message = 'get Ingress ' + name
+        else:
+            url = "/apis/extensions/v1beta1/namespaces/%s/ingresses" % name
+            message = 'get Ingresses'
+
+        response = self.http_get(url, params=self.query_params(**kwargs))
+        if self.unhealthy(response.status_code):
+            raise KubeHTTPException(response, message)
+
+        return response
+
+    def create(self, ingress, namespace, hostname):
+        url = "/apis/extensions/v1beta1/namespaces/%s/ingresses" % namespace
+
+        if hostname == "":
+            raise KubeHTTPException("empty hostname value")
+
+        data = {
+            "kind": "Ingress",
+            "apiVersion": "extensions/v1beta1",
+            "metadata": {
+                "name": ingress
+            },
+            "spec": {
+                "rules": [
+                    {"host": ingress + "." + hostname,
+                     "http": {
+                         "paths": [
+                             {"path": "/",
+                              "backend": {
+                                  "serviceName": ingress,
+                                  "servicePort": 80
+                              }}
+                         ]
+                     }
+                     }
+                ]
+            }
+        }
+        response = self.http_post(url, json=data)
+
+        if not response.status_code == 201:
+            raise KubeHTTPException(response, "create Ingress {}".format(namespace))
+
+        return response
+
+    def delete(self, namespace, ingress):
+        url = self.api("/namespaces/{}/ingresses/{}", namespace, ingress)
+        response = self.http_delete(url)
+        if self.unhealthy(response.status_code):
+            raise KubeHTTPException(response, 'delete Ingress "{}"', namespace)
+
+        return response
--- a/rootfs/scheduler/resources/ingress.py
+++ b/rootfs/scheduler/resources/ingress.py
@@ -25,9 +25,6 @@ class Ingress(Resource):
     def create(self, ingress, namespace, hostname):
         url = "/apis/extensions/v1beta1/namespaces/%s/ingresses" % namespace
 
-        if hostname == "":
-            raise KubeHTTPException("empty hostname value")
-
         data = {
             "kind": "Ingress",
             "apiVersion": "extensions/v1beta1",
--- a/charts/controller/templates/controller-ingress-rule-http-80.yaml
+++ b/charts/controller/templates/controller-ingress-rule-http-80.yaml
@@ -1,0 +1,21 @@
+{{ if .Values.global.experimental_native_ingress }}
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  namespace: "deis"
+  name: "controller-api-server-ingress-http"
+  labels:
+    app: "controller"
+    chart: "{{ .Chart.Name }}-{{ .Chart.Version }}"
+    release: "{{ .Release.Name }}"
+    heritage: "{{ .Release.Service }}"
+spec:
+  rules:
+  - host: deis.{{ .Values.platform_domain }}
+    http:
+      paths:
+      - path: /
+        backend:
+          serviceName: deis-controller
+          servicePort: 80
+{{- end }}


README.md
```diff
@@ -21,7 +21,7 @@ The Controller is the central API server for [Deis Workflow][workflow]. It is in
 * Configure an application
 * Create a new user
 
-# Development
+# Contributing
```
mboersma: These clarifications to the docs are great, but should be in a different PR ideally since they're not related to ingress changes.

.gitignore
```diff
@@ -13,6 +13,7 @@ develop-eggs
 .installed.cfg
 lib
 lib64
+.idea*
```
krancour: As @mboersma has noted on other PRs, I would suggest this be a separate PR, just to keep is tractable.

charts/controller/templates/controller-ingress-rule-controller.yaml
```diff
@@ -0,0 +1,21 @@
+{{ if .Values.global.experimental_native_ingress }}
```
krancour: I'm almost certain that the "parent" chart's global configuration isn't visible to this chart. i.e. There's no "looking up." Remember also that each Workflow component's own chart is installable independent of the "parent" Workflow chart and all its sibling components. With that in mind, you can see why "looking up" isn't a thing.

krancour: I've made a similar comment on some other PRs related to this same feature.

I think it's clear to me that we're going to have to explore some more creative patterns for how, in the "parent" chart, one flag can be set that enables this feature and have that setting can _propagate_ down to dependent charts like the router chart and controller chart. This might be easier said than done, but we're going to have to find a way.

kris-nova: @krancour So I think your assumption isn't entirely accurate. I was able to test this under the following situation:

1. Define `global. experimental_native_ingress` in the `values.yaml` for **workflow**
2. Neglect to define `global.experimental_native_ingress` in the `values.yaml` for **controller** 
3. Nest the two charts together on my local filesystem as they are hosted https://charts.deis.com/workflow
```
kris-nova:workflow kris$ tree
.
├── Chart.yaml
├── charts
│   ├── builder
│   │   ├── Chart.yaml
│   │   ├── templates
│   │   │   ├── builder-controller-secret-key-auth.yaml
│   │   │   ├── builder-deployment.yaml
│   │   │   ├── builder-secret-ssh-private-keys.yaml
│   │   │   ├── builder-service-account.yaml
│   │   │   └── builder-service.yaml
│   │   └── values.yaml
│   ├── controller
│   │   ├── Chart.yaml
│   │   ├── templates
│   │   │   ├── controller-deployment.yaml
│   │   │   ├── controller-ingress-rules.yaml
│   │   │   ├── controller-secret-django-secret-key.yaml
│   │   │   ├── controller-service-account.yaml
│   │   │   ├── controller-service.yaml
│   │   │   └── deploy-hook-secret.yaml
│   │   └── values.yaml
```

And run a helm install locally 

```
helm install . --namespace deis --set global.experimental_native_ingress=true,controller.platform_domain=fabulous.af
```

And a quick check of our pods will demonstrate that the flag **in fact was inherited from the parent** (Notice the lack of a router)

```
kris-nova:workflow kris$ kubectl get po -n deis
NAME                                     READY     STATUS    RESTARTS   AGE
deis-builder-542482751-h844h             0/1       Running   1          1m
deis-controller-2957360810-sj2ik         0/1       Running   0          1m
deis-database-3140037217-i7fm5           1/1       Running   0          1m
deis-logger-176328999-8k11f              1/1       Running   2          1m
deis-logger-fluentd-620pz                1/1       Running   0          1m
deis-logger-fluentd-jx2fi                1/1       Running   0          1m
deis-logger-fluentd-ormq9                1/1       Running   0          1m
deis-logger-redis-2637728276-5e5ul       1/1       Running   0          1m
deis-minio-1095953168-w30z5              1/1       Running   0          1m
deis-monitor-grafana-432496062-rz5n6     1/1       Running   0          1m
deis-monitor-influxdb-2729657543-l63s4   1/1       Running   0          1m
deis-monitor-telegraf-if94d              1/1       Running   0          1m
deis-monitor-telegraf-mv4ul              1/1       Running   0          1m
deis-monitor-telegraf-pwcid              1/1       Running   0          1m
deis-nsqd-3597503299-flbfk               1/1       Running   0          1m
deis-registry-1455273266-5293m           1/1       Running   3          1m
deis-registry-proxy-hkpp5                1/1       Running   0          1m
deis-registry-proxy-hsouy                1/1       Running   0          1m
deis-registry-proxy-s9aaa                1/1       Running   0          1m
deis-workflow-manager-2528409207-z7xoh   1/1       Running   0          1m
```


---

I think we **should** add the directive to **both** charts (Which is already in the latest commit), so that they can be installed independently of each other.. but even after trying a second test with the directive defined in **both** charts, I was able to demonstrate that the value was passed down, and the parent chart's value took precedence over the child chart.

---

> I'm almost certain that the "parent" chart's global configuration isn't visible to this chart

A quick demonstration was able to debunk this assumption, hence why I always prefer to test things before making assumptions. 

##### Helm Version

```
kris-nova:workflow kris$ helm version
Client: &version.Version{SemVer:"v2.1.3", GitCommit:"5cbc48fb305ca4bf68c26eb8d2a7eb363227e973", GitTreeState:"clean"}
Server: &version.Version{SemVer:"v2.1.3", GitCommit:"5cbc48fb305ca4bf68c26eb8d2a7eb363227e973", GitTreeState:"clean"}
```

sheerun: Indeed, global variables are documented here:
https://github.com/kubernetes/helm/blob/master/docs/charts.md#global-values

In short, with normal variables child charts see variables set by parent chart, but without the namespace. So for example `.Values.foo.bar` is seen as `.Values.bar` from `foo` subchart.

The `global` namespace is special, because it's not stripped when passing variables to child subcharts, also:

> If a subchart declares a global variable, that global will be passed downward (to the subchart's subcharts), but not upward to the parent chart. There is no way for a subchart to influence the values of the parent chart.

README.md
```diff
@@ -77,6 +77,14 @@ Running `python --version` to verify the correct version is recommend.
...
 Follow the linked documentation to learn about [virtualenv][virtualenv] and how to `activate` and `deactivate` the environment.
 
+## Developing with the controller
+
+Run the server in a container
+
+```bash
+$ run -p 127.0.0.1:8000:8000 -it <image_id> sudo -E -u deis gunicorn -c /app/deis/gunicorn/config.py api.wsgi
+```
+
```
krancour: Should additions such as this one be in a separate PR?

README.md
```diff
@@ -98,6 +106,8 @@ export IMAGE_PREFIX=arschles
 make deploy
 ```
 
+The `make deploy` target is sensitive to deltas in version control. In order for you to test local code, you will first need to `git add` and `git commit` the code.
```
krancour: Same question applies re: whether this should be in a separate PR. However... I have a solution for this (which I recently applied to router's `make deploy` that resolves this adequately such that it doesn't even need to be documented. Let's chat about it later.

charts/controller/templates/controller-deployment.yaml
```diff
@@ -24,7 +24,7 @@ spec:
       serviceAccount: deis-controller
       containers:
         - name: deis-controller
-          image: quay.io/{{.Values.org}}/controller:{{.Values.docker_tag}}
+          image: quay.io/knova/controller-dev:canary
```
krancour: I assume this change was meant to be undone?

charts/controller/templates/controller-deployment.yaml
```diff
@@ -58,6 +58,11 @@ spec:
             # NOTE(bacongobbler): use deis/registry_proxy to work around Docker --insecure-registry requirements
             - name: "DEIS_REGISTRY_SERVICE_HOST"
               value: "127.0.0.1"
+            # Environmental variable value for $EXPERIMENTAL_NATIVE_INGRESS
+            - name: "EXPERIMENTAL_NATIVE_INGRESS"
+              value: "{{ .Values.global.experimental_native_ingress }}"
+            - name: "EXPERIMENTAL_NATIVE_INGRESS_HOSTNAME"
+              value: "{{ .Values.global.hostname }}"
```
krancour: I think that upstream in the Workflow chart, you changed this to `controller.platform_domain`, so here, it should just be `{{ .Values.platform_domain }}`

charts/controller/templates/controller-ingress-rule-builder-2222.yaml
```diff
@@ -0,0 +1,21 @@
...
+spec:
+  rules:
+  - host: deis.{{ .Values.platform_domain }}
+    http:
+      paths:
+      - path: /
+        backend:
+          serviceName: deis-controller
+          servicePort: 2222
+{{- end }}
```
krancour: Does this ingress rule end up having the intended effect? afaik, the builder speaks ssh, meaning the traffic flowing to it is raw TCP... whereas ingress _only_ works for HTTP/S. 😢 (How I wish that weren't so.)

I _think_ that when experimental native ingress is enabled, we have to make builder's service one of `type: LoadBalancer` to compensate for the fact that deis-router won't be around handling the non-HTTP builder traffic as a special case.

charts/controller/templates/controller-ingress-rule-healthz-9090.yaml
```diff
@@ -0,0 +1,21 @@
...
+spec:
+  rules:
+  - host: deis.{{ .Values.platform_domain }}
+    http:
+      paths:
+      - path: /
+        backend:
+          serviceName: deis-controller
+          servicePort: 9090
+{{- end }}
```
krancour: Not sure I understand this rule. afaik, the only thing that (optionally) uses 9090 is the router.

(TMI: Router is set up to guarantee PROXY protocol is never used on 9090, _regardless_ of whether it's used on 80 and 443. This is to allow load balancers to perform HTTP-based healthchecks. Healthchecks, in an ELB at least, don't support PROXY protocol.)

Back to the matter at hand... I could be wrong, but I am pretty sure controller's container doesn't listen on 9090 at all.

charts/controller/values.yaml
```diff
@@ -45,3 +49,9 @@ global:
   host_port: 5555
   # Prefix for the imagepull secret created when using private registry
   secret_prefix: "private-registry"
+  # Experimental feature to toggle using kubernetes ingress instead of the Deis router.
+  #
+  # Valid values are:
+  # - true: The Deis router will NOT be deployed. Inherently workflow will not be usable until a Kubernetes ingress controller is installed.
```
krancour: The controller chart would never install the router anyway. Maybe this sentence can be reworded to articulate controller's behavior differences when this is set-- i.e. it will create ingress resources for each app _in addition to_ the usual "routable services."

rootfs/api/models/app.py
```diff
@@ -240,6 +263,7 @@ def delete(self, *args, **kwargs):
 
             try:
                 self._scheduler.ns.delete(self.id)
+                self._scheduler.ingress.delete(self.id, self.id)
```
krancour: Since the line above deletes the application's namespace, is this line redundant?

rootfs/api/models/app.py
```diff
@@ -657,7 +681,7 @@ def verify_application_health(self, **kwargs):
             time.sleep(1)
 
         # Endpoint did not report healthy in time
-        if ('response' in locals() and response.status_code == 404) or failed:
+        if (('response' in locals() and response.status_code == 404) or failed) and not is_experimental_native_ingress():
```
krancour: Ha. This is funny. This would wait for the app to become available, right? But if experimental ingress is enabled, the possibility exists that no ingress controller is installed...

@mboersma should make the call on this, but I think it might be better to leave this the way it was and instead make the documentation crystal clear that if you want to BYO ingress controller, you need to have that installed _before_ you can rightfully expect Workflow to start working.

I'm just worried about what happens if we totally give up this idea of waiting for the app to be up and responding.

krancour:
I think something that is missing here is that when custom / vanity domains are either added or removed from an application, the corresponding ingress resources also need to be added or removed.

rootfs/api/models/app.py
mboersma: > make the documentation crystal clear that if you want to BYO ingress controller, you need to have that installed before you can rightfully expect Workflow to start working.

I think that's a reasonable approach for an experimental feature. We can change it if this is tripping users up.

charts/controller/templates/controller-ingress-rule-healthz-9090.yaml
mboersma: >  I am pretty sure controller's service doesn't listen on 9090 at all.

@krancour is correct AFAIK.

charts/controller/templates/controller-ingress-rule-builder-2222.yaml
kris-nova: I took the rule form here: https://github.com/deis/router/blob/756f31e19b9e2c2d31a002e7d457c5120f65cf06/charts/router/templates/router-service.yaml#L24

Is that not correct?

charts/controller/templates/controller-ingress-rule-healthz-9090.yaml
kris-nova: I took the rule from here https://github.com/deis/router/blob/756f31e19b9e2c2d31a002e7d457c5120f65cf06/charts/router/templates/router-service.yaml#L27

Is that not correct?

charts/controller/templates/controller-ingress-rule-builder-2222.yaml
mboersma: > we have to make builder's service one of type: `LoadBalancer` 

I think so too. This rule won't work since it needs raw TCP and ingress doesn't understand that (and my testing shows it's not routing to builder, although controller API does work). So another PR in deis/builder is probably required for everything to work.

charts/controller/templates/controller-ingress-rule-healthz-9090.yaml
mboersma: It is correct that deis-router listens on 9090 for optional healthchecks, but I think for the experimental ingress feature we can just drop that behavior. @krancour agreed?

rootfs/api/models/app.py
```diff
@@ -33,6 +34,26 @@
 logger = logging.getLogger(__name__)
 
 
+def is_experimental_native_ingress():
+    """
+    Will determine the value of the experimental_native_ingress flag
+    """
+    is_ingress = os.getenv('EXPERIMENTAL_NATIVE_INGRESS', False)
```
mboersma: This function could be simplified:
```python
return bool(os.getenv('EXPERIMENTAL_NATIVE_INGRESS'))
```

rootfs/api/models/app.py
```diff
@@ -33,6 +34,26 @@
...
+    if not is_ingress:
+        return False
+    return True
+
+
+def get_experimental_native_ingress_hostname():
+    """
+    Will return the experimental native ingress platform_domain value, or an empty string
```
mboersma: This docstring doesn't match the code. Instead maybe:
"Returns the experimental native ingress hostname or raises ServiceUnavailable if it's not set."

rootfs/api/models/app.py
```diff
@@ -33,6 +34,26 @@
...
+    """
+    host = os.getenv('EXPERIMENTAL_NATIVE_INGRESS_HOSTNAME', "")
```
mboersma: It would be a little simpler to test for truthy/falsey rather than an empty string explicitly IMHO:
```python
host = os.getenv('EXPERIMENTAL_NATIVE_INGRESS_HOSTNAME')
if not host:
    raise ServiceUnavailable("Unable to find hostname")
return host
```

mboersma:
I wasn't able to do a `git push deis master` (see above) but `deis pull` worked fine. Looks like the appropriate ingress rule was created in the app namespace and Traefik is successfully exposing it. That is pretty boss. 💥

kris-nova:
@krancour 

> I think something that is missing here is that when custom / vanity domains are either added or removed from an application, the corresponding ingress resources also need to be added or removed.

Is that a requirement for the core feature here? Or can we try to get that in in a follow up PR after the feature goes out? 

@mboersma 

Can I steal some of your time to talk about this builder PR you think we need?

kris-nova:
I opened up https://github.com/deis/builder/pull/495 with the builder changes

rootfs/api/models/app.py
```diff
@@ -33,6 +34,23 @@
...
 
 
+def is_experimental_native_ingress():
+    """
+    Will determine the value of the experimental_native_ingress flag
+    """
+    return bool(os.getenv('EXPERIMENTAL_NATIVE_INGRESS'))
+
+
+def get_experimental_native_ingress_hostname():
```
bacongobbler: same here, see above comment about EXPERIMENTAL_NATIVE_INGRESS

rootfs/api/models/app.py
```diff
@@ -33,6 +34,23 @@
 logger = logging.getLogger(__name__)
 
 
+def is_experimental_native_ingress():
```
bacongobbler: when we can, we should be using `settings.EXPERIMENTAL_NATIVE_INGRESS` instead as seen in some other places. We can still make it [come from an environment variable](https://github.com/deis/controller/blob/a8002a2172bc37801fc062f9a8a9fb65e50f4be3/rootfs/api/settings/production.py#L260), but Django has better support in general for settings than os envvars. The `is_experimental_native_ingress()` function could be removed entirely in lieu of `DEIS_EXPERIMENTAL_NATIVE_INGRESS = bool(os.getenv('EXPERIMENTAL_NATIVE_INGRESS', 0))` in production.py, for example.

Example usage:

https://github.com/deis/controller/blob/a8002a2172bc37801fc062f9a8a9fb65e50f4be3/rootfs/api/settings/production.py#L147
https://github.com/deis/controller/blob/a8002a2172bc37801fc062f9a8a9fb65e50f4be3/rootfs/api/models/app.py#L783
https://github.com/deis/controller/blob/a8002a2172bc37801fc062f9a8a9fb65e50f4be3/rootfs/api/tests/test_app.py#L657-L658

mboersma: > we should be using settings.DEIS_EXPERIMENTAL_NATIVE_INGRESS instead

Good point. That would be more consistent with the way other flags are handled.

mboersma:
And 💯  for unit tests being added here. Controller has a high level of coverage and we aim for each PR including test coverage. I'm happy to help with those as well.

charts/controller/templates/controller-ingress-rule-builder-2222.yaml
mboersma: The `LoadBalancer` in deis-builder does seem to do the trick, so this ingress rule should be removed.

charts/controller/templates/controller-ingress-rule-healthz-9090.yaml
mboersma: @kris-nova let's drop this ingress rule too, especially since we're planning on just getting the core (experimental) functionality out there and then filling in the gaps we find.

krancour:
@mboersma 

> (Another issue I see is that the git remote entries created by deis create could be wrong, since the builder gets its own ELB when this feature is enabled. I'll test that next.)

Because of GCP (long story that I won't go into) we had actually anticipated that people might have dedicated load balancers for the builder. As such, we made the git remotes start with `deis-builder.<platform domain>` instead of the old `deis.<platform domain>`. That allowed the possibility of a separate record in DNS directing `deis-builder.<platform domain>` to the builder's load balancer whilst `*.<platform domain>` (which includes the `deis`/controller subdomain) gets router to the other load balancer.

Long story short... this was anticipated and there shouldn't be too much trouble making this work.

kris-nova:
Thanks for the tip on settings @bacongobbler, your concerns have been addressed. 

Working on unit tests now, will ping you or @mboersma if I need anything. Thanks again

rootfs/api/models/app.py
```diff
@@ -225,6 +225,18 @@ def create(self, *args, **kwargs):  # noqa
             raise ServiceUnavailable('Kubernetes resources could not be created') from e
 
         try:
+            # In order to create an ingress, we must first have a namespace.
+            if settings.EXPERIMENTAL_NATIVE_INGRESS():
```
bacongobbler: since `settings.EXPERIMENTAL_NATIVE_INGRESS` is a variable this will break with the trailing parentheses.

rootfs/api/settings/production.py
```diff
@@ -260,6 +260,10 @@
 SECRET_KEY = os.environ.get('DEIS_SECRET_KEY', random_secret)
 BUILDER_KEY = os.environ.get('DEIS_BUILDER_KEY', random_secret)
 
+# experimental native ingress
+EXPERIMENTAL_NATIVE_INGRESS = os.environ.get('EXPERIMENTAL_NATIVE_INGRESS', 0)
```
bacongobbler: we will need to convert this to a bool in order for this to work, like so

```
EXPERIMENTAL_NATIVE_INGRESS = bool(os.environ.get('EXPERIMENTAL_NATIVE_INGRESS', 0))
```

os.environ.get('EXPERIMENTAL_NATIVE_INGRESS', 0) will be of type string or int otherwise.

rootfs/scheduler/resources/ingress.py
```diff
@@ -0,0 +1,63 @@
...
+
+        data = {
+            "kind": "Ingress",
+            "apiVersion": "extensions/v1beta1",
+            "metadata": {
+                "name": ingress
+            },
+            "spec": {
+                "rules": [
+                    {"host": ingress + "." + hostname,
```
bacongobbler: should we error if `hostname` is unset? For example if the user enables ingress via EXPERIMENTAL_NATIVE_INGRESS but doesn't set a hostname for EXPERIMENTAL_NATIVE_INGRESS_HOSTNAME, this field will end up being `"host": "mynamespace."`.

bacongobbler:
Great work by the way. I'm loving how this is all shaping out. :)

kris-nova:
@mboersma @bacongobbler 

On the note of unit tests here are my concerns after looking at our dev cycle.

1. We have no way of running the unit test suite other than pushing a change out to a cluster. The docker image in the Dockerfile itself doesn't even ship with the required executables to run the tests :(
2. Even after taking a vanilla Controller image into a cluster and jumping on a shell. The tests are STILL failing with 

```bash
  File "/app/manage.py", line 12, in <module>
    execute_from_command_line(sys.argv)
  File "/usr/local/lib/python3.5/dist-packages/django/core/management/__init__.py", line 367, in execute_from_command_line
    utility.execute()
  File "/usr/local/lib/python3.5/dist-packages/django/core/management/__init__.py", line 359, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/usr/local/lib/python3.5/dist-packages/django/core/management/commands/test.py", line 29, in run_from_argv
    super(Command, self).run_from_argv(argv)
  File "/usr/local/lib/python3.5/dist-packages/django/core/management/base.py", line 286, in run_from_argv
    parser = self.create_parser(argv[0], argv[1])
  File "/usr/local/lib/python3.5/dist-packages/django/core/management/base.py", line 260, in create_parser
    self.add_arguments(parser)
  File "/usr/local/lib/python3.5/dist-packages/django/core/management/commands/test.py", line 56, in add_arguments
    test_runner_class = get_runner(settings, self.test_runner)
  File "/usr/local/lib/python3.5/dist-packages/django/test/utils.py", line 146, in get_runner
    test_module = __import__(test_module_name, {}, {}, force_str(test_path[-1]))
ImportError: No module named 'api.tests'
```


So my proposal here is that I go ahead and fix the concerns, but do it in another PR. Right now the Controller documentation is pretty vague, and requires the user to install a lot of noisy things on their workstation to run the tool. Frankly this seems wrong, and I am not to keen on running apps like this directly on my workstation. 

*I would like to dockerize the app so we can run tests easily*

But that seems like a different PR.

Option 1) Hold the experimental ingress things on the new test PR
Option 2) Merge this in without tests, and hopefully get the test PR in by next release.


 Thoughts?

bacongobbler:
We actually just run all the unit tests locally since they just rely on postgres, but I can understand the point on it being a cliff to get set up and that it requires installing tools on local workstations. We've been pretty spoiled with Docker packaging all the things. 📦

For what it's worth, here's my local setup:

```
><> brew list | grep postgresql
postgresql
><> python3 --version
Python 3.5.2
><> make setup-venv
python3 -m venv venv
venv/bin/pip3 install --disable-pip-version-check -q -r rootfs/requirements.txt -r rootfs/dev_requirements.txt
><> make test-unit
```

I'm probably missing a few things, but that *should* be all that's required to get unit tests running locally.

I'll leave it to @mboersma to make the call, but I'm happy with deferring tests if we're in a rush to get this in.

mboersma:
Part of the "Submitting a Pull Request" doc is "Include Tests" (and docs). So the goal is for any PR to document and test itself before it gets merged.

We can of course be flexible, and merge this with a promise to backfill the unit tests immediately afterward. I just worry that could be derailed or delayed--you never know what might come up next week. And generally we won't merge anything that's going to drop controller's unit test coverage noticeably, which this PR probably will without its own unit tests.

So I would rather hold these PRs out until this one has unit tests. I am happy to help with that!

kris-nova:
@mboersma @bacongobbler 

It's not that I have anything about writing tests, I just honestly **can't run the tests** without putting a large amount of work in as it stands right now. So it just seems like getting the tests in general up to a more useable state is another effort, which naturally feels like another PR.

I am hard pressed to alter my workstation just to run a unit test suite. We really should have ways of running it in the cloud, or at least in a containerized environment. So my point here was mostly 

> Fix the unit testing problem in general before writing tests

and not

> I don't think we should write tests

Does that make sense? I guess I don't really know what I am asking here. Do I need a blessing to do this or can I just do this?

mboersma:
> We really should have ways of running it in the cloud, or at least in a containerized environment.

So true. We do `make test-unit` so often that we'd forgotten about what a pain it is to set up, and how it's not well documented. I'm going to resurrect #1049 to see if testing in a container is feasible, or at the very least write better docs about the current process. Then we can revisit writing some tests for this PR. 🛠

rootfs/scheduler/resources/ingress.py
kris-nova: I just added in a check for an empty string, and raise an exception if it's empty

mboersma:
This PR is currently failing because of some minor `flake8` style violations in Travis CI. I'm trying to get `make test-style` and `make test-unit` to run in a container, but don't want to block this PR any longer.

@kris-nova once this is passing, if I'm still working on the unit test refactor then perhaps we can merge this anyway and backfill the tests as the next task. I wouldn't usually recommend that, but this is an experimental feature and we want to get it out there for feedback soon.

bacongobbler:
Agreed. Considering that the dockerized controller stuff is a blocker for unit tests, is this good to review again?

bacongobbler:
@kris-nova sorry about that! The fix to e2e is in deis/e2e-runner#92.

kris-nova:
@bacongobbler sorry you will have to bear with me here. What do you mean fix? Was there something amiss in `e2e` recently? I am trying to make sense of these logs but it's a lot to take in right now. Also what does this mean for this PR? Do I need to wait until that is merged and try to build this again?

Basically I am just trying to get this PR merged in ASAP, and am wondering what the next steps are :woman_shrugging:

kris-nova:
@bacongobbler ohh just kidding - the jenkins build passed 🎉

bacongobbler:
I can help shed a bit of light on what's going on in e2e.

Basically, e2e-runner is the core component that runs all e2e jobs. It defines what happens during an e2e run. Before https://github.com/deis/e2e-runner/pull/90 was merged, we would [gather pod logs and delete the deis namespace](https://github.com/deis/e2e-runner/blob/master/scripts/lease.sh#L34-L42), purging the cluster and returning it to a "clean" slate for the next job.

Now, when https://github.com/deis/e2e-runner/pull/90 was merged, we accidentally removed that cleanup step, so what you saw here was an e2e run on a cluster that already ran through e2e but was never cleaned up, which caused the internal server errors. I'm assuming @vdice went ahead and cleaned up the k8s clusters' states which allowed your run to go green, but now the cluster is in a dirty state once again. It will likely need to be cleaned up once more after deis/e2e-runner#92 is merged so we can get everything back to normal.

Does that help clear things up?

rootfs/api/models/app.py
```diff
@@ -191,6 +191,7 @@ def create(self, *args, **kwargs):  # noqa
 
         # create required minimum resources in k8s for the application
         namespace = self.id
+        ingress = self.id
```
bacongobbler: Doesn't have to happen in this PR, but a neat python trick is to set these all to self.id in one shot:

```
namespace = ingress = service = self.id
```

As shown here:

```
><> python3
Python 3.5.2 (default, Aug  8 2016, 12:14:07) 
[GCC 4.9.2] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> a = b = c = 3
>>> a
3
>>> b
3
>>> c
3
>>>
```

rootfs/bin/test
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+python /app/manage.py test ingress
```
bacongobbler: I think this should be removed in favour of #1049. Is this something required for this PR or would you agree?

bacongobbler:
Indeed, looks like e2e failed once again. I think we can consider this OK though since we've already seen it go green once.

rootfs/api/models/app.py
kris-nova: I blame my time spent working in Scala, but I have never been a fan of syntactical sugar very much. I often find it introduces complexity, without changing the behavior of the program. Just my 2 cents 😄

mboersma:
Jenkins, test this please.

mboersma:
Replaced by #1274.

PR #1247, opened 2017-02-15
fix(app.py): use current namespace instead of "deis"

### LINKED ISSUE #1246: Secret not found for buildpack apps if Workflow is in a different namespace
After deis/builder#485, buildpack apps get farther if Workflow was installed elsewhere than the "deis" namespace. But they stall out here:
```shell
$ git push deis master
...
-----> Checking for changes inside the cache directory...
       Files inside cache folder changed, uploading new cache...
       Done: Uploaded cache (82M)
-----> Compiled slug size is 1.9M
Build complete.
Launching App...
remote: 2017/02/15 18:02:12 Error running git receive hook [The controller returned an error when publishing the release: Unknown Error (400): {"detail":"(app::deploy): ('failed to get Secret \"objectstorage-keyfile\" in Namespace \"deis\": 404 Not Found secrets \"objectstorage-keyfile\" not found', 'objectstorage-keyfile', 'deis')"}]
To ssh://deis-builder.192.168.99.100.nip.io:2222/easier-buttress.git
 ! [remote rejected] master -> master (pre-receive hook declined)
error: failed to push some refs to 'ssh://git@deis-builder.192.168.99.100.nip.io:2222/easier-buttress.git'
```

Refs deis/builder#483.

mboersma:
Uses the k8s downward api to make the current namespace available to controller so it isn't hard-coded as "deis."

Closes #1246.
Closes deis/charts#18.
Closes deis/builder#483.

### DIFF
--- a/charts/controller/templates/controller-deployment.yaml
+++ b/charts/controller/templates/controller-deployment.yaml
@@ -125,6 +125,10 @@ spec:
                   key: password
             - name: RESERVED_NAMES
               value: "deis, deis-builder, deis-workflow-manager, grafana"
+            - name: WORKFLOW_NAMESPACE
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.namespace
           volumeMounts:
             - mountPath: /var/run/docker.sock
               name: docker-socket
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -992,7 +992,8 @@ class App(UuidAuditedModel):
             username = registry.get('username')
             password = registry.get('password')
         elif settings.REGISTRY_LOCATION == 'off-cluster':
-            secret = self._scheduler.secret.get('deis', 'registry-secret').json()
+            secret = self._scheduler.secret.get(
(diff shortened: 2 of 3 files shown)


charts/controller/templates/controller-deployment.yaml
```diff
@@ -125,6 +125,10 @@ spec:
                   key: password
             - name: RESERVED_NAMES
               value: "deis, deis-builder, deis-workflow-manager, grafana"
+            - name: WORKFLOW_NAMESPACE
```
kris-nova: How does the environmental variable translate into the YAML file here? Is this a convention that we replace out later or something?

mboersma: It's k8s magic through the downward API--I just followed the example here: https://kubernetes.io/docs/user-guide/downward-api/

Kubernetes will ensure that `$WORKFLOW_NAMESPACE` is populated by the deployment's actual namespace when deployed since we reference `metadata.namespace`.

kris-nova:
LGTM - Thanks @mboersma!

PR #1249, opened 2017-03-06
feat(config.py): Add ability to setup default tags for applications

rvadim:
We have big enough k8s cluster with many nodes, some parts of nodes very specific and also have specific labels (another version of docker, perf test nodes, etc).
So in this situation all of our users need to set tag like "role=worker", it's not very good. And sometimes we detect applications w/o tags working on nodes where it should not.

docs: https://github.com/deis/workflow/pull/750


deis-admin:

mboersma:
Jenkins, add to whitelist.

rootfs/api/settings/testing.py
```diff
@@ -37,3 +38,5 @@
 # How long k8s waits for a pod to finish work after a SIGTERM before sending SIGKILL
 KUBERNETES_POD_TERMINATION_GRACE_PERIOD_SECONDS = int(os.environ.get('KUBERNETES_POD_TERMINATION_GRACE_PERIOD_SECONDS', 2))  # noqa
 KUBERNETES_NAMESPACE_DEFAULT_QUOTA_SPEC = '{"spec":{"hard":{"pods":"10"}}}'
+
+DEIS_DEFAULT_CONFIG_TAGS = str(os.environ.get('DEIS_DEFAULT_CONFIG_TAGS', ''))
```
vdice: I'm noticing how here the value is cast to string w/ `str()` but not in `rootfs/api/settings/production.py` -- is there a reason?

vdice:
Also, rebase needed it appears.

rootfs/api/settings/testing.py
rvadim: Thank you. Just forgot fix production.py

vdice:
Jenkins, test this please.

vdice:
Apologies, @rvadim -- master branch in controller was broken for a short amount of time -- a quick rebase is needed (should be trivial) and then we can kick off CI again.

rootfs/api/models/config.py
```diff
@@ -91,7 +92,14 @@ def set_registry(self):
     def set_tags(self, previous_config):
         """verify the tags exist on any nodes as labels"""
         if not self.tags:
-            return
+            if settings.DEIS_DEFAULT_CONFIG_TAGS != '':
```
mboersma: It's safer and more Pythonic to make this just a boolean comparison IMHO:
```python
if settings.DEIS_DEFAULT_CONFIG_TAGS:
```

rootfs/api/settings/testing.py
mboersma: We should remove the unnecessary `str()` cast in both cases. Python's `os.environ` will only return string values anyway.

rootfs/api/settings/production.py
```diff
@@ -309,6 +309,8 @@
 
 KUBERNETES_DEPLOYMENTS_REVISION_HISTORY_LIMIT = os.environ.get('KUBERNETES_DEPLOYMENTS_REVISION_HISTORY_LIMIT', None)  # noqa
 
+DEIS_DEFAULT_CONFIG_TAGS = str(os.environ.get('DEIS_DEFAULT_CONFIG_TAGS', ''))
```
mboersma: The `str()` cast here is not needed.

rvadim:
Rebase done, thank you for review.

PR #1264, opened 2017-03-13
fix(api): give `deis pull` apps with a cmd process type an entrypoint

bacongobbler:
closes deis/workflow#715


bacongobbler:
since we already make it clear that bash should be present in `deis pull` apps, this should not affect anyone.

bacongobbler:
On the contrary, this may not be the fix as this will override docker images that have an entrypoint defined. I think a separate fix is necessary here.

cristiangraz:
@bacongobbler Will this only apply to entries in the Procfile named `cmd`? Just want to confirm that using a Docker image deployed via deis pull with a Procfile like this would still work:

```
cmd: yarn start-service
worker1: ./worker --name worker1
worker2: ./worker --name worker2
```
Since the other process types are still using the same docker image, we'd need to take into account multiple commands in each process type, not just the `cmd` one:
```
cmd: ["yarn", "start-service"]
worker1: ["./worker", "--name", "worker1"]
worker2: ["./worker", "--name", "worker2"]
```

I apologize in advance if this is already the case, I'm not familiar with the Deis internals or python, but saw a reference to a dockerfile and `container_type == cmd` in the diff so just wanted to clarify.

Also will the `Dockerfile` need to be present in the `WORKDIR` of the image for this to work (I have it in `.dockerignore`)

bacongobbler:
Yes, the bug in question only applies to `cmd` process types. If you rename `cmd` to `web` it should work for you just fine.

bacongobbler:
k, so the new proposed fix should do the following:

1. respect the docker image's defined ENTRYPOINT
2. supply the arguments as a list to the manifest

With this, I can successfully deploy example-dockerfile-procfile-http as expected:

```
><> cat Procfile
cmd: python -m SimpleHTTPServer 5000
><> deis pull deis/example-dockerfile-python
```

PR #1269, opened 2017-03-16
feat(deploys): deploy releases asynchronously

kwent:
## Deploy releases asynchronously

Requires https://github.com/deis/workflow/pull/776
Requires https://github.com/deis/workflow-cli/pull/296
Requires https://github.com/deis/controller-sdk-go/pull/118

### Idea

This idea of this PR is too be able to create a build (which is creating a release) without deploying the release right away.

### Solution implemented

This is possible via a `deploy_now` flag on the follow endpoint:

```
POST /v2/apps/example-go/builds/ HTTP/1.1
Host: deis.example.com
Content-Type: application/json
Authorization: token abc123

{"image": "deis/example-go:latest", "deploy_now": false}
```

A new endpoint allows us to deploy a release later on:

```
POST /v2/apps/example-go/releases/deploy/ HTTP/1.1
Host: deis.example.com
Content-Type: application/json
Authorization: token abc123

{"version": 1}
```

### TODO

- [x] Update `workflow` API docs. https://github.com/deis/workflow/pull/776
- [x] Update `workflow-cli` to show `deploy_now` column of build objects. https://github.com/deis/workflow-cli/pull/296
- [x] Update `workflow-cli` to show `deployed` column of release objects. https://github.com/deis/workflow-cli/pull/296
- [x] Update `controller-sdk-go`. https://github.com/deis/controller-sdk-go/pull/118

### Other idea

Another idea would be to implement a `Deploy` model/object/table and keep track of the deploys with the following attributes `app`, `timestamp`, `release`, `who`, `status`. This will introduce a `deis deploys -a app` command as well.

2 scenarios would be the following:

-  **Scenario 1 (Default, Synchronous)**: When you push a build, it will always attempt to deploy and so create a deploy object with the release associated and the current status of the deployment.

- **Scenario 2 (Asynchronous)**: Otherwise in async mode just a build will be created and only a deploy object will be created if you call the endpoint to deploy the specific release.


deis-admin:

bacongobbler:
I'm kind of concerned about adding new fields to the request, but as long as older versions of the client can still interact with the controller then I'm fine with this. Our semver contract dictates that we maintain API backwards compatiblilty.

vdice:
Jenkins, test this please.

kwent:
@bacongobbler If you don't pass `deploy_now` attribute to the `POST /v2/apps/example-go/builds/` endpoint. It's defaulted to `true` by default. So not changing the default behavior and maintaining backwards compatibility.

vdice:
@kwent thanks for the PR! Will the additional code and workflow-cli PR(s) covering the TODOs mentioned in the description also be ready soon as well?  I've added this to the 2.13 milestone currently scheduled for April 5th.

kwent:
@vdice I did some work on it . (Updated description). But i might need some help. Not sure i covered everything that needs to be updated.

vdice:
@kwent Great, thanks.  I'll take a look at all of the PRs soon but in the meantime, when convenient, can you squash and rebase this into one commit, while also adding `requires workflow-cli#296` somewhere in the git commit message [like so](https://deis.com/docs/workflow/contributing/submitting-a-pull-request/#cross-repo-commits)?  That way the e2e run should pick up/use the workflow-cli artifact as well...

kwent:
@vdice Done.

vdice:
Jenkins, test this please.

kwent:
Jenkins failed with

```
Installing Workflow CLI version '37ede0d' via url 'https://storage.googleapis.com/workflow-cli-master/37ede0d/deis-37ede0d-linux-amd64'
17:17:05 curl: (22) The requested URL returned error: 404 Not Found
```

vdice:
@kwent Ah yes -- we need to get the workflow-cli artifact building (and uploaded) from https://github.com/deis/controller-sdk-go/pull/118 -- encountering build errors currently; will comment on that ticket.

vdice:
Jenkins, test this please.

vdice:
E2E tests success! :tada: https://ci.deis.io/job/workflow-chart-e2e/1076/consoleFull

bacongobbler:
So this PR adds the idea of delaying the release... But after the release has been delayed, how does one deploy it with the CLI? Can you run me through the typical story on how one would use this feature? I see reference documentation on the API but nothing on how one would actually use this feature in the real world.

bacongobbler:
Essentially I'm looking for another section in https://deis.com/docs/workflow/applications/deploying-apps/ to flesh this feature out.

kwent:
@bacongobbler 

Here is my workflow story:

We have an Jenkins listening to our git push and doing the build for us. Output of this process is docker image. We don't want to deploy them now but we want deis to be aware of those builds and let the developer choose what release to deploy later on via `deis deploy` / `deis releases:deploy`.

rootfs/api/models/release.py
```diff
@@ -24,6 +24,7 @@ class Release(UuidAuditedModel):
     version = models.PositiveIntegerField()
     summary = models.TextField(blank=True, null=True)
     failed = models.BooleanField(default=False)
+    deployed = models.BooleanField(default=True)
```
bacongobbler: I'm in favour of this feature overall, however this really irks me. We should not be maintaining "cluster state" in the database. Rather, I feel like any check to see if the release is deployed should be a call out to kubernetes to see if the deployment is present or not.

What do you think?

mboersma: > any check to see if the release is deployed should be a call out to kubernetes...

That sounds more robust. We have run into issues trying to mirror cluster state in the database before (see the now-defunct `Containers` model for an example).

mboersma:
These PRs haven't been updated in a few months. @kwent I'm going to close this, but please re-open if you can address the remaining issues, and thanks again!

PR #1271, opened 2017-03-22
feat(Makefile,rootfs): run unit and style tests in container

### LINKED ISSUE #18: Proposal: containerize the development environment

mboersma:
Coerces `make test-style` and `make test-unit` to run in containers, which avoids the need to have Python 3.5 and the packages in `dev_requirements.txt` on your local workstation. Also updates the fork of requests-mock to that in the deis org.

Closes #18.
Closes #1049.

See also deis/docker-python-dev#11 and deis/dockerbuilder#120. If that approach is acceptable, this could be refactored similarly for `make test-style` at least.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -32,6 +32,7 @@ _tests/example-*/
 # coverage reports
 .coverage
 coverage.out
+coverage.xml
 htmlcov/
 
 # python virtual environments for testing
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,20 +5,12 @@ branches:
   only:
     - master
 services:
-  - postgresql
   - docker
 env:
   - DOCKER_BUILD_FLAGS="--pull --no-cache"
 sudo: required
-addons:
-  postgresql: "9.4"
 cache: pip
-before_install:
-  - createdb -U postgres deis
-install:
-  - pip install -r rootfs/requirements.txt
-  - pip install -r rootfs/dev_requirements.txt
 script:
   - make test
 after_success:
-  - pushd rootfs/ && codecov && popd
+  - make upload-coverage
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@ SHORT_NAME ?= $(COMPONENT)
 
 include versioning.mk
 
-SHELLCHECK_PREFIX := docker run --rm -v ${CURDIR}:/workdir -w /workdir quay.io/deis/shell-dev shellcheck
+SHELLCHECK_PREFIX := docker run -v ${CURDIR}:/workdir -w /workdir quay.io/deis/shell-dev shellcheck
 SHELL_SCRIPTS = $(wildcard rootfs/bin/*) $(shell find "rootfs" -name '*.sh') $(wildcard _scripts/*.sh)
 
 # Test processes used in quick unit testing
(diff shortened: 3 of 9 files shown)


vdice:
Excited for this!

Should I be able to now run `make test` right off the bat w/o performing any of the classic, python-related setup?  I'm seeing:
```
docker run -v `pwd`/rootfs:/app quay.io/vdice/controller:git-b39ee95.test /app/bin/test-style
docker run --rm -v /Users/vaughndice/work/go/src/github.com/deis/controller:/workdir -w /workdir quay.io/deis/shell-dev shellcheck rootfs/bin/boot rootfs/bin/reload rootfs/bin/test-style rootfs/bin/test-unit  _scripts/deploy.sh
cd rootfs && python manage.py check
  File "manage.py", line 10, in <module>
    from django.core.management import execute_from_command_line
ModuleNotFoundError: No module named 'django'
make: *** [test-check] Error 1
```
Full output [here](https://gist.github.com/vdice/f5c77b413c81623dd86b85ef9fb5a478)

bacongobbler:
Not trying to poopoo on this PR, but have you considered using docker-compose to run the unit tests instead of bundling everything inside one big container, or does that miss the mark in some form?

I am also excited about this :)

mboersma:
>  have you considered using `docker-compose`

I did actually, but the impetus for this was to avoid installing tools and libraries locally. I initially tried to rebase and polish up #1049 but it was too rusty and had some unrelated changes in it.

It's a bit hacky, and the biggest downside I can think of is that `Dockerfile` and `Dockerfile.test` could diverge if we aren't paying attention. Also it would be nice to solve this in a general way using our [python-dev]() image, but that was turning out to be a bit painful, most of the changes end up being specific to controller rather than common, and this is the only Python component that has unit tests.

mboersma:
> Should I be able to now run `make test` right off the bat

Yes, except I hadn't cleaned up that target yet, only `make test-style` and `make test-unit`. Should be good now.

bacongobbler:
> but the impetus for this was to avoid installing tools and libraries locally.

Very good point. I guess that just would bring us back to the root issue: hacking on the controller is significantly different than any other Workflow component.

As far as Dockerfile changes go, we can improve it over time. This should get us through the door. Thanks for indulging!

seanknox:
very much dig this! runs goodly for me on a fresh repo clone after the `make test` fix.

mboersma:
> I guess that just would bring us back to the root issue

Maybe, unless we want to require docker-compose, which doesn't seem too onerous. This isn't blocking #1243, strictly speaking, so I'm glad to take a different approach if that's the consensus.

mboersma:
Looks like codecov.io hasn't reported on this PR. Maybe we broke that? I will check it out...

bacongobbler:
not a moment too soon... python-ldap can't be compiled on Windows and they don't provide a compiled wheel for python 3.5. I'll give this a spin.

mboersma:
I fixed the issue with codecov.io reporting--it should be commenting here, *he said hopefully*.

mboersma:
Coverage reporting on this PR seems to work now at [codecov.io](https://codecov.io/gh/deis/controller/pull/1271).

PR #1279, opened 2017-04-04
chore(requirements): update Django to 1.11 LTS

mboersma:
See https://www.djangoproject.com/weblog/2017/apr/04/django-111-released/ This is an LTS release of Django, so it's supported for three years.

I tested all the changed links to ensure they still point to relevant areas in the Django documentation.


mboersma:
VS Code seems to have aggressively reformatted some lines I didn't intend to touch. I'll revert those changes in a bit to minimize the diff here.

PR #1283, opened 2017-04-11
chore(ingress): Adding unit tests

kris-nova:
Adding unit tests for ingress

Non breaking change

closes https://github.com/deis/controller/issues/1275

### DIFF
--- a/rootfs/scheduler/resources/ingress.py
+++ b/rootfs/scheduler/resources/ingress.py
@@ -55,7 +55,7 @@ class Ingress(Resource):
         return response
 
     def delete(self, namespace, ingress):
-        url = self.api("/namespaces/{}/ingresses/{}", namespace, ingress)
+        url = "/apis/extensions/v1beta1/namespaces/%s/ingresses/%s" % (namespace, ingress)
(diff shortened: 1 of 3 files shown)


bacongobbler:
looks like `make test-style` is failing on this PR. Would you mind fixing that up? Thanks!

kris-nova:
Oh dang, thanks @bacongobbler they should be fixed now.

rootfs/scheduler/resources/ingress.py
```diff
@@ -55,7 +55,7 @@ def create(self, ingress, namespace, hostname):
         return response
 
     def delete(self, namespace, ingress):
-        url = self.api("/namespaces/{}/ingresses/{}", namespace, ingress)
+        url = "/apis/extensions/v1beta1/namespaces/%s/ingresses/%s" % (namespace, ingress)
```
bacongobbler: good catch. I also just checked to confirm and this API endpoint is also the same for v1.5. We'll have to refactor the `self.api` function now that kubernetes is going with the aggregated API route, but this works for now. 👍 

I'll file an issue for this.

PR #1287, opened 2017-04-27
ref(ci): remove travis, update build status

vdice:
ref https://github.com/deis/jenkins-jobs/issues/343


mboersma:
We don't care about Travis CI's testing on this PR, so this is ready for merge.

PR #1300, opened 2017-05-22
chore(requirements): update requests and requests-toolbelt

mboersma:
See http://docs.python-requests.org/en/latest/community/updates/#release-history
and https://pypi.python.org/pypi/requests-toolbelt


mboersma:
Jenkins, test this please.

PR #1314, opened 2017-07-25
chore(requirements): update support libraries

mboersma:
Lots of bug fixes in here--see each commit for the specific changelog.


bacongobbler:
this actually may have a fix for someone who was commenting that LDAP wasn't working for them recently:

> Fix #73: Work around Django 1.11 bug with multiple authentication backends.

LGTM

PR #1317, opened 2017-08-07
fix(api/models/app): Catch unhandled error

### LINKED ISSUE #1313: Critical flaw in scale method allows for entire namespace deletion
I brought this up in the Slack channel and it didn't seem to be taken very seriously so I'm creating this issue in hopes that someone takes a deeper look. Based on my own investigation it seems that there is a certain scenario in which `deis scale` can actually destroy an entire app within it's namespace. This also includes other components deployed to the same namespace. 

Calling `deis scale` runs the scale method in app.py which then calls self.create which is where the problem occurs. On line 199 there is a call to `_scheduler.ns.get(namespace)` which could raise a KubeException error. The code assumes that an error thrown at that moment means the namespace doesn't exist but this isn't always the case. In the controller log I got a "There was a problem retrieving data from the Kubernetes API server". This DOES NOT mean the namespace doesn't exist.

Following the exception from line 199, an attempt to create the namespace is made on line 201. This call will fail because of "409 Conflict namespaces [namespace] already exists". And that's when the outer exception is caught on line 215.

From there it's pretty clear what happens. The code assumes that "something really horrible" happened and proceeds to delete the apps namespace on line 218.

Let me know if there's anything I can provide to assist or if I'm just way off in my investigation of the code.

lshemesh:
Fixes #1313

An unhandled exception in ns.create due to "Conflict namespaces" will wipe out an entire namespace. I've added a catch here to protect against that occurring.

### DIFF
--- a/rootfs/api/models/app.py
+++ b/rootfs/api/models/app.py
@@ -198,7 +198,11 @@ class App(UuidAuditedModel):
             try:
                 self._scheduler.ns.get(namespace)
             except KubeException:
-                self._scheduler.ns.create(namespace)
+                try:
+                    self._scheduler.ns.create(namespace)
+                except KubeException as e:
+                    raise ServiceUnavailable('Could not create the Namespace in Kubernetes') from e
+
             if settings.KUBERNETES_NAMESPACE_DEFAULT_QUOTA_SPEC != '':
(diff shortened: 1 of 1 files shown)


deis-admin:

mboersma:
Jenkins, add to whitelist.

